PR #3, opened 2015-01-30
Feature memory write

rjktcby:
This patch moves code of ktxStream and its implementations into separate files. It also adds function ktxWriteKTXM which writes KTX data into memory region.

### DIFF
--- a/include/ktx.h
+++ b/include/ktx.h
@@ -364,7 +364,7 @@ ktxWriteKTXN(const char* dstname, const KTX_texture_info* imageInfo,
  * Writes a KTX file using supplied data into memory.
  */
 KTX_error_code
-ktxWriteKTXM(const void **bytes, GLsizei *size, const KTX_texture_info* textureInfo,
+ktxWriteKTXM(unsigned char** bytes, GLsizei* size, const KTX_texture_info* textureInfo,
                          GLsizei bytesOfKeyValueData, const void* keyValueData,
                          GLuint numImages, KTX_image_info images[]);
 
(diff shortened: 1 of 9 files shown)


msc-:
Thank you very much for contributing these changes. Please address the comments I have made. Also it would be very helpful if you could add some tests for the memory streams to the loader test program. For example, a test could read a known KTX file, write it to memory, read it from memory and then display it. (I'd like to display the expected image as well as the actual result, but that feature is currently missing for all the tests.)

I am currently traveling and so an unable to build and test any changes. So it will be about 3 weeks before I can merge anything.

rjktcby:
> Why change from void** to unsigned char**?
Just thought to use the same type as in ktxHashTable_Serialize

Thanks a lot for the feedback, i'll fix everything you mentioned.

rjktcby:
There might be a mess in indentation, i'm more used to use spaces instead of tabs, hope there's a way to solve it. Maybe i'll have to make some project specific preferences for my editor and autocorrect all this indentation stuff.
As for the memory test, I'll try to add it later, don't have much time right now to dive into current tests architecture.

msc-:
I merged the change along with a fix to build loadtests from the reorganized source and a minor tweak to the ktxWriteKTXM description.

We still need a test for this new code. How to do you know it works? Unfortunately loadtests only runs on Windows at present. Do you have access to a Windows system?

rjktcby:
> Do you have access to a Windows system?
Unfortunately no, i don't. Maybe loadtests should be made cross-platform? Any ideas for that?

msc-:
I'm in the middle of making them cross-platform by using SDL for the platform interface and UI and GYP to generate projects. Currently I have them running on iOS, OS X and Windows. I expect to commit this version soon to a new incoming branch. Then I will look at making them run on Android and GNU/Linux.

What system(s) are you working on.

rjktcby:
i'm all Linux. Debian and Ubuntu

PR #5, opened 2015-05-06
Improve GNU build system support

Haringat:


msc-:
Please tell me in a nutshell the purpose of these changes. I could already build the KTX library for GNU/Linux using the GNUmakefile in build/GNUmake. The loadtests framework is Windows only so the GNUmakefile does not build that.

I am in the middle of an update to make the load tests portable. This uses GYP to generate project files for multiple platforms and SDL2 to provide a portable framework.

It is not clear to me how this major change will interact with this pull request given my lack of understanding of the purpose of these changes. Please give me your thoughts.

I just attempted to push a new branch with these changes, so that you could look at them, but it uses the new GIT LFS (large file storage) feature, which doesn't seem to be quite ready yet, and the push failed.

Haringat:
Well the reason why I did that was simply that your GNUmakefile did not work for me.
OS: Mageia 5
graphics card: NVIDIA GTX 750 TI
driver: nvidia official driver version 340.58
Compiling lib/checkheader.c failed for gl with "* not declared" on »glCompressedTexImage1D«, »glTexImage1D«, »glCompressedTexImage3D« and »glTexImage3D« (GLES1-3 worked fine).
And as you do not provide any input files for your Makefile finding the cause of the problem would have likely been more work than changing the source tree to generate another one (so I did that). As this worked fine (or worked at all) I kept it and improved it further.
I just pushed my latest changes which should make it all portable to any system having the required tools (e.g. autoconf) installed but sadly I could not test it due to a lack of windows or mac or any other OS but linux.
And for your portable framework: have a look at the "glcore" folder. I used a tool called "glLoadGen" to create a source file that loads all those gl functions directly from the graphics card and provide a fitting header for declarations.

msc-:
Thanks for the info. I compiled on Ubuntu using the GNU makefile but I used the standard Khronos GL header files not those included in the linux distribution. There is really no need to use any other headers.

The GNU makefile was hand created, that is why there are no input files.

Thanks for pointing me at glLoadGen. Why didn't you use GLEW, which has equivalent functionality and is already used by ktx.h and libktx? It's available on Linux too, isn't it.

In view of the major changes coming to the build system for app portability, which, among other things, enable the test app to run on any platform, I have decided not to merge this pull request but will leave it open a while longer in case problems emerge with the new system.

Once I push the changes - I am waiting for access to GitHub's new LFS feature - I would be very grateful if you would take a look at generating make files from the GYP files that will be included and building for Linux. I expect to push the initial version in about 2 weeks. They will appear first in a new branch.

msc-:
The new portable loader tests and build system have been merged to master. If something is still needed to build on Magela 5 with the new build system, please submit a new pull request based on the latest master.

PR #7, opened 2015-06-01
WIP: set GL_TEXTURE_MAX_LEVEL to complete texture

corngood:
Setting GL_TEXTURE_MAX_LEVEL allows a texture with a partial mipmap chain to be complete regardless of filtering mode.

Should it use glTexStorage instead if available?

What should happen on ES1/2?

### DIFF
--- a/lib/loader.c
+++ b/lib/loader.c
@@ -470,6 +470,10 @@ ktxLoadTextureS(struct ktxStream* stream, GLuint* pTexture, GLenum* pTarget,
 	if (texinfo.generateMipmaps && (glGenerateMipmap == NULL)) {
 		glTexParameteri(texinfo.glTarget, GL_GENERATE_MIPMAP, GL_TRUE);
 	}
+#ifdef GL_TEXTURE_MAX_LEVEL
+	if (!texinfo.generateMipmaps)
+		glTexParameteri(texinfo.glTarget, GL_TEXTURE_MAX_LEVEL, header.numberOfMipmapLevels - 1);
+#endif
 
 	if (texinfo.glTarget == GL_TEXTURE_CUBE_MAP) {
 		texinfo.glTarget = GL_TEXTURE_CUBE_MAP_POSITIVE_X;


msc-:
Thanks for this change. I need a little time I don't have right now to check a couple of things before merging it:

1. What, if anything, does the KTX spec say about mipmap completeness. I.e is a file with an incomplete set of mipmaps a valid KTX file.
2. The complexity, or otherwise, of make this a run-time rather than compile time choice. I prefer run-time, if not too complex.

Yes, I would very much like to use glTexStorage when available. It is the way of the future in GL. I simply haven't had the time. I would welcome a contribution.

corngood:
There's not much in the spec other than:
https://www.khronos.org/opengles/sdk/tools/KTX/file_format_spec/#2.11

So apparently a valid ktx texture can either request generation, or specify any number of mipmap levels up to the maximum.  'toktx' will generate textures with numberOfMipmapLevels == 1 when neither --mipmap or --automipmap is specified.

I'll update my patch to set MAX_LEVEL before any image data is set, which should be more friendly to the driver.  I'll also take a look at TexStorage if I get a chance.  There's a good overview of this stuff on:

https://www.opengl.org/wiki/Common_Mistakes#Creating_a_complete_texture

Cheers

msc-:
Thanks for the link. It clearly states that the KTX file does not need a complete mipmap pyramid so setting MAX_LEVEL is good. Please update your patch as you suggested.

Regarding TexStorage, over in issue #8 I have made a suggestion that covers both that request and a way of supporting TexStorage. If you have a chance to take a look at TexStorage, please consider following that suggestion.

corngood:
How about this for now?

I changed the ifdef to look for GL_TEXTURE_MAX_LEVEL.  I know it's not part of ES1/2, but it is available on iOS, so this should work there.  My only concern is if any platforms define it but throw an error at runtime.  I think that's unlikely.

Does this project have automated testing running anywhere?

msc-:
Finally merged. Thanks for the change. Sorry for the delay. My computer broke which backed up my work.

Please look at making the loader use TexStorage when available.

PR #9, opened 2015-08-14
Fix dependency for toktx. Name needs to be cmake name of library

Psy-Fi:
This spawns a warning in cmake 3.0


msc-:
Merged. Thank you.

PR #12, opened 2016-01-31
Modifications to several files to allow for the use of GL3W

Evangel63:
A project I'm working on uses GL3W instead of GLEW, this change allows GL3W to be used as a CMake option.

Commented out case GL_ALPHA_INTEGER in writer.c, was unable to find what it was meant to be, and was undefined on both my GLEW and GL3W build attempts.

Added KHR/khrplatform.h from [1] as it was not include by default, and I have not been able to find it on my Linux Mint system.

[1] http://www.opensource.apple.com/source/X11libs/X11libs-60/mesa/Mesa-7.8.2/include/KHR/khrplatform.h?txt


lib/loader.c
```diff
@@ -128,7 +128,9 @@ static GLboolean supportsSRGB = GL_TRUE;
  * @brief workaround mismatch of glGetString declaration and standard string
  *        function parameters.
  */
-#define glGetString(x) (const char*)glGetString(x)
+#ifndef KTX_USE_GL3W_GETPROC
```
msc-: Why is this #ifdef needed? Does GLEW have the same define? Or is it just unnecessary because the compiler you are using makes char unsigned? If the latter there is no need to #ifdef it. The define is here because gl.h defines this as returning GLubyte but most compilers make char signed.

msc-:
Thank you for this pull request.

I am about to merge a huge set of changes from the 'incoming' branch, amongst which is the addition of a new directory 'other_include' that contains all the Khronos header files including KHR/khrplatform.h. The purpose of these changes is to provide a portable framework for the test programs.

Please integrate your changes with that branch instead. Note that the linux version of gl3loadtests uses static linking, not `GetProcAddr` queries. These are only used on Windows. Also the tests are using SDL not GL3W so probably your changes should be limited to libktx as is this pull request.

GL_ALPHA_INTEGER was added in OpenGL 3.0. It is defined in glext.h. It is also defined in the glew.h that I am using which I downloaded from the glew project here on GitHub. It should not be commented out.

Also please respond to my comment on the change in loader.c.

msc-:
There are further changes coming that affect this pull request. #includes of `gl.h`, etc, have been moved from `ktx.h` to `loader.c`. The changes are on the branch `noglcontext`.  I will be merging them to incoming in the next few days. If you still want to me to include changes for using GL3W, please make them on the `noglcontext` branch, or `incoming` if `noglcontext` has disappeared, and submit a new pull request.

msc-:
Closing due to lack of any response in 6 months to the points raised.

PR #13, opened 2016-02-03
CPU decode optimize.

yutaka-takeda-drecom:
Faster CPU decoding of ETC2 EAC.
about 5% faster. (Measured in Android and iOS)

### DIFF
--- a/lib/etcdec.cxx
+++ b/lib/etcdec.cxx
@@ -726,21 +726,21 @@ void decompressBlockPlanar57c(unsigned int compressed57_1, unsigned int compress
 
 	int xx, yy;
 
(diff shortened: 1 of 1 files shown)


msc-:
This change introduces artefacts when decoding ETC2 images. Please fix and update this merge request.

BTW, you can loadtests on the incoming branch can now be built for iOS.

yutaka-takeda-drecom:
Thank you for reply.

I have to do loadtest.
So, I found a bug .
Bug I was able to fix it,
In that code , there was no speed improvement .

sorry.

msc-:
Thanks for trying.

PR #19, opened 2016-04-11
Issue 18 linux

rjktcby:
This resolves shared libraries builds on Linux systems. Tested on Debian 8, build systems generated with forked gyp: https://github.com/msc-/gyp


msc-:
Thanks. Merged to incoming. Will merge to master after testing on Windows.

msc-:
Merged to master after fixing issues with the Mac build.

PR #29, opened 2016-12-25
Fix toktx crash when argv[0] has no slash.

mpersano:


msc-:
Thank you for the fix.

PR #41, opened 2018-04-19
Fix compilation error when defined KTX_OPENGL_ES2=1.

mmusial:
GLAPIENTRY was used instead of GL_APIENTRY in PFNGLGETSTRINGIPROC typedef.


msc-:
Thank you for the fix and apologies for the delay merging. I have been working on some major items which I'll be landing in incoming in the near future.

PR #42, opened 2018-05-09
Fixed texture arrays loading

kuncarous:


msc-:
Thank you for the fix.

I will soon be landing a change in incoming which includes a load test for array textures along with many other major changes. This will help prevent such errors in future.

PR #63, opened 2018-12-11
Fixed typo

Vallentin:


MarkCallow:
Thanks for the fix. I can't merge it yet because of CI issues even though they are completely unrelated to this PR. Sorry. A download site has stopped service and the place I need to report the problem is insisting on completing a reCaptcha uisng a version that has been shutdown. Sigh!

PR #105, opened 2019-08-01
Ktx2 branch - added a new API function for multithreaded Basis compression

swapsen:
Hey,

this is what I changed to add the ability to do multi-threaded Basis compression. It's up to the client software to figure out how many threads it wants to request.

If one initializes the setup struct to 0 (e.g. " = {};"), it should "just work" with one thread.


MarkCallow:
Please close this pull request, which is against incoming and open a new one against the ktx2 branch.

PR #108, opened 2019-08-05
Added more flexible API for Basis encoding.

swapsen:
Hey,

this is what I changed to add the ability to do multi-threaded Basis compression. It's up to the client software to figure out how many threads it wants to request.

If one initializes the setup struct to 0 (e.g. " = {};"), it should "just work" with one thread.


MarkCallow:
Thank you for this change. It sets the ground nicely for additional compression controls.

MarkCallow:
In adding the additional compression controls I have renamed `quality` and `countThreads` to `qualityLevel` and `threadCount`. This is to warn you of the forthcoming change that will break your code. Apologies. The new names fit better with the other controls.

swapsen:
Perfectly sensible, thanks!

MarkCallow:
I just merged PR #114 which exposes more Basis compress options and has the name changes I mentioned. Thanks again for kick-starting this.

PR #117, opened 2019-09-03
Support NPOT textures

antonio-caggiano-arm:
Hi there.

While trying to generate a KTX with NPOT ASTC mipmaps, I got an error while verifying the correctness of the image size while calling `ktxTexture_setImageFromStream(...)`. To be precise, the check in [writer.c:111](https://github.com/KhronosGroup/KTX-Software/blob/master/lib/writer.c#L111) fails:
```cpp
if (srcSize != packedBytes)
    return KTX_INVALID_OPERATION;
```
That check is needed to verify that the input image size is actually equal to a value computed by calling `ktxTexture_GetImageSize(...)`.

Having a closer look at the implementation of that function, I realized that the method to calculate `blockCount` may be incorrect, thus I added a `ceil` operation to make sure that we get the right number of blocks.

### DIFF
--- a/lib/texture.c
+++ b/lib/texture.c
@@ -886,13 +886,13 @@ ktxTexture_GetImageSize(ktxTexture* This, ktx_uint32_t level)
 
     formatInfo = &((ktxTextureInt*)This)->formatInfo;
 
-    float levelWidth  = This->baseWidth >> level;
-    float levelHeight = This->baseHeight >> level;
-    int x = ceilf(levelWidth / formatInfo->blockWidth);
-    int y = ceilf(levelHeight / formatInfo->blockHeight);
-    blockCount.x = MAX(1, x);
-    blockCount.y = MAX(1, y);
-    blockSizeInBytes = formatInfo->blockSizeInBits / 8;
+    float levelWidth  = (float) (This->baseWidth >> level);
+    float levelHeight = (float) (This->baseHeight >> level);
+    blockCount.x      = (ktx_uint32_t) ceilf(levelWidth / formatInfo->blockWidth);
+    blockCount.y      = (ktx_uint32_t) ceilf(levelHeight / formatInfo->blockHeight);
+    blockCount.x      = MAX(1, blockCount.x);
+    blockCount.y      = MAX(1, blockCount.y);
+    blockSizeInBytes  = formatInfo->blockSizeInBits / 8;
 
     if (formatInfo->flags & GL_FORMAT_SIZE_COMPRESSED_BIT) {
         assert(This->isCompressed);
--- a/lib/texture.c
+++ b/lib/texture.c
@@ -40,6 +40,7 @@
 #include "memstream.h"
 #include "gl_format.h"
 #include "uthash.h"
+#include <math.h>
 
 /**
  * @internal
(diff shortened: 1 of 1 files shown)


MarkCallow:
Thank you for looking into this. Please provide an example of the case where the current code is failing, i.e the texture dimensions and the level where it fails.

antonio-caggiano-arm:
I will give you this image (from San Miguel 2.0 [CC BY 3.0](https://creativecommons.org/licenses/by/3.0/) © [Guillermo M. Leal Llaguno](http://www.evvisual.com/)), as a reference:



I converted it to astc using [Mali Texture Compression Tool](https://developer.arm.com/tools-and-software/graphics-and-gaming/graphics-development-tools/mali-texture-compression-tool) to generate a file named `map.astc`.

Then I follow these steps and I get the error mentioned above:
```cpp
ktxTextureCreateInfo info = {};

info.numLayers = 1;
info.isArray = KTX_FALSE;
info.glInternalFormat = GL_COMPRESSED_SRGB8_ALPHA8_ASTC_8x8_KHR;
info.baseWidth = 99;
info.baseHeight = 200;
info.baseDepth = 1;
info.numDimensions = 2;
info.numFaces = 1;
info.numLevels = 1;
info.generateMipmaps = KTX_FALSE;

ktxTexture* ktx_texture;
ktxTexture_Create( &info, KTX_TEXTURE_CREATE_ALLOC_STORAGE, &ktx_texture );

ktx_uint8_t* data = map_astc_data();
ktx_size_t size = map_astc_size();

ktxTexture_SetImageFromMemory( ktx_texture, 0, 0, 0, data, size );
```

MarkCallow:
Thank you. I was confused by the mention of NPOT. the issue really is that the calculation is not accounting for the fact that block-compressed textures are padded to the next full block so in this case the actual data in effect has baseWidth = 104.

I'll merge this in a few days. I want to wait until I'm ready to merge a matching fix into the ktx2 branch.

lib/texture.c
```diff
@@ -884,8 +885,13 @@ ktxTexture_GetImageSize(ktxTexture* This, ktx_uint32_t level)
...
-    blockCount.x = MAX(1, (This->baseWidth / formatInfo->blockWidth)  >> level);
-    blockCount.y = MAX(1, (This->baseHeight / formatInfo->blockHeight)  >> level);
+
+    float levelWidth  = This->baseWidth >> level;
+    float levelHeight = This->baseHeight >> level;
+    int x = ceilf(levelWidth / formatInfo->blockWidth);
+    int y = ceilf(levelHeight / formatInfo->blockHeight);
+    blockCount.x = MAX(1, x);
+    blockCount.y = MAX(1, y);
     blockSizeInBytes = formatInfo->blockSizeInBits / 8;
```
MarkCallow: Have you compiled this with Visual Studio? I got some warnings from VS2017 about truncation. Some casts are needed to avoid these. Please update this PR. Something like the following should work
```
    float levelWidth  = (float)(This->baseWidth >> level);
    float levelHeight = (float)(This->baseHeight >> level);
    blockCount.x = (ktx_uint32_t)ceilf(levelWidth / formatInfo->blockWidth);
    blockCount.y = (ktx_uint32_t)ceilf(levelHeight / formatInfo->blockHeight);
    blockCount.x = MAX(1, blockCount.x);
    blockCount.y = MAX(1, blockCount.y);
```

MarkCallow:
Sorry for the extended delay.

PR #130 is the equivalent fix for the ktx2 branch.

PR #129, opened 2019-10-07
Round width & height to block-size multiple in calcImageSize. 

MarkCallow:


MarkCallow:
GitHub PR edit failed to change base even when I clicked the confirmation button so close this without action.

PR #139, opened 2019-11-06
Fixed inconsistent use of malloc/free vs. new[]/delete for ktxTexture…

atteneder:
…->pData.

When transcoding a basis texture, pData was allocated via `new uint8_t[image_data_size];`, but within `ktxTexture_destruct` it is released with `free(This->pData);`.

It seems to not bother at first, but it caused crashes on some platform/build type configurations (e.g. Android release, WebAssembly, macOS release).


MarkCallow:
Thank you for this fix.

PR #149, opened 2019-11-29
Fix ktxinfo to print Basis information for all levels

zeux:
This fixes a typo layerCount -> levelCount so that we can print all
image descriptions for all levels.


MarkCallow:
Thank you for the fix.

PR #170, opened 2020-02-08
Change mip level padding to match draft 18 rules.

### LINKED ISSUE #151: [ktx2] ktxcheck doesn't catch incorrect mip level layout
The .ktx files in this .zip file have been encoded incorrectly:

[corset-separate.zip](https://github.com/KhronosGroup/KTX-Software/files/3906767/corset-separate.zip)

Specifically, the mip level index here is wrong - the file offsets increase in the level index (instead of decreasing); additionally, because the index is wrong, attempting to decode the levels using Basis would yield insufficient data - for example, here the largest mip level in the index is specified as having just 4 bytes of supercompressed data.

It would be nice if ktxcheck checked that the mip levels are located sequentially from smallest to largest in the file (which would flag this file as being invalid), and also tried to decode supercompressed data / validate individual mip level sizes for non-supercompressed data as well.

### LINKED ISSUE #152: ktx2check should check levelCount != 0 for  block-compressed format
Make sure this is validated. Also typeSize = 1.

MarkCallow:
Changes writer, reader, loaders, tools (especially ktx2check) and tests to be compliant with the padding specification of the KTX2 Specification draft 18.

Fixes #151. Fixes #152.


tools/ktx2check/ktx2check.cpp
```diff
@@ -964,21 +1046,50 @@ ktxValidator::validateLevelIndex(validationContext& ctx)
...
+                || levelIndex[level].byteOffset == 0) {
+                 addIssue(logger::eError, LevelIndex.ZeroOffsetOrLength, level);
+                 continue;
+            }
+            if (levelIndex[level].byteOffset != expectedOffset) {
+                addIssue(logger::eError,
+                         LevelIndex.IncorrectByteOffset,
+                         level);
+            }
+            if (levelIndex[level].byteLength < lastByteLength)
```
zeux: I don't think this is a correct check. This compares byte lengths, not offsets, and with super compression there's no guarantee that a smaller mip level will compress to a smaller number of bytes. For example, here's what basisu generates for one of the textures I was running it on (FlightHelmet_normal.png from glTF-Sample-Models):

Level Index

Level0.byteOffset: 0x1ae4f
Level0.byteLength: 265427
Level0.uncompressedByteLength: 0
Level1.byteOffset: 0x8b32
Level1.byteLength: 74525
Level1.uncompressedByteLength: 0
Level2.byteOffset: 0x3f73
Level2.byteLength: 19391
Level2.uncompressedByteLength: 0
Level3.byteOffset: 0x2ae0
Level3.byteLength: 5267
Level3.uncompressedByteLength: 0
Level4.byteOffset: 0x251c
Level4.byteLength: 1476
Level4.uncompressedByteLength: 0
Level5.byteOffset: 0x2389
Level5.byteLength: 403
Level5.uncompressedByteLength: 0
Level6.byteOffset: 0x231c
Level6.byteLength: 109
Level6.uncompressedByteLength: 0
Level7.byteOffset: 0x22fc
Level7.byteLength: 32
Level7.uncompressedByteLength: 0
Level8.byteOffset: 0x22f1
Level8.byteLength: 11
Level8.uncompressedByteLength: 0
Level9.byteOffset: 0x22ed
Level9.byteLength: 4
Level9.uncompressedByteLength: 0
Level10.byteOffset: 0x22e8
Level10.byteLength: 5
Level10.uncompressedByteLength: 0
Level11.byteOffset: 0x22e3
Level11.byteLength: 5
Level11.uncompressedByteLength: 0

Note that level 8 is smaller than level 9 (4 bytes vs 5 bytes).

MarkCallow: I didn't think of that. Can you think of another way to check the mip levels of a supercompressed texture are in the correct order in the file?

zeux: So in my original file with the problematic layout (#151), the file offsets were ordered incorrectly - the byte offsets have to be decreasing (like in the snippet above), and I think they were increasing. So the same check for byteOffset should work? (although since byteOffset is already validated maybe that's not necessary?)

zeux: Yeah, if I run my original incorrect file against the new ktx2check I get this:

Issues in: Corset_baseColor.ktx
    ERROR: Level 11 byteOffset does not match expected value.
    ERROR: Level 10 byteOffset does not match expected value.
    ERROR: Larger mip levels are before smaller.
    ERROR: Level 9 byteOffset does not match expected value.
    ERROR: Larger mip levels are before smaller.
    ERROR: Level 8 byteOffset does not match expected value.
    ERROR: Larger mip levels are before smaller.
    ERROR: Level 7 byteOffset does not match expected value.
    ERROR: Larger mip levels are before smaller.
    ERROR: Level 6 byteOffset does not match expected value.
    ERROR: Larger mip levels are before smaller.
    ERROR: Level 5 byteOffset does not match expected value.
    ERROR: Larger mip levels are before smaller.
    ERROR: Level 4 byteOffset does not match expected value.
    ERROR: Larger mip levels are before smaller.
    ERROR: Level 3 byteOffset does not match expected value.
    ERROR: Larger mip levels are before smaller.
    ERROR: Level 2 byteOffset does not match expected value.
    ERROR: Larger mip levels are before smaller.
    ERROR: Level 1 byteOffset does not match expected value.
    ERROR: Level 0 byteOffset does not match expected value.
    ERROR: Size of image data in file does not match size calculated from levelIndex.

So I don't think the "larger mip levels are before smaller" check is actually crucial.

What *would* be nice to do is to actually decompress the super-compressed data with the correct mip level width/height and verify that transcoder doesn't fail, but this can probably be done separately.

MarkCallow: For Supercompressed or VK_FORMAT_UNDEFINED The byteOffset check only serves to validate the chain of offsets and lengths, i.e offset+length = next offset. It cannot tell if larger levels come before smaller.

zeux: Sure, but "larger" should be compared in context of the mip level dimensions, not encoded data.

I feel like the only good solution here is to try to decode supercompressed data if present (and validate exact byteLength for non-supercompressed data - not sure if this is already being done) with the mip level dimensions.

MarkCallow: Non-supercompressed is already being validated to exact byte length.

Adding decoding will have to wait until later.  I think a smaller level having more data than a larger is only likely to happen at the very smallest image sizes. So for now I will add a 3 byte tolerance and see if I get more complaints.

zeux: > So for now I will add a 3 byte tolerance and see if I get more complaints.

There is no expectations that the validator should have about contents of individual mip levels. It's common place in computer graphics to encode mips in a special way to achieve certain effects, and it's going to be impossible to predict the behavior of the supercompressor. If the largest mip level is completely blank and the rest of the mip chain isn't, to achieve effects like a pattern disappearing really close by, the supercompressed size of the first mip level is going to be insignificant.

I understand that decoding is more involved, but what I'm suggesting is to *remove* this check from this PR, and just wait until decoding can be implemented.

MarkCallow: Thanks for the valuable insight. I have moved the test so it is not done for supercompressed.

PR #172, opened 2020-02-10
Import KTX2Loader from threejs PR.

donmccurdy:
Imports KTX2Loader from open PR (https://github.com/mrdoob/three.js/pull/18490) to consolidate development in one place. The loader now depends on changes to the threejs library (addition of ETC2 support) which means development is easier to manage over there.

- Adds an import map shim, to ensure KTX2Loader uses the right threejs CDN during development. The release on CDN does not yet have the ETC2 support.
- Removes requestIdleCallback invocation, so demo can be tested on Safari.

The eventual goal here is that — once the KTX2Loader is released with threejs — this demo will can demonstrate KTX2 loading with the official threejs release library. If/when we later add a separate pure-JS KTX2 loader, that is engine-agnostic, we can decide how to best publish that.


tests/webgl/llt-three/index.html
```diff
@@ -61,13 +72,6 @@ <h2>msc_basis_transcoder test</h2>
     const mesh = new THREE.Mesh(geometry, material);
     scene.add(mesh);
 
-    var dict = {
```
MarkCallow: Thanks for noticing and removing this leftover from troubles using constants as keys. I was trying to understand if I was doing something wrong in the way I was specifying the dictionary. I wasn'.t I looked further and I discovered the bracket notation for doing what I wanted.

tests/webgl/llt-three/index.html
```diff
@@ -77,22 +81,18 @@ <h2>msc_basis_transcoder test</h2>
         [THREE.RGB_ETC1_Format]: "RGB_ETC1"
```
MarkCallow: Can we add THREE.RGB_ETC2_Format?

MarkCallow:
From where will users of KTX2Loader.js download msc_basis_transcoder.js? (The index.html here is correctly referring to the local copy.)

tests/webgl/llt-three/index.html
donmccurdy: Done – although the latest threejs release doesn't have ETC2 support yet, so (regardless of this change) the demo will likely fail on Android for now. I expect the ETC2 support will be released at the end of February.

donmccurdy:
For now using a local copy of `msc_basis_transcoder.js` seems best, especially in case any changes are made within this repository. ThreeJS will likely publish a copy in our releases as well, as we do for `.basis`.

MarkCallow:
Damn. I should have make a branch and tested this before merging. I am getting the message

```
Loading failed for the <script> with source “https://unpkg.com/es-module-shims@0.4.6/dist/es-module-shims.js”.
```

in the Firefox web console. That is the only message. There is no indication of why it failed. There is nothing in the browser console either.

I may have to back out this PR.

donmccurdy:
Hm let me check on that, I had tested in Firefox, safari, and chrome earlier.

MarkCallow:
In case it matters, I'm running the http-server included with node.js. I don't think it does because there is no sign of the browser attempting to fetch this script from my test server.

MarkCallow:
Sorry. My fault. I use NoScript and had not trusted unpkg. Now I've done that, it works.

PR #175, opened 2020-02-23
Fix transcoder memory management

zeux:
Instead of returning a memory view into a temporary std::vector, copy
the vector data to a typed array and return that.

### DIFF
--- a/interface/js_binding/transcoder_wrapper.cpp
+++ b/interface/js_binding/transcoder_wrapper.cpp
@@ -254,8 +254,12 @@ namespace msc {
             val ret = val::object();
             ret.set("error", static_cast<uint32_t>(error));
             if (error == KTX_SUCCESS) {
-                // FIXME: Who deletes dst and how?
-                ret.set("transcodedImage", typed_memory_view(dst.size(), dst.data()));
+                val memoryView = val::global("Uint8Array").new_(memory, reinterpret_cast<uintptr_t>(dst.data()), dst.size());
+
+                val memoryCopy = val::global("Uint8Array").new_(dst.size());
+                memoryCopy.call<void>("set", memoryView);
+
+                ret.set("transcodedImage", memoryCopy);
             }
             return std::move(ret);
         }


MarkCallow:
I do not like having to copy the data out. On input there is no alternative but for output there is. Why not use `new std::vector<uint8_t> dst` and a memoryView on to it? In doing this, it may be necessary to change Emscripten compile options to allow memory growth. I'm not sure how to delete the memory when done with it. The Emscripten docs say it is the app's responsibility and to use `.delete()`. On exactly what you use `.delete()` is not clear. The returned arrayBuffer?

zeux:
If you return a memory view, you need to delete the returned memory in C++ once JS is done with it - you can't do this from JS easily, so you'd need something like deleteTranscodedImage in the interface. This makes the interface less idiomatic.

Since the transcoder is already stateful, how about something like:

1) we change transcodeImage to just return the error code
2) we add getTranscodedImage

That way the interface guarantees that you can't leak memory, at the cost of only being able to hold one transcoded image at a time?

MarkCallow:
> (although it's worth noting that zero-copy is only really practical if the transcoder runs outside of a web worker, which is probably unlikely to happen in practice)

Hmm? Why is it not practical with web workers?

I posted a question to emscripten_discuss and got the following response from Gabriel Cuvillier. If not for your comment about web workers I'd say go with this idea. It does require a call to .delete but e.g the THREE.js loader can take care of that. [@donmccurdy](https://github.com/donmccurdy) what do you think? (I hope this mention will get his attention.)

> Using "new std::vector" will prevent the memory delete/overwrite issue, but as you noticed, it is not possible to .delete anything.


> The best for your use case would be to create a new class "TranscodedImage", containing the std::vector dst, and exposed using Embind.  That class will hold responsibility of the vector, and exposing the class using Embind will allow you to call .delete() on it later on.


> Something like (pseudo-code):
>
>     class TranscodedImage {
>         std::vector<uint8_t> dst;
>         val get_typed_memory_view() {
>	       return typed_memory_view(dst.size(), dst.data())
>         }
>
>         void compute_data( ... ) {
>             // actually fills 'dst'
>         }
>     }
>
>
>     EMSCRIPTEN_BINDINGS( mylib )
>     {
>         emscripten::class_<TranscodedImage>( "TranscodedImage" )
>        .constructor( )
>        .function( "compute_data", &TranscodedImage::compute_data )
>        .function( "get_typed_memory_view()", &TranscodedImage::get_typed_memory_view )
>
>     }


>// In your javascript code
>
>     let trans_img = new Module.TranscodedImage()
>
>     trans_img.compute_data(... whatever needed to compute the image ...)
>
>     let img_data = trans_img.get_typed_memory_view()
>
>      ... do whatever needed with img_data. Then, later on, when the data is no more needed
>
>     trans_img.delete();    // will call the TranscodedImage class destructor => deletes the vector contents. So, everything is reclaimed in the end (NB: img_data will no more point to something valid!!)
>
> Note the .delete() call : this is called on the Emscripten-exposed C++ class

zeux:
With web workers you will need to copy data out of the WebAssembly heap so you can’t make the process zero-copy.

The proposal above seems roughly equivalent to storing the decoded data inside transcoder I suggested (but requires extra .delete calls).

MarkCallow:
> With web workers you will need to copy data out of the WebAssembly heap so you can’t make the process zero-copy.

Okay. Can you do that even if the function returns a memory view onto the heap?

> The proposal above seems roughly equivalent to storing the decoded data inside transcoder I suggested (but requires extra .delete calls).

It gives the application the choice of transcoding & uploading  to WebGL 1 image at a time or transcoding several then uploading. In other words its more flexible so I prefer it.

zeux:
> Okay. Can you do that even if the function returns a memory view onto the heap?

Yes, by copying after the function returns (from JS)

> In other words its more flexible so I prefer it.

Sure, that’s fine - I’m going to close this then.

MarkCallow:
Thanks for creating this PR and sparking this discussion. It has been helpful. Please make a new PR based on the proposal above.

donmccurdy:
Is it possible for the JS to pass in a TypedArray or ArrayBuffer, and let the transcoder write into that? I don't have a strong preference, but that seems to be what the WASM transcoder in the basis_universal repository is doing:

```js
const basisFile = new BasisFile( new Uint8Array( buffer ) );
basisFile.startTranscoding();

for ( mip of mips ) {
  const dst = new Uint8Array( basisFile.getImageTranscodedSizeInBytes( 0, mip, format ) );
  const status = basisFile.transcodeImage( dst, 0, mip, format, 0, hasAlpha );
  if (status) { results.push( dst ); }
  else { ... }
}

basisFile.close();
basisFile.delete();
```

Full version: https://github.com/mrdoob/three.js/blob/afc703be33cd6e1a8ea107fea6ff0d4c1aec4aa8/examples/js/loaders/BasisTextureLoader.js#L424-L500

On the other hand, I don't particularly need to do the array allocation in JS, either. I think the API you're describing sounds fine too. three.js will eventually be using the transcoder in a Web Worker; that use case is important for our users. 

Thanks!

MarkCallow:
The last lines of the `transcodeImage function` in the basis transcoder are:

```	 
std::vector<uint8_t> dst_data;
	  
// do stuff to get the transcoded image into dst_data

emscripten::val memory = emscripten::val::module_property("HEAP8")["buffer"];
emscripten::val memoryView = emscripten::val::global("Uint8Array").new_(memory, reinterpret_cast<uintptr_t>(dst_data.data()), dst_data.size());

dst.call<void>("set", memoryView);
```

If `dst.call` "set" is copying the data from the memory view to the passed in array buffer then that works but it involves a data copy which I'd like to avoid. If it is not copying then it is broken just like msc_basis_transcoder.

Since I want to avoid a copy I still favor Gabriel's solution.

donmccurdy:
> If dst.call "set" is copying the data from the memory view to the passed in array buffer...

That would copy the data, yes: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/set

MarkCallow:
@donmccurdy please give PR #179 a try.

MarkCallow:
> @donmccurdy please give PR #179 a try.

By which I mean please modify your loader to use the new pattern and test. The use comment in the transcoder source shows how to use it.

PR #179, opened 2020-03-07
Fix transcoder memory management

MarkCallow:
Allocate a new dst for each image in the form of a TranscodedImage object which is returned by transcode_image. .delete() can be called on this object once the data has been uploaded to WebGL.

### DIFF
--- a/interface/js_binding/transcoder_wrapper.cpp
+++ b/interface/js_binding/transcoder_wrapper.cpp
@@ -399,7 +399,11 @@ the actual data is not specific to that container format.
                                      false);
             if (!error) {
                 let imgData = transcodedImage.get_typed_memory_view();
-                // Upload data in imgData to WebGL.
+
+                // Upload data in imgData to WebGL...
+
+                // Do not call delete() until data has been uploaded
+                // or otherwise copied.
                 transcodedImage.delete();
             }
         }
--- a/interface/js_binding/transcoder_wrapper.cpp
+++ b/interface/js_binding/transcoder_wrapper.cpp
@@ -28,6 +28,21 @@ namespace msc {
     class BasisTranscoderState: public basisu_transcoder_state {
     };
 
+    class TranscodedImage {
+      public:
+        TranscodedImage(size_t size) : image(size) { }
+
+        uint8_t* data() { return image.data(); }
+        size_t size() { return image.size(); }
+
+        val get_typed_memory_view() {
+           return val(typed_memory_view(image.size(), image.data()));
+        }
+
+      protected:
+        std::vector<uint8_t> image;
+    };
+
     class BasisTranscoder : public ktxBasisImageTranscoder {
       public:
         BasisTranscoder() : ktxBasisImageTranscoder(buildSelectorCodebook())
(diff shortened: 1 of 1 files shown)


donmccurdy:
@MarkCallow I'm not sure how to get data out of the TranscodedImage. From your code it looks like you meant to expose a `get_typed_memory_view` method, but I can't access it in JavaScript:

Screen Shot 2020-03-09 at 10 42 15 PM

MarkCallow:
@donmccurdy, sorry about that. There were spurious parens at the end of the `get_type_memory_view` name string in the embind binding. I've just pushed a commit to fix it.

MarkCallow:
Don,

Please try the latest msc_basis_transcoder. I want to get it merged to the ktx2 branch so I can then merge everything to KTX-Software master.

Regards

    -Mark

donmccurdy:
@MarkCallow the new transcoder method works, but I still have to make a copy of its data before transcoding additional levels to avoid data being overwritten. My current usage is something like this:

```js
for ( var imageIndex = 0; imageIndex < numImagesInLevel; imageIndex ++ ) {

	var imageDesc = this.sgd.imageDescs[ imageDescIndex++ ];

	var result = transcoder.transcodeImage(
		imageDesc.imageFlags,
		new Uint8Array( this.arrayBuffer, this.levels[ level ].byteOffset + imageDesc.rgbSliceByteOffset, imageDesc.rgbSliceByteLength ),
		new Uint8Array( this.arrayBuffer, this.levels[ level ].byteOffset + imageDesc.alphaSliceByteOffset, imageDesc.alphaSliceByteLength ),
		targetFormat,
		level,
		levelWidth, levelHeight,
		numBlocksX,
		numBlocksY,
		isVideo,
		false
	);

	if ( result.error ) {

		throw new Error( 'THREE.KTX2Loader: Unable to transcode image.' );

	}

	var levelData = result.transcodedImage.get_typed_memory_view();

	// Make a copy!
	levelData = levelData.slice();

	mipmaps.push( { data: levelData, width: levelWidth, height: levelHeight } );

	result.transcodedImage.delete();

}
```

MarkCallow:
> 	// Make a copy!
>	levelData = levelData.slice();
>
>	mipmaps.push( { data: levelData, width: levelWidth, height: levelHeight } );
>
>	result.transcodedImage.delete();

If `mipmaps.push` is just pushing the level data into some kind of vector for later upload to WebGL then you need to hold off calling `result.transcodedImage.delete()` until you have uploaded the data. `delete()` will free up the memory the returned memory view is looking at which means it will get allocated again for  the next TranscodedImage.

donmccurdy:
Thanks, that makes sense and now works as expected. 👍

What would be the effect of _not_ calling `.delete()`? While, say, transcoding and uploading some 200 MB of textures? Does that just mean 200 MB of memory unnecessarily held, as it would in JS? Or do you know if WASM has some additional memory limit that users of the transcoder need to be aware of?

MarkCallow:
I'm compiling with ALLOW_MEMORY_GROWTH=1 so it should just mean 200MB unnecessarily held.

donmccurdy:
Ok, sounds good!

PR #194, opened 2020-04-02
Don't use function pointers for GLUpload, VkUpload and related functions.

MarkCallow:
Ensures these functions are only linked in when an app explicitly calls them.


MarkCallow:
Ensures these functions are only linked in when an app explicitly calls them.

PR #257, opened 2020-07-05
Fix ImageT behavior in MSVC 2019 in Debug mode

zeux:
This change is necessary because the existing code crashes in MSVC 2019
when debug iterators are enabled (in Debug) on Image destruction.

This happens because Image<T> frees the backing store before destroying
std::vector, and std::vector destructor accesses memory from the backing
store as part of debug iterator code.

The existing code seems error prone and unsafe for other reasons, and
I'm not sure std::vector is justified here.

The change removes std::vector and ImageTBase/ImageT distinction; ImageT
now just uses a plain C array. resize()/crop() weren't used and are
removed as a result to keep the change simpler.


zeux:
Note: there's one more issue in MSVC 2019 in Debug which this change doesn't address yet, but fixing that might require a larger rework of the Image class (briefly, the PreAllocator and the destruction order in Image conflicts with debug iterator mechanism).

MarkCallow:
Manually rebased and merged.

Issue #244, opened 2020-06-26
Issues with draft Windows installer


MarkCallow:
I've had the following issues with the draft Windows installer:

- [ ]  1. The installer name has _Windows_ in it. It needs to indicate whether it is for 32- or 64-bit Windows.
- [ ]  2. Warning from Win 10 requiring bypass to install software from an unidentified developer. (I.e. we need to sign it.)
- [x] 3. Norton antivirus accuses the installer of suspicious activity (SONAR.Heuristic.170) before it even displays its first window. A similar report said the file triggering it was a .ini fle in an installer. That was a false positive as I suspect this is. I've reported it to, er, Broadcom, who seems to have bought Symantec who years ago bought Norton.
- [ ] 4. Attempt to add the install directory to %PATH% fails saying the PATH is too long. It isn't. The maximum supported length is 2048 and my %PATH% is around 1968 characters so there is room for `c:\Program Files\KTX-Software\bin`.
- [ ] 5. Application version numbers have "-dirty" appended.
- [ ] 6. The license display does not understand rtf. The text is large and links aren't displayed correctly.

MarkCallow:
It is necessary to also sign the uninstaller. Since this is normally created during the install, special steps are needed, outlined in https://nsis.sourceforge.io/Signing_an_Uninstaller.

MarkCallow:
No. 5 is because the OpenGL ES emulator is being loaded into a directory in the git workarea. Either put the following outside the workarea else which is preferable, or add a .gitignore rule.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	libEGL.dll
	libEGL.lib
	libGLES_CM.dll
	libGLES_CM.lib
	libGLESv2.dll
	libGLESv2.lib

MarkCallow:
Unbelievably weird. Cmd prompt continues to parse lines starting with REM and if it sees &, it executes what follows. I had to put another REM after the & to stop the 32-bit download.

MarkCallow:
Norton, er ... Broadcom, has confirmed the false positive and say the detection will be removed from their products in the next virus definitions update. The whitelisting may take 24 hours to take effect.

MarkCallow:
I can confirm that Broadcom has fixed item 3. Norton users, and probably Symantec users too, will have to use LiveUpdate to get the latest definitions with the fix.

zeux:
[...]

PR #260, opened 2020-07-06
Attempt to fix Windows uninstaller empty name (`.exe`) by explicitly setting u…

atteneder:
…ninstaller name.

The original problem isn't reproducable on my local setup, but is present inthe CI/CD installer.
I'm not sure this actually fixes the problem, but we have to run it through appveyor to know for sure.


atteneder:
Relatest to #244 point 7

MarkCallow:
I have to figure out how to test this.

MarkCallow:
This appears to have been fixed in master.

Issue #276, opened 2020-07-21
Emscipten builds fail on windows


lexaknyazev:
https://github.com/KhronosGroup/KTX-Software/blob/10ce74544a0577da133ba8f1200ec23c1e19f5ef/cmake/version.cmake#L170-L187

The `WIN32` condition does not work when compiling on Windows with Emscripten, so CMake tries to run `./mkversion` directly and fails.

MarkCallow:
Do you have a suggestion for how to fix this?

DDoS:
Hi, I'm having the same issue. According to the Emscripten.cmake file CMAKE_HOST_WIN32  should be used instead.

https://github.com/emscripten-core/emscripten/blob/master/cmake/Modules/Platform/Emscripten.cmake#L34

MarkCallow:
Fixed by PR #277.

PR #277, opened 2020-07-22
Fix CMake Emscripten build on Windows

DDoS:
Fix for issue #276.


MarkCallow:
Thank you for the fix.

PR #289, opened 2020-08-13
fix: By default create 1D texture when height == 1

kacprzak:
From toktx tool help:
  --2d         If the image height is 1, by default a KTX file for a 1D
               texture is created. With this option one for a 2D texture is
               created instead.


MarkCallow:
Thanks for the fix.

PR #293, opened 2020-08-18
Improved Google Test CMake integration

atteneder:
The CMake scripts now use the [GoogleTest CMake module](https://cmake.org/cmake/help/v3.10/module/GoogleTest.html) to integrate and discover tests.

Main benefit: Tests are invoked by ctest individually, thus it is easier to see which exact test failed.


MarkCallow:
Please rebase on latest master. Many files have been changed due to reuse compliance work.

Regarding the Appveyor build failure, is transcodetests writing to stderr even when successful? Because I've set stop on error in PS, this causes the build to exit. PS considers writing to stderr to indicate an error has occurred, as well as if the exit status != 0.

PR #307, opened 2020-09-05
fix: Raised minimum required CMake version to 3.15

### LINKED ISSUE #304: Incorrect cmake_minimum_required 3.13 (should be 3.15)
Hi, 

I tried to generate the project using Cmake 3.14 and it failed. 

This is because the support for comma separated list of compiler ids ($<CXX_COMPILER_ID:compiler_ids>) was added in version 3.15.

Cheers,

Luca

atteneder:
since we use `$<CXX_COMPILER_ID:compiler_ids>` which was introduced in 3.15:

<https://cmake.org/cmake/help/v3.15/manual/cmake-generator-expressions.7.html#variable-queries>

fixes #304

Let's see if CI/CD succeeds as well.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 # Copyright 2015-2020 The Khronos Group Inc.
 # SPDX-License-Identifier: Apache-2.0
 
-cmake_minimum_required(VERSION 3.13)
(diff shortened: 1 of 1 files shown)


MarkCallow:
Thanks @atteneder.

PR #311, opened 2020-09-11
 Fixed loading PPM files where maxval in not equal to 255 or 65535 for 2 byte color component

kacprzak:
Also swapEndian worked only for greyscale (1 component) images.


tools/toktx/npbmimage.cc
```diff
@@ -440,27 +458,39 @@ readImage(FILE* src, Image& image, int32_t maxval)
...
     }
 
     if (buffer16 != nullptr) {
-        rescale8(image, (uint16_t*)pBuffer, maxval, image.getPixelCount());
+        rescale16to8(image, (uint16_t*)pBuffer, maxval,
+                     image.getPixelCount() * image.getComponentCount());
         delete[] buffer16;
     }
+    // Maxval is whitepoint. Rescale needed if white is not uint MAX.
+    else if( maxval < 255 ) {
```
MarkCallow, line 485: Please change style to match the other ifs, i.e if (foo).

tools/toktx/npbmimage.cc
```diff
@@ -440,27 +458,39 @@ readImage(FILE* src, Image& image, int32_t maxval)
...
+        rescale8((uint8_t*)pBuffer, (uint8_t*)pBuffer, maxval,
+                 image.getPixelCount() * image.getComponentCount());
+    }
+    else if( maxval < 65535 ) {
```
MarkCallow, line 489: And this one.

MarkCallow (changes requested):
This PR breaks all the toktx tests that use npbm images as sources. Please fix the problem. See `tests/toktx-tests.cmake` for the tests. Use the RUN_TESTS target to run them.

tools/toktx/npbmimage.cc
```diff
@@ -482,11 +482,11 @@ readImage(FILE* src, Image& image, int32_t maxval)
         delete[] buffer16;
     }
     // Maxval is whitepoint. Rescale needed if white is not uint MAX.
-    else if( maxval < 255 ) {
```
MarkCallow: space after the if, please.

tools/toktx/npbmimage.cc
```diff
@@ -482,11 +482,11 @@ readImage(FILE* src, Image& image, int32_t maxval)
...
+    else if(maxval < 255) {
         rescale8((uint8_t*)pBuffer, (uint8_t*)pBuffer, maxval,
                  image.getPixelCount() * image.getComponentCount());
     }
-    else if( maxval < 65535 ) {
+    else if(maxval > 255 && maxval < 65535) {
```
MarkCallow, line 489: And here.

MarkCallow (changes requested):
Thank you for the great fix. Do you have some npbm files I could add to to the test suite for the cases where maxval = 65535 and where rescaling is needed?

kacprzak:
Here you go:
[pgm_maxval_testimages.zip](https://github.com/KhronosGroup/KTX-Software/files/5212717/pgm_maxval_testimages.zip)
There are original ascii pgm files with weird maxval. The *_raw.pgm are generated binary versions by ImageMagick. ImageMagick tool converts maxval to next 2^n-1 and rescales every color accordingly. So file map1000.pgm will have maxval==1000 but file map1000_raw.pgm will have maxval set to 1023! I do not know why ImageMagick does that but it is legal behavior and this file is opened correctly by GIMP (If I use GIMP to convert from ascii to binary it will set maxval==65535 in this scenario). I also attached ktx files that I generated (toktx --linear) when testing.

MarkCallow (approved):

MarkCallow:
Thanks again for this. I'll see about adding the images to the tests in a day or 2.

MarkCallow:
macOS Finder preview and the Preview program shows differences between map100_raw.pgm & map100.pgm and map1000_raw.pgm & map1000.pgm. Only Gimp make them look the same. I'll have to try some other programs. Another strange thing is that Finder preview shows a nice gradient while both Gimp and Preview show steps between the shades. I'll have to try some other software.

Are you sure these files have linear OETF?

kacprzak:
Finder probably interpolates between pixels because those images have only 5 pixels. I made them by hand in text editor. Image map100.pgm has maxval==100 and pixels 0, 25, 50, 75 and 100. ImageMagick coverts it to map100_raw.pgm, but sets maxval==123 so all pixels had to be rescaled and after that they were rounded to nearest integer value, so those images are not exactly the same!

MarkCallow:
Yes Finder interpolates and filters giving a nice gradient.

max100_raw.pgm does not appear to have come from max100.pgm. max100.pgm has the values `0, 25, 50, 75, 100`, (which, if you are expecting a smooth gradient, are sRGB values). max100_raw.pgm has values `0, 32, 64, 95, 128`. max100.ktx has values `64, 129, 191, 255` which are the correct values when scaling the content of max100.pgm by 255/100.

I haven't looked at max1000 yet, but I expect I'll find a similar issue.

Could you remake the files with the correct values and, in the case of `max[12]00`, preferably with a bt709 OETF and truly linear for `max1000` and `max40000`. Do you know if files with maxval > 255 are supposed to be in bt709 space or linear? For maxval >= 2<sup>14</sup> there is no need for non-linear encoding.

kacprzak:
After reading this http://netpbm.sourceforge.net/doc/pgm.html multiple times I would say that 2 byte files should also be nonlinear. This is what GIMP does when saving an 16-bit image but it uses sRGB instead bt709. More about this here:
https://stackoverflow.com/questions/56019321/understanding-netpbms-pnm-nonlinear-rgb-color-space-for-converting-to-grayscale

I created some more files. Maybe they will be of any use for you. Since I cannot trust any image editing software when working with this format I created ascii (P2) files using script with consecutive values in every pixel. Then I used pnmgamma to convert them from linearspace to bt709. Details in `sh` file. This preserves maxval so images with maxval==63, and maxval==16383 will need to be rescaled when converting to KTX.
[pgm_test.zip](https://github.com/KhronosGroup/KTX-Software/files/5246447/pgm_test.zip)

PR #324, opened 2020-10-07
KTX read-only library

atteneder:
Added an additional library target `ktx_read` for users who only want to read KTX files. Especially useful if binary size is a concern.

- Had to restructure the CMake config a bit to avoid redundancies.
- Added an option to force creation of static libraries. I needed this to create a .bundle library for Unity macOS.

At the moment this additional lib is built by default. We could make options to en-/disable it (also for the default lib).

Thanks!


lib/texture1.c
```diff
@@ -1323,11 +1323,19 @@ struct ktxTexture_vtbl ktxTexture1_vtbl = {
...
     (PFNKTEXSETIMAGEFROMSTDIOSTREAM)ktxTexture1_SetImageFromStdioStream,
     (PFNKTEXWRITETOSTDIOSTREAM)ktxTexture1_WriteToStdioStream,
     (PFNKTEXWRITETONAMEDFILE)ktxTexture1_WriteToNamedFile,
     (PFNKTEXWRITETOMEMORY)ktxTexture1_WriteToMemory,
+#else
+    (PFNKTEXSETIMAGEFROMMEMORY)NULL,
+    (PFNKTEXSETIMAGEFROMSTDIOSTREAM)NULL,
+    (PFNKTEXWRITETOSTDIOSTREAM)NULL,
+    (PFNKTEXWRITETONAMEDFILE)NULL,
+    (PFNKTEXWRITETOMEMORY)NULL,
```
MarkCallow, line 1337: Instead of NULL and therefore a crash, these should point at a stub function that returns KTX_INVALID_OPERATION.

lib/texture2.c
```diff
@@ -2069,11 +2069,19 @@ struct ktxTexture_vtbl ktxTexture2_vtbl = {
...
     (PFNKTEXSETIMAGEFROMSTDIOSTREAM)ktxTexture2_SetImageFromStdioStream,
     (PFNKTEXWRITETOSTDIOSTREAM)ktxTexture2_WriteToStdioStream,
     (PFNKTEXWRITETONAMEDFILE)ktxTexture2_WriteToNamedFile,
     (PFNKTEXWRITETOMEMORY)ktxTexture2_WriteToMemory,
+#else
+    (PFNKTEXSETIMAGEFROMMEMORY)NULL,
+    (PFNKTEXSETIMAGEFROMSTDIOSTREAM)NULL,
+    (PFNKTEXWRITETOSTDIOSTREAM)NULL,
+    (PFNKTEXWRITETONAMEDFILE)NULL,
+    (PFNKTEXWRITETOMEMORY)NULL,
```
MarkCallow, line 2083: Same thing, stub returning KTX_INVALID_OPERATION.

MarkCallow (changes requested):
I'd prefer if the additional library was only built when an option is set to build the read-only library. However it should always be built by CI so CI should set the option.

CMakeLists.txt
```diff
@@ -27,6 +27,7 @@ include(cmake/xcode.cmake)
 CMAKE_DEPENDENT_OPTION(KTX_FEATURE_TOOLS "Create KTX tools" ON "NOT IOS;NOT ANDROID;NOT EMSCRIPTEN" OFF)
 option( KTX_FEATURE_DOC "Create KTX documentation" OFF )
 option( KTX_FEATURE_LOADTEST_APPS "Create load tests apps that load and display various KTX textures" OFF )
+option( KTX_FEATURE_FORCE_STATIC "Always create static libraries" OFF )
```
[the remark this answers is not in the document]
MarkCallow: Perhaps it would be better for the static library option to always appear and be set and unchangeable on iOS and Emscripten, unset and changeable everywhere else. This would show people that the library on those platforms is static, something hidden at present. If we do this the pop-up help should explain why it may be unchangeable.

What do you think @atteneder?

CMakeLists.txt
```diff
@@ -27,7 +27,7 @@ include(cmake/xcode.cmake)
 CMAKE_DEPENDENT_OPTION(KTX_FEATURE_TOOLS "Create KTX tools" ON "NOT IOS;NOT ANDROID;NOT EMSCRIPTEN" OFF)
 option( KTX_FEATURE_DOC "Create KTX documentation" OFF )
 option( KTX_FEATURE_LOADTEST_APPS "Create load tests apps that load and display various KTX textures" OFF )
-option( KTX_FEATURE_FORCE_STATIC "Always create static libraries" OFF )
+CMAKE_DEPENDENT_OPTION(KTX_FEATURE_STATIC_LIBRARY "Create static libraries (shared otherwise)" OFF "NOT IOS;NOT EMSCRIPTEN" ON)
```
MarkCallow, line 30: `NOT IOS; NOT EMPSCRIPTEN`? It looks like this should be setting `KTX_FEATURE_STATIC_LIBRARY` ON for IOS and EMSCRIPTEN so why the NOT?

lib/texture1.c
```diff
@@ -1303,6 +1303,49 @@ ktxTexture1_NeedsTranscoding(ktxTexture1* This)
...
+
+KTX_error_code
+ktxTexture1_WriteToMemory(ktxTexture1* This,
+                          ktx_uint8_t** ppDstBytes, ktx_size_t* pSize)
+{
+    return KTX_INVALID_OPERATION;
+}
+
+#endif
+
```
MarkCallow, line 1348: I think we should remove the parameter names from these to indicate to pedantic compilers that the parameters are unused. Actually, I'm not sure this works for C, perhaps it is only C++.

CMakeLists.txt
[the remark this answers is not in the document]
atteneder: Plural (libraries), because ktx_read is added with this changes making it two libraries.
I've updated this to an option, that's not available on iOS/Emscripten.

I can see the benefit of having the static option always there, but I don't think you can make it unchangeable (as in greyed-out in CMake GUI). A fallback could be to correct invalid configurations (e.g. OFF on iOS) by throwing a warning or an error. I'd prefer an error (warnings may not be read and can lead to confusion).

I have no strong opinion on whether the static option should be present on all platforms or not.

MarkCallow (changes requested):

CMakeLists.txt
```diff
@@ -146,126 +145,228 @@ add_library( ktx ${LIB_TYPE}
...
+        find_package(zstd REQUIRED)
+        find_package(Threads REQUIRED)
+        target_link_libraries(
+            ${lib}
+            dl
+            Threads::Threads
+            zstd::zstd
+        )
+    elseif(ANDROID)
+        # add_library(zstd STATIC IMPORTED)
```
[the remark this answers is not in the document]
atteneder: I did some initial testing a while ago and want to finish it soon.

The CMake code isn't ready though and must have slipped in somehow.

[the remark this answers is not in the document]
atteneder: Should I remove the Android tests until they are usable or is it enough to not list Android as supported platform for now ?

CMakeLists.txt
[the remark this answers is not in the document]
MarkCallow: If you can't disable the option in the GUI then its better to just not have it  on iOS and Emscripten.

CMakeLists.txt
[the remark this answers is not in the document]
MarkCallow: I think it is enough to not list Android as supported.

lib/texture1.c
atteneder: According to this SO it is C++ specific:
https://stackoverflow.com/questions/3487135/unnamed-parameters-in-c

So i guess even pedantic compilers shouldn't complain, right?

atteneder:
The tools `ktx2check` and `ktxinfo` should also work if linking against `ktx_read` (instead of `ktx`). I tried and the tests did not break.

@MarkCallow Would you approve that?

CMakeLists.txt
atteneder: Yeah, the makro is a bit confusing to decipher. In words this means:
"if `NOT IOS` AND `NOT EMSCRIPTEN` enable the option with the default `OFF`. Set it to `ON` otherwise."

atteneder: See `KTX_FEATURE_TOOLS` above for comparison

MarkCallow: Thanks for the explanation.

lib/texture1.c
MarkCallow: We'll see.

MarkCallow:
CI should certainly build `ktx_read` and link `ktxinfo` and `ktx2check` against it. However during development I don't want to be building 2 libraries so I'd like the default build to be the `ktx` with the tools linking against its `libktx`.

Do you think we should include the read only library in the installers? Even if we do, these 2 tools should be linked against the full `libktx` to avoid having 2 libraries resident in memory.

MarkCallow (approved):

MarkCallow:
I've changed my mind about not having `KTX_FEATURE_STATIC_LIBRARY` appearing for iOS and Emscripten. Leave it the way it is now but please add an error should it be set false on these 2 platforms. After that I'll merge this change.

atteneder:
> CI should certainly build `ktx_read` and link `ktxinfo` and `ktx2check` against it. However during development I don't want to be building 2 libraries so I'd like the default build to be the `ktx` with the tools linking against its `libktx`.
> 
> Do you think we should include the read only library in the installers? Even if we do, these 2 tools should be linked against the full `libktx` to avoid having 2 libraries resident in memory.

My thought was saving memory by using the smaller ktx_read, but in that case this does not make sense.

atteneder:
Alright, `KTX_FEATURE_STATIC_LIBRARY` is now always present, but a proper error is shown in case it is misconfigured.

I quickly tested if a read-only WebAssembly variant makes sense, but the current version already offers no write capability (thus it's already very small).

MarkCallow:
> I quickly tested if a read-only WebAssembly variant makes sense, but the current version already offers no write capability (thus it's already very small).

I do expect to add write capabilities to the WebAssembly build in the future so people can write web tools to encode files. Please check that the Emscripten build disables writing in the same way as ktx_read, i.e. using the same compile options. However do not change the name of the Emscripten library at this time as I do not want to confuse existing users. We'll need to provide warning of the change.

atteneder:
I now link `ktx_js` and `msc_basis_transcoder_js` against `ktx_read`.

Once write functionality in web is required, we should create additional targets and (optionally) set the `KTX_FEATURE_WRITE` compile flag on those.

MarkCallow:
I cannot find any compile option being set to omit write from the Emscripten build. I think this functionality is missing from the .wasm module just because the wrapper does not have any write functions so they're never linked from libktx. Basically an example of why a static read-only library is not needed. When we make a .wasm module that supports write, at that time we must use `KTX_FEATURE_WRITE` to enable building of a read-only wrapper.

For now I think this PR is good to go.

PR #334, opened 2020-10-18
Fix basisu C binding

atteneder:
Most functions of the BasisU C binding were not included previously due to a missing definition.

Reminder: Those are useful for creating apps/libs that can parse .basis file as well as .ktx files.


MarkCallow:
Please fix the build failures here before we spent any more time on Android support.

atteneder:
@MarkCallow I fixed the build. It's ready now!

PR #335, opened 2020-10-18
Android Build

atteneder:
Consider this a starting point for Android builds.

The only adjustment required was to make Zstd available.

Instead of including zstd binary builds for all Android platforms, I created a setting (`ZSTD_PATH`) pointing to the sources of [ZStandard](https://github.com/facebook/zstd)). If set properly, it loads its CMake configuration and directly links it.

Also provided a build shell script as a starting point for CI/CD (currently arm64-v8a only). Use it for testing or as a reference how to configure for Android.

Open questions:

- Is this a good way of including Zstd?
  - if yes: Should we do this for other platforms as well? Should we include zstd as git submodule/subrepo?
  - if no: should we add binary builds (debug+release) for all android ABIs (armeabi-v7a, arm64-v8a, x86, x86_64 ) ?

- CI/CD is missing. Should we add this to TravisCI / Linux ?


MarkCallow:
> Is this a good way of including Zstd?

I think you are more knowledgeable about this than me. What are the pros & cons? With the method you are using for Android does the user have to clone the Zstd repo then point ZSTD_PATH at the clone? One small drawback I see to submodule/subrepo is that on some platforms (iOS, Web and probably Android too) we only want the decoder. If Zstd for these platforms is a static library then the linker will solve this problem. If it is a shared library some extra work will be needed in the build files.

> CI/CD?

We should add this to the build for whichever platform has the most reliable and up to date Android NDK.

atteneder:
I try to sum up options plus pros/cons:

### Option A (current): User-provided Zstd clone via ZSTD_PATH

Pro:

- Easy to integrate, since Zstd comes with a CMake config

Cons:

- Extra step in setup (user has to clone Zstd and set path)
- Exact Version of Zstd is out of control

### Option B: Add zstd as static library binaries (just like with SDL2, etc.)

Pro:

- No extra configuration (for users)
- Precise control over Zstd version

Con:

- Additional weight on repo (~16MB for release only libs + ~50MB for debug libs)
- Requires more platform specific CMake script

### Option C: Add Zstd sources as sub-repo (just like basisu)

Pro:

- Easy to integrate, since Zstd comes with a CMake config
- No extra path configuration (for users; like in Option A)
- Precise control over Zstd version
- Requires less space than Option B ( ~6.5 MB )
- We could remove/replace other zstd static libs/binaries with this as well.

Con:

- Zstd is built from source -> takes slighlty longer

### Option D: Use `lib/zstddeclib.c` for Android (just like for Web)

Pro:

- No need for external/additional dependecies

Con:

- No write support (we could probably build `ktx_read` only in that case.)

Would be an option for iOS as well.

---

My opinion:

I'd go for Option D (if write support is not a topic on Android/Web/iOS), Option C otherwise.

atteneder:
@MarkCallow Let me know which Option you prefer. I try to implement it asap.

MarkCallow:
I can't see write support being a topic on iOS or Android. I don't think people will want to create textures there. It may become interesting for the web though in case someone wants to provide a service to encode .ktx2 files.

Let's use option C for Linux, macOS, Windows and Web (when building libktx). If it's not too difficult to run the script to create `zstddeclib.c`  from the zstd sources we could use it for Android and iOS as well. Otherwise let's use the pre-built `zstddeclib.c` (option D) for those platforms.

atteneder:
@MarkCallow the latest build fails on linux due to license issue:

```
# SUMMARY

* Bad licenses:

* Deprecated licenses:

* Licenses without file extension:

* Missing licenses: GPL-2.0, GPL-2.0-only

* Unused licenses:

* Used licenses: Apache-2.0, BSD-1-Clause, BSD-2-Clause, BSD-3-Clause, BSL-1.0, CC-BY-3.0, CC0-1.0, GPL-2.0-or-later, LicenseRef-Cesium-Trademark-Terms, LicenseRef-ETCSLA, LicenseRef-HI-Trademark, LicenseRef-Kodak, LicenseRef-PNGSuite, MIT, Zlib

* Read errors: 0

* Files with copyright information: 1676 / 1835

* Files with license information: 1383 / 1835
```

Do we have a legal issue here?
If it's a build-config issue, please guide me how to fix it.

Thanks!

MarkCallow:
This is output from `reuse link` checking for compliance with Reuse. It indicates that the directory `LICENSES` has no entry for GPL-2.0 and GPL-2.0-only. Presumably these are used by Zstd. There may be a legal issue as well. I've have to check with Khronos whether it is okay to include GPL licensed code in this repo. Actually the `GPL-2.0-only` make me wonder if parts of Zstd are dual licensed. I'll have to take a look.

MarkCallow:
The code is dual licensed with BSD & GPLv2. The GPL-only files are in the `contrib` folder and are for integrating zstd into the Linux kernel so we won't be using those. I need to ask someone how to handle this situation so we can remain Reuse compliant.

MarkCallow:
Sorry for the delay in responding. I've been gathering information.

The basic fix needed for reuse is to add an entry to `.reuse/dep5` for the files in the zstd repo. Something like

```
Files: lib/zstd/*
Copyright: Copyright (c) 2016-present, Facebook, Inc.
# We choose to use the BSD license.
License: BSD-3-Clause
```

Unfortunately this will not prevent `reuse` looking in files for SPDX license identifiers so it will still report those files from `zstd/contrib` that have identifiers saying they are GLPv2. We do not want to confuse people using our repo by making them think part of the s/w is GPLv2 licensed. The best workaround we've come up with is to replace the files with stubs after making the subrepo. This should only cause issues in future git subrepo pulls if those files have been changed upstream. There are 4 files. They can be found with

```
find lib/zstd/contrib/linux-kernel -name '*' -type f -exec grep -H SPDX {} \;
```

Another suggestion is to try the sparse clone support in git so as to omit this directory from the repo. I have not investigated whether such is possible with `git subrepo`. Please take a look as it would be a preferable solution.

atteneder:
Thanks for the reply!

Unfortunately git subrepo [does not support partial clones](https://github.com/ingydotnet/git-subrepo/issues/454)

Stubs is a solution. We could also plain remove the linux-kernel directory (have to check if CMake still works then) and add it to the main .gitignore to prevent repetitive trouble there. I'd have to test that.

MarkCallow:
> We could also plain remove the linux-kernel directory (have to check if CMake still works then) and add it to the main .gitignore to prevent repetitive trouble there. I'd have to test that.

Please test it. I suspect having files from the subrepo in .gitignore will confuse `git subrepo` which is why I did not propose it. It is worth testing though.

atteneder:
Hi @MarkCallow ,

I tested like so:
- removed said `linux-kernel` folder
- added it to .gitignore
- pulled the sub-repo

result: the folder was not re-added and no errors whatsoever

MarkCallow:
Sounds good. Does this PR remove the pre-built zstd libraries and use the subrepo for all platforms?

atteneder:
> Does this PR remove the pre-built zstd libraries and use the subrepo for all platforms?

Yes! But it still uses `lib/zstddeclib.c` for target `ktx_read`

MarkCallow:
Please make a target that builds `zstddeclib.c` if the zstd sources are newer and make the targets that use `zstddeclib.c` dependent on this new target. See `lib/zstd/contrib/single_file_libs/README.md` for details on how to create `zstddeclib.c`

atteneder:
> 
> 
> Please make a target that builds `zstddeclib.c` if the zstd sources are newer and make the targets that use `zstddeclib.c` dependent on this new target. See `lib/zstd/contrib/single_file_libs/README.md` for details on how to create `zstddeclib.c`

done!

MarkCallow:
For some reason GitHub has not run the CI build for this PR. If it had,`reuse lint` still fails. There is no entry in `.reuse/dep5` for lib/zstd. Please add it and figure out why CI is not being run for this PR.

MarkCallow:
I modified `.reuse/dep5`. [dep5.txt](https://github.com/KhronosGroup/KTX-Software/files/5545482/dep5.txt) (Be sure to rename this. I had to add a .txt extension to get GitHub to upload it).

While that fixes lib/zstd, there are some new zstd-related files that are missing information:

```
The following files have no copyright and licensing information:
* ci_scripts/build_android.sh
* cmake/zstddeclib.cmake
* tests/webgl/zstddec.wasm

The following files have no licensing information:
* tests/webgl/zstddec.js
```

You will need to add the missing information.

MarkCallow:
A couple of other issues:

1. There are some zstd header files in `other_include`. Presumably these are no longer needed and can be deleted. If they are deleted, the `dep5` file I attached in an earlier comment needs modifying to remove these files.

2. This PR is not adding the Android build to CI. It does not change either .appveyor.yml or .travis.yml and there appears to be nothing calling `ci_scripts/build_android.sh`.

MarkCallow:
Please push a change to this PR to trigger a build. I need to test the fix I made for Travis builds not being triggered.

MarkCallow:
The missing Travis PR builds has something to do with Khronos's Travis membership. It will probably take a few days to sort out. In the meantime, please fix the other issues I pointed out.

atteneder:
@MarkCallow I fixed reuse lint removed the obsolete zstd headers and fixed the build afterwards.

I had a look into Android CI, read the Travis CI docs a bit and have a suggestion:
Why not split up into more jobs?
Why not have per-job lifecycle steps (install, before-script, script etc.) with common code in ci-script/* files?

Also strange: We have multiple `language` entries, which does not seem to be supported. A way to do this is via job matrix (see [this SO](https://stackoverflow.com/a/44054333)).

Status quo:

- Job 1: Linux
  - Linux (3 package formats)
  - Web (2 libraries; via docker)
- Job 2: macOS
  - macOS
  - iOS
  - docs/pages

Suggestion:

- Job 1: Linux
  - Linux (3 package formats)
- Job 2: macOS
  - macOS
  - iOS
  - docs/pages
- Job 3:
  - Web (2 libraries; via docker)
- Job 4:
  - Android

Lastly: I already have GitHub action based CI up for [KTX-Software-Unity](https://github.com/atteneder/KTX-Software-Unity/blob/main/.github/workflows/main.yml) (a lib that consumes this project) for win/linux/macos/android/ios/web/uwp and to me it seems to be less complex (granted, it does not have all the same features). I'm not implying to move everything, but if you're open to it, we could use it for Android for starters.

MarkCallow:
The Windows build is failing because it is still looking in other_lib for the zstd library.

I'm fine with creating more jobs as proposed. I don't understand the suggestion for "per-job lifecycle steps". Please explain. 

On the multiple languages, I don't recall adding node_js. I think originally I had gcc and clang but then commented out clang because that is what is used on the macOS build so I didn't see any need to also run a clang build on Linux. However I have no opinion. Please fix as you see fit.

I agree to try GitHub actions for the Android build.

Please make these build changes in a new PR. I don't think we should add them to this.

MarkCallow:
> Please make these build changes in a new PR. I don't think we should add them to this.

Change of mind. Please add the GitHub action for the Android build to this PR.

atteneder:
> Change of mind. Please add the GitHub action for the Android build to this PR.

Sure. I won't be able to do so before mid of next week though.

Do you know why appveyor fails? Does it trap over a CMake configuration warning message?

MarkCallow:
> Do you know why appveyor fails? Does it trap over a CMake configuration warning message?

Can you not see the Appveyor logs? I thought they were all public for open source projects. Here is one of them. It seems to be failing because CMake did not find dot. It's only a warning I think but any output to standard error makes PowerShell think there has been an error. I think it does this because Windows tools are not consistent about  exiting with non-zero exit codes on error.

```
Build via Visual Studio 15 2017 Arch:x64 dir:build-win-Release
-- Found Bash: C:/Program Files/Git/bin/bash.exe  
-- The C compiler identification is MSVC 19.16.27035.0
-- The CXX compiler identification is MSVC 19.16.27035.0
-- Check for working C compiler: C:/Program Files (x86)/Microsoft Visual Studio/2017/Community/VC/Tools/MSVC/14.16.27023/bin/Hostx86/x64/cl.exe
-- Check for working C compiler: C:/Program Files (x86)/Microsoft Visual Studio/2017/Community/VC/Tools/MSVC/14.16.27023/bin/Hostx86/x64/cl.exe -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: C:/Program Files (x86)/Microsoft Visual Studio/2017/Community/VC/Tools/MSVC/14.16.27023/bin/Hostx86/x64/cl.exe
-- Check for working CXX compiler: C:/Program Files (x86)/Microsoft Visual Studio/2017/Community/VC/Tools/MSVC/14.16.27023/bin/Hostx86/x64/cl.exe -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found Perl: C:/Perl/bin/perl.exe (found version "5.24.1") 
-- Could NOT find zstd (missing: zstd_LIBRARIES zstd_INCLUDE_DIRS) 
-- ZSTD VERSION: 1.4.6
-- Performing Test C_FLAG_MP
-- Performing Test C_FLAG_MP - Success
-- Performing Test CXX_FLAG_MP
-- Performing Test CXX_FLAG_MP - Success
-- Performing Test C_FLAG_D_UNICODE
-- Performing Test C_FLAG_D_UNICODE - Success
-- Performing Test CXX_FLAG_D_UNICODE
-- Performing Test CXX_FLAG_D_UNICODE - Success
-- Performing Test C_FLAG_DUNICODE
-- Performing Test C_FLAG_DUNICODE - Success
-- Performing Test CXX_FLAG_DUNICODE
-- Performing Test CXX_FLAG_DUNICODE - Success
-- CMAKE_INSTALL_PREFIX: C:/Program Files (x86)/KTX-Software
-- CMAKE_INSTALL_LIBDIR: lib
-- ZSTD_LEGACY_SUPPORT not defined!
-- ZSTD_MULTITHREAD_SUPPORT is enabled
-- Found OpenGL: opengl32   
-- Found Vulkan: C:/VulkanSDK/1.2.141.2/Lib/vulkan-1.lib  
-- Looking for pthread.h
-- Looking for pthread.h - not found
-- Found Threads: TRUE  
-- Found Doxygen: C:/Program Files/doxygen/bin/doxygen.exe (found version "1.8.20 (f246dd2f1c58eea39ea3f50c108019e4d4137bd5)") found components: doxygen missing components: dot
-- Configuring done
-- Generating done
The running command stopped because the preference variable "ErrorActionPreference" or common parameter is set to Stop: CMake Warning:
```

atteneder:
@MarkCallow 

Yes, I could read the log, but I wasn't sure what was causing the problem. Thanks for clarifying.

MarkCallow:
> I removed all CMake warnings and still get the error. Ideas?

The string after "stop" in the ErrorAction message is the string seen on stderr, if I'm not mistaken. So it seems CMake is still writing "CMake Warning" to stderr. Could it be related to not finding `pthread.h`?

atteneder:
Good spot! May very well be the case. I'll be able to continue in a couple of days.

MarkCallow:
The iOS, Linux & macOS builds are all still failing. See [build 1435](https://travis-ci.com/github/KhronosGroup/KTX-Software/builds/203238559). In the Linux build it looks like writer2.h is not including zstd.h. I keep getting caught by this kind of thing as Xcode unhelpfully searches your entire project for header files. In the macOS build zstd.h is not being found when texture2.c is being compiled.

atteneder:
@MarkCallow : Got one step further, but now the unittests are not built. Maybe it's because of the prior errors building the documenation (dot fails). The suspicious log:

```
If you installed Graphviz/dot after a previous failing run, 
  try deleting the output directory and rerun doxygen.
```

That applies to us. Is the build directory cached and can we wipe it somehow?

thanks

MarkCallow:
From the logs it looks like nothing but docs builds are ever attempted. I have no idea why.  Building of unittests is completely independent of the doxygen build. There shouldn't be any dependency on docs. Despite the errors about not finding map files, VS should build the other targets. Even when there is a failure compiling a target which is a dependency of other targets VS does not stop. For example if the libktx build fails the first the CI knows about it is when the tests fail to run because they don't exist because they couldn't be linked successfully.

There is no build cache.

Why is the build trying to run something called `unittests_NOT_BUILT.exe`? This sounds like a configuration problem. If a build failure was causing the unittests build to not start or not finish, I'd expect the build to still try to run `unittests.exe` and tell us it can't find that.

We have another problem, Travis CI has once again started ignoring requests so we only have a Windows build for these latest commits.

atteneder:
> From the logs it looks like nothing but docs builds are ever attempted.

Good point.

> I have no idea why. Building of unittests is completely independent of the doxygen build. There shouldn't be any dependency on docs.

True. There is none, as far as I can see.

> Despite the errors about not finding map files, VS should build the other targets. Even when there is a failure compiling a target which is a dependency of other targets VS does not stop. For example if the libktx build fails the first the CI knows about it is when the tests fail to run because they don't exist because they couldn't be linked successfully.

Seems like that's not the case and I don't have access to Windows to test it. Building the docs should work anyways, but I'll disable it on Windows for now to see if the rest of the build works.

> There is no build cache.

Thanks for clarifying.

> Why is the build trying to run something called `unittests_NOT_BUILT.exe`?

Seems like (among most other targets) unittests was not built.

> We have another problem, Travis CI has once again started ignoring requests so we only have a Windows build for these latest commits.

After more already working things broke again, I like this whole CI setup less and less :)

MarkCallow:
>> Why is the build trying to run something called unittests_NOT_BUILT.exe?

> Seems like (among most other targets) unittests was not built.

I've only seen `NOT_BUILT` before for things that are no configured to build. If a build gets a error, the name of the target is not changed. Hence my question. It looks like unittests are not configured to be built.

MarkCallow:
The problem blocking the macOS & Linux builds has reportedly been fixed, however redelivering GitHub webhook payloads to Travis has no visible effect. Please push another commit to this PR to trigger a build. Please do it as soon as possible.

MarkCallow:
I managed to get another Travis build to run on this PR. For some reason it doesn't show up in the checks here. Please see https://travis-ci.com/github/KhronosGroup/KTX-Software/jobs/445056506 for the log.

The Linux build is failing because

```
../lib/writer2.c:701:20: error: ‘ZSTD_error_workSpace_tooSmall’ undeclared (first use in this function); did you mean ‘ZSTD_error_dstSize_tooSmall’?

               case ZSTD_error_workSpace_tooSmall:

                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~

                    ZSTD_error_dstSize_tooSmall
```

I'm pretty sure this is because the compiler isn't finding zstd_errors.h though why the compiler isn't reporting that I do not know. I'm not entirelly sure of the reason for not finding it. `lib/zstd/lib` does not appear to be in the `target_include_directories` for the `ktx` or `ktx_read` targets. Even if it is and I'm missing it, `zstd_errors.h` is in  `lib/zstd/lib/common`. This problem must surely be affecting Windows as well. I have no idea why we aren't even seeing a library build attempted.

The macOS build is failing because

```
In file included from /Users/travis/build/KhronosGroup/KTX-Software/tools/ktxsc/ktxsc.cpp:19:

/Users/travis/build/KhronosGroup/KTX-Software/utils/scapp.h:7:10: fatal error: 'zstd.h' file not found

#include <zstd.h>

         ^~~~~~~~

1 error generated.

** BUILD FAILED **
```
So same reason. But note that this is a different target. Somehow the library is building for macOS. I suspect this is because Xcode helpfully searches all the directories in a project (?) or target (?) in which case I'm not sure why it isn't being found by scapp.h.

MarkCallow:
@atteneder please rebase and fix the failure to find the zstd header files.

atteneder:
After fixing some configuration errors I'm running into what seems to be random CI errors.
I'm starting to think prepone switching to GH actions makes more sense. It's not fun fixing those with iteration times of 30 minutes and more.
@MarkCallow What do you think?

MarkCallow:
> I'm running into what seems to be random CI errors

If by this you mean the appveyor failures, the issue with fetching stuff from the chocolatey server has been ongoing for a long time. A "Rerun incomplete" is usually successful. I've just started one now.

> I'm starting to think prepone switching to GH actions makes more sense. It's not fun fixing those with iteration times of 30 minutes and more.

I want to get the Android PR merged as soon as possible so prefer to delay switching to GH actions until that is done. I don't know how long it might take to get those working. I absolutely agree about the 30 minute turnaround being a horrendous problem. Hopefully GH actions will be faster.

MarkCallow:
I expect you can reproduce the macOS build failure running the test `toktx-cmp-cimg5293_uastc_zstd` locally.

atteneder:
Hi @MarkCallow,
finally got time to get back to it.
Test mentioned failed because of differences in the resulting ktx2 files. Turns out I replaced the pre-built single file zstd (1.4.5) with a newer version (1.4.6) which was binary incompatible. I changed the new zlib subrepo to tag v1.4.5 which solved the problem. I leave it up to you to upgrade zstd and update the ktx2 reference files eventually someday.
pushing update soon!

atteneder:
Next hurdle:

`The job exceeded the maximum log length, and has been terminated.`

Getting really tired of this CI setup :/

MarkCallow:
@atteneder can you finish this please. The Vulkan samples people are planning to use KTX v2 and they need Android support.

Since you last looked at this, BasisU has added KTX v2 support which led them to include zstd in their repo and, by exgtension,  in our subrepo clone. I have modified the builds to use the single file encoders and decoders from `lib/basisu/zstd` and have removed the pre-built zstd libs. There are still `other_include/zstd{,_errors}.h`. I plan to remove the former but will have to keep zstd_errors.h as libktx needs the declarations in order to handle potential errors. You should be able to use the same zstd source for the Android build. I regenerated all the reference files using this zstd.

MarkCallow:
I've landed a PR to remove `other_include/zstd.h`.

atteneder:
Hi @MarkCallow ,
I'll try to rebase it this week.

Congratulations on the 4.0.0 release!    🎉

atteneder:
The only main problem for Android builds to work was include zstd, which was done in the meantime. This PR became out of date...will create another one.

PR #348, opened 2020-12-21
fix: Xcode's build system (non-legacy) requires a direct dependency t…

atteneder:
…o source generating targets.


MarkCallow:
Please make some trivial change to trigger another build. The macOS build credits on Travis CI had run out. They've been replenished so this time a build should be done.

PR #391, opened 2021-04-08
Add ktxTexture2_Destroy and ktxTexture2_GetImageOffset

ifiddynine:
Examples for KTX2 that I found used these two methods that weren't provided.

Let me know if there is another way to destroy and get offset.

### DIFF
--- a/include/ktx.h
+++ b/include/ktx.h
@@ -917,6 +917,14 @@ ktxTexture2_GetPremultipliedAlpha(ktxTexture2* This);
 KTX_API ktx_bool_t KTX_APIENTRY
 ktxTexture2_NeedsTranscoding(ktxTexture2* This);
 
+KTX_API void ktxTexture2_Destroy(ktxTexture2* This);
+
+KTX_API KTX_error_code KTX_APIENTRY
+ktxTexture2_GetImageOffset(ktxTexture2* This, ktx_uint32_t level,
+                            ktx_uint32_t layer, ktx_uint32_t faceSlice,
+                            ktx_size_t* pOffset);
+
+
 /**
  * @~English
  * @brief Flags specifiying UASTC encoding options.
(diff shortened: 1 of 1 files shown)


MarkCallow:
You can use the macros wrapper around the virtual functions: `ktxTexture_Destroy` and `ktxTexture_GetImageOffset`. Where did you find the examples you mention? I'd rather fix those than expose the ktx2 versions. I'd also have to expose the ktx1 versions for consistency. Basically if the function exists on both ktxTexture1 and ktxTexture2 use the virtual function wrapper. Only if the function is unique to one or the other do you need a ktxTexture2 or ktxTexture1 prefix.

ifiddynine:
Ah cool. The examples were found here:  https://github.khronos.org/KTX-Software/libktx/index.html#writektx. Under the section "Writing a BasisLZ/ETC1S-compressed Texture".

ktxTexture_Destroy is a bit confusing when using the KTX2 API though, since by default `ktxTexture_Destroy(texture)` where `texture` is a `ktxTexture2` doesn't compile. Having to cast to `ktxTexture` wasn't clear to me. Maybe there's something that can be done in the backend, i.e. adding implicit operators or an non-macro overload for `ktxTexture_Destroy` to remove the need to cast?

Feel free to close the PR when all is well, I had the wrong git config locally either way so wouldn't be able to sign the CLA with those commits in.

MarkCallow:
I see the `ktxTexture2_Destroy` call but I don't see a `ktxTexture2_GetImageOffset` call in ""Writing a BasisLZ/ETC1S-compressed Texture" or any other section of that file. Please verify that is the place you saw it. I am  just about to push a fix for the example's Destroy call.

MarkCallow:
I'm not going to take this PR. I will look again to see if there is a way to provide a cast in the macro without throwing away compiler type checking completely. Unfortunately this is a c API so name named functions with different arg types and implicit operators are not possible. Thanks again for pointing out the error in the example.

Issue #390, opened 2021-04-08
Fix face bytelength calculation for PVRTC1


lexaknyazev:
https://github.com/KhronosGroup/KTX-Software/blob/4e0a2cea60431ef757d7e7d93b245fc074c60e19/lib/texture2.c#L2018-L2019

From the PVRTC1 spec:
> For PVRTC 4BPP formats the imageSize is calculated as:
> `(max(width, 8) * max(height, 8) * 4 + 7) / 8`
> For PVRTC 2BPP formats the imageSize is calculated as:
> `(max(width, 16) * max(height, 8) * 2 + 7) / 8`

MarkCallow:
Are width & height in pixels? If so, this seems a really strange calculation as it easily results in a value that is not a multiple of the block size.

lexaknyazev:
This calculation limits the minimum payload size because any PVRTC1 texture must comprise at least two words of PVRTC1 data in each dimension.

lexaknyazev:
Additionally, the dimensions of PVRTC1 data must always be powers-of-two, so the result is always a multiple of 8.

MarkCallow:
Ah! yes. I was forgetting about the powers-of-two requirement.

MarkCallow:
I guess I shouldn't have put "issue" between Fixes and the number since the auto close failed.

PR #398, opened 2021-04-15
Handle PVRTC minimum 2 block requirement. Fixes issue #390.

MarkCallow:
Note that the key thing about PVRTC1 is that there must be at least 2 texel blocks in each dimension. That is what was missing. The formula in the GL EXT spec. for the image size is expressed bizarrely, especially the + 7 part for rounding up which is unnecessary given the image sizes have to be a power of 2. The calculation after this PR is equivalent.

### DIFF
--- a/lib/texture2.c
+++ b/lib/texture2.c
@@ -235,9 +235,13 @@ ktxFormatSize_initFromDfd(ktxFormatSize* This, ktx_uint32_t* pDfd)
     This->blockSizeInBits = KHR_DFDVAL(pBdb, BYTESPLANE0) * 8;
     This->paletteSizeInBits = 0; // No paletted formats in ktx v2.
     This->flags = 0;
+    This->minBlocksX = This->minBlocksY = 1;
     if (KHR_DFDVAL(pBdb, MODEL) >= KHR_DF_MODEL_DXT1A) {
         // A block compressed format. Entire block is a single sample.
         This->flags |= KTX_FORMAT_SIZE_COMPRESSED_BIT;
+        if (KHR_DFDVAL(pBdb, MODEL) >= KHR_DF_MODEL_PVRTC) {
+            This->minBlocksX = This->minBlocksY = 2;
+        }
     } else {
         // An uncompressed format.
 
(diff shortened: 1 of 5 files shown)


lib/texture2.c
```diff
@@ -235,9 +235,13 @@ ktxFormatSize_initFromDfd(ktxFormatSize* This, ktx_uint32_t* pDfd)
     This->blockSizeInBits = KHR_DFDVAL(pBdb, BYTESPLANE0) * 8;
     This->paletteSizeInBits = 0; // No paletted formats in ktx v2.
     This->flags = 0;
+    This->minBlocksX = This->minBlocksY = 1;
     if (KHR_DFDVAL(pBdb, MODEL) >= KHR_DF_MODEL_DXT1A) {
         // A block compressed format. Entire block is a single sample.
         This->flags |= KTX_FORMAT_SIZE_COMPRESSED_BIT;
+        if (KHR_DFDVAL(pBdb, MODEL) >= KHR_DF_MODEL_PVRTC) {
+            This->minBlocksX = This->minBlocksY = 2;
```
lexaknyazev, line 243: PVRTC2 doesn't seem to have this requirement.

MarkCallow: It is supposed to be == not >=. Fixed.

lib/vk_format.h
```diff
@@ -1207,6 +1208,19 @@ static inline void vkGetFormatSize( const VkFormat format, ktxFormatSize * pForm
 			pFormatSize->blockHeight = 4;
 			pFormatSize->blockDepth = 1;
 			break;
+		case VK_FORMAT_PVRTC1_2BPP_SRGB_BLOCK_IMG:
+		case VK_FORMAT_PVRTC1_2BPP_UNORM_BLOCK_IMG:
+        case VK_FORMAT_PVRTC2_2BPP_SRGB_BLOCK_IMG:
+        case VK_FORMAT_PVRTC2_2BPP_UNORM_BLOCK_IMG:
```
lexaknyazev, line 1214: Should these cases be split as well (also 4bpp options below)?

MarkCallow: Why? What do you mean?

lexaknyazev: PVRTC2 shouldn't get `minBlocksX = 2`.

MarkCallow: Fixed.

PR #399, opened 2021-04-15
Minor reformat

lexaknyazev:
Follow-up to #398.


joe198:
ok

joe198:
#399

MarkCallow:
Thanks. I hadn't realized the file still had tabs in it. All my editors are set up to use spaces instead of tabs.

lexaknyazev:
@MarkCallow 
Should we reformat all sources with spaces?

MarkCallow:
Yes I want to do that. Simply haven't had the time due to more pressing needs.

PR #408, opened 2021-05-01
Android build script

atteneder:
> Good news: Now that basisu's bundled zstd is used, building the libraries for Android works out of the box!

Builds libs `libktx` and `libktx_read` for Android (ABI arm64-v8a only atm).

Requires [Android NDK](https://developer.android.com/ndk) to be installed and the `ANDROID_NDK` environment variable pointing to its directory.

Works for local development and should work in CI/CD integration as well.


MarkCallow:
Please add the Android build to CI. Use Appveyor or GitHub Actions. Appveyor supports Linux now.

atteneder:
I'd just ZIP the binaries( `libktx_read.so` and `libktx.so`), right?
Which Android ABIs should it support?

I'd go with GH actions (as I've used them on other projects before; might be a starting point to convert other jobs as well). Can't promise when I'll have time though.

MarkCallow:
The Linux build is failing because you need to provide licensing information for `.github/workflows/main.yml`. The Appveyor build failure was due to an Appveyor problem. I am "re-run(ning) incomplete" there.

MarkCallow:
Can we add the GH action to the PR checks before I merge to master? I'd like to make sure it works before merge. Also still need instructions in BUILDING.md.

atteneder:
Please have a look at the successful [run of the GH action and the downloadable artifact ZIP](https://github.com/atteneder/KTX-Software/actions/runs/846018077).

> Note: The version is still off, I hope the next run fixes this.

Once merged, builds are triggered by PRs, releases and manually.

It contains, for each of the 4 Android ABIs (armv7,arm64,x86 and x86_64): Whatever `cmake install` would deliver, which is the `libktx.so`, the public header files and some CMake magic (for easier integration in CMake projects).

Remarks/Questions:

It builds both Debug and Release builds for all 4 ABIs . Would it be sufficient to build Release only or Debug only for the first ABI?

Notes not relevenat for / not blocking this PR
- git clone fetches all commits to include the latest release tag (slow, but required for versioning)
- docs are not included atm
- libktx_read is not included, since it's not installed (not in any platform actually)
- On [this line](https://github.com/KhronosGroup/KTX-Software/blob/877bca30f3c6eaca28628033180406f07cc34663/CMakeLists.txt#L42), it probably should say `x86_32` instead of `x86`, as that's what it's set to [here](https://github.com/KhronosGroup/KTX-Software/blob/877bca30f3c6eaca28628033180406f07cc34663/cmake/cputypetest.cmake#L113)

atteneder:
> Can we add the GH action to the PR checks before I merge to master? I'd like to make sure it works before merge.

I was hoping this is added automatically, but probably its not....I have to look how it's done tomorrow.

MarkCallow:
All builds are failing now due to an error in cputypetest at line 108. There's a STREQUALS with only one operand.

> It builds both Debug and Release builds for all 4 ABIs . Would it be sufficient to build Release only or Debug only for the first ABI?

The first ABI is probably sufficient unless we introduce ABI-specific #ifdefs.

> git clone fetches all commits to include the latest release tag (slow, but required for versioning)

There is not much we can do about this.

> docs are not included atm

That's fine. I don't expect they'll ever be included in any Android package.

> libktx_read is not included, since it's not installed (not in any platform actually)

That's fine too.

> On this line, it probably should say x86_32 instead of x86, as that's what it's set to here

I think you should change [this line](https://github.com/KhronosGroup/KTX-Software/blob/877bca30f3c6eaca28628033180406f07cc34663/cmake/cputypetest.cmake#L113 to `x86` as that is what is used in other cases in the test program). B.t.w we will have to modify cputypetest for the Apple M1 chip.

atteneder:
I implemented all feedback and started yet another test run.

Regarding checks on pull requests: After a small fix the action is included in PR checks. Apparently not in *this* PR, but in a [test-commit ](https://github.com/atteneder/KTX-Software/pull/1)I did on my fork [this action](https://github.com/atteneder/KTX-Software/actions/runs/847849889) was started.

So I believe it will be included in PRs and releases once it was merged.

MarkCallow:
> I implemented all feedback and started yet another test run.

BUILDING.md still needs updating.

atteneder:
> BUILDING.md still needs updating.

@MarkCallow Can you provide feedback on what's missing in the Android section I already wrote?

BUILDING.md
```diff
@@ -21,6 +21,7 @@ Supported platforms (please to through their specific requirements first)
 - [iOS and macOS](#ios-and-macos)
 - [Web/Emscripten](#webemscripten)
 - [Windows 64-bit](#windows-x64)
+- [Android](#android)
 
 > **Note:** Android builds will follow
```
MarkCallow, line 26: This line needs deleting.

MarkCallow:
> Can you provide feedback on what's missing in the Android section I already wrote?

Very sorry. Missed the addition last time I reviewed this PR. The old "Android builds will follow" needs to be deleted. Other than that, it looks fine.

atteneder:
No worries. Thanks for your support!

PR #413, opened 2021-05-02
Extended BasisU C binding

atteneder:
feat: Added more functions to BasisU C binding:

- getYFlip();
- getIsEtc1s();
- getTextureType();

This enables users to apply correct flipping and support transcoding of both ETC1s *and* UASTC.

Certified that these functions are properly exported on all platforms:

- ktxTexture2_GetNumComponents
- ktxTexture2_NeedsTranscoding
- ktxTexture2_TranscodeBasis

Fix: replaced bool by uint32_t in interface to avoid problems on some platforms.

### DIFF
--- a/interface/basisu_c_binding/inc/basisu_c_binding.h
+++ b/interface/basisu_c_binding/inc/basisu_c_binding.h
@@ -3,11 +3,11 @@
 
 #pragma once
 
-#ifdef DLL_EXPORT_FLAG
+#if defined(_MSC_VER)
 #define DLL_EXPORT __declspec(dllexport)
 #else
 #define DLL_EXPORT
-#endif
+#endif // defined(_MSC_VER)
 
 #include <basisu_transcoder.h>
 
--- a/interface/basisu_c_binding/inc/basisu_c_binding.h
+++ b/interface/basisu_c_binding/inc/basisu_c_binding.h
@@ -36,8 +36,8 @@ public:
     uint32_t getNumLevels(uint32_t image_index);
     uint32_t getImageWidth(uint32_t image_index, uint32_t level_index);
     uint32_t getImageHeight(uint32_t image_index, uint32_t level_index);
-    bool getYFlip();
-    bool getIsEtc1s();
+    uint32_t getYFlip();
+    uint32_t getIsEtc1s();
     basis_texture_type getTextureType();
     uint32_t getImageTranscodedSizeInBytes(uint32_t image_index, uint32_t level_index, uint32_t format);
     uint32_t startTranscoding();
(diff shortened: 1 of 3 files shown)


MarkCallow:
I have an incoming change to the binding to fix some warnings. Hopefully this change and my fixes will merge smoothly. (I am upping the warning setting from /W1 to /W4 on Windows and -Wall -Wextra on clang/gcc. I may backpedal on the -Wextra. I am getting nearly 2000 warnings from gcc in code that compiles without warnings at the same settings on clang.)

atteneder:
Sounds good. Let me know if this PR needs adjustments.

lib/internalexport.def
```diff
@@ -41,6 +41,9 @@ EXPORTS
     ktxTexture1_glTypeSize
     ktxTexture2_GetImageOffset
     ktxTexture2_calcLevelOffset
+    ktxTexture2_GetNumComponents
+    ktxTexture2_NeedsTranscoding
+    ktxTexture2_TranscodeBasis
```
MarkCallow, line 46: These 3 are already exported by the KTX_API decorations in ktx.h. Why do you need to add them here?

MarkCallow:
Please resolve the conflicts.

lib/internalexport.def
atteneder: True. I'll remove those and investigate which problem I (thought) this solved later.

atteneder:
@MarkCallow I resolved all issues, but the build failed by mistake (networking problem on appveyor's side?). How should we proceed?

MarkCallow:
> but the build failed by mistake (networking problem on appveyor's side?)

This is becoming a PITA. I've started a "re-run incomplete".

interface/basisu_c_binding/inc/basisu_c_binding.h
```diff
@@ -3,11 +3,11 @@
 
 #pragma once
 
-#ifdef DLL_EXPORT_FLAG
+#if defined(_MSC_VER)
 #define DLL_EXPORT __declspec(dllexport)
```
MarkCallow, line 7: This is going to be `__declspec(dllexport)` for clients trying to link to the resulting library too (if they `#include "basisu_c_binding" to get the function declarations)  so links will fail. clients need `__declspec(dllimport)`. What am I missing? How are the transcodetests building?

atteneder: Right!

transcodetests link against `obj_basisu_cbind`, which is an `OBJECT` type library, so it might works because of that.

This is subject to change in the [apple silicon PR](https://github.com/KhronosGroup/KTX-Software/pull/415), so we probably should do that one first.

I'll look into it, but making a dedicated compile option (similar to `KTX_API`) is probably a more robust solution.

MarkCallow: I've landed the Apple Silocon PR. Please take a look at this issue.

MarkCallow: Please fix this. It is still always defining `__declspec(export)`. The usual way to do this is what we are doing with `KTX_API` in `ktx.h`. For _ MSC_VER, unless already defined, it is defined as `__declspec(dllimport)`. When compiling the library the compile commands define it via `-DKTX_API=__declspec(dllexport)`. That way users don't need to do anything to ensure apps are compiled with `dlimport`.

atteneder: I introduced `KTX_BASISU_API`, which is similar to `KTX_API`

PR #415, opened 2021-05-03
macOS Apple Silicon support

atteneder:
This patch enables building macOS binaries that support Apple Silicon (arm64) platform or universal binaries supporting both Intel (x86_64) and Apple Silicon Macs.

There's a catch:
Basis Universal SSE support has to be disabled, in arm builds and even in universal builds. SSE support is detected at configuration time, but a fat arm/intel binary would require detecting it at compile time per architecture.

I've changed the CI build to create universal binaries and disabled SSE for now Alternative solutions could be:

- Dedicated binaries per architecture
- Compile time SSE detection in BasisU

I've mentioned this to BasisU in [this comment](https://github.com/BinomialLLC/basis_universal/issues/123#issuecomment-831142969).


MarkCallow:
You need to fix the Windows build. the transcodetest build is exiting due to a multiply defined symbol in basisu_c_binding.lib and ktx.lib. It's the etc1s global codebook.

There ia cmake/c program to detect the cpu type: cmake/cputypetest.cmake. It needs fixing to recognize the M1. Clearly it is currently returning x86_64 because the SSE option in CMake is disabled for every other case. Please look at fixing it. It's a bit of a hack right now for reasons explained in comments there.

MarkCallow:
When I switched the CI build to Xcode 12.4 it started overrunning Travis's 4MB log file limit so I changed it to use `xcpretty` to reduce the amount of output. Just as well as we'll have even more builds after this. If with these additional builds it overruns even with `xcpretty` we may have to resort to `xcodebuild -quiet`. I think this can be passed from cmake like this

```bash
cmake --build . --config Release -- -quiet
```

MarkCallow:
As well as fixing the Windows build, please add info about building for the M1 to BUILDING.md. Do we need to change anything in the iOS build for the new m1 iPad?

One thing that comes to mind for both macOS and iOS is the set of transcode targets supported. The M1 supports all modern formats and PVRTC so we should include all transcoders. It will be good for testing.

lexaknyazev:
> One thing that comes to mind for both macOS and iOS is the set of transcode targets supported. The M1 supports all modern formats and PVRTC so we should include all transcoders. It will be good for testing.

For whatever reason, iOS on M1 does not support BC* formats.

atteneder:
> You need to fix the Windows build. The transcodetest build is exiting due to a multiply defined symbol in basisu_c_binding.lib and ktx.lib. It's the etc1s global codebook.

I think I found a way around this difference in linking behaviour.. let's see what the CI results in.

> There is a cmake/c program to detect the cpu type: cmake/cputypetest.cmake. It needs fixing to recognize the M1. Clearly it is currently returning x86_64 because the SSE option in CMake is disabled for every other case. Please look at fixing it. It's a bit of a hack right now for reasons explained in comments there.

I agree, but I don't see how this can be fixed at configuration time (only) for universal libraries. That's why I think BasisU should do some sort of compile time feature detection.

atteneder:
> As well as fixing the Windows build, please add info about building for the M1 to BUILDING.md

I'll look at it.

> Do we need to change anything in the iOS build for the new m1 iPad?

My rather uninformed gut feeling answer would be no. and I feel a bit confirmed by @lexaknyazev saying iOS/M1 doesn't support BC*.

> One thing that comes to mind for both macOS and iOS is the set of transcode targets supported. The M1 supports all modern formats and PVRTC so we should include all transcoders. It will be good for testing.

Good point. I'll have a look.

atteneder:
> > As well as fixing the Windows build, please add info about building for the M1 to BUILDING.md
> 
> I'll look at it.

Done

> > One thing that comes to mind for both macOS and iOS is the set of transcode targets supported. The M1 supports all modern formats and PVRTC so we should include all transcoders. It will be good for testing.
> 
> Good point. I'll have a look.

As far as [I can see](https://github.com/KhronosGroup/KTX-Software/blob/d5ae1ec0003e68a297e927ab32ab0096eb3ad3ba/CMakeLists.txt#L254) there is no configuration time preselection of supported transcode targets (except for [WebGL](https://github.com/KhronosGroup/KTX-Software/blob/d5ae1ec0003e68a297e927ab32ab0096eb3ad3ba/CMakeLists.txt#L287)).

It would be good if we can make a smart per-platform preselection and offer CMake options that users can tweak, but:

- For macOS universal this would be a shared set for both ARM and Intel though, I'm afraid.
- I consider this another feature and out of scope of this PR

atteneder:
> > There is a cmake/c program to detect the cpu type: cmake/cputypetest.cmake. It needs fixing to recognize the M1. Clearly it is currently returning x86_64 because the SSE option in CMake is disabled for every other case. Please look at fixing it. It's a bit of a hack right now for reasons explained in comments there.
> 
> I agree, but I don't see how this can be fixed at configuration time (only) for universal libraries. That's why I think BasisU should do some sort of compile time feature detection.

I added detection of Apple Silicon to this script. This is now used to prevent cross compiling load tests apps on M1 to Intel. Still, this only detects the host's architecture and not the target architecture(s).

ci_scripts/build_macos.sh
```diff
@@ -22,14 +22,19 @@ echo "Configure KTX-Software (macOS)"
...
+  -DCMAKE_OSX_ARCHITECTURES="\$(ARCHS_STANDARD)" \
+  -DBASISU_SUPPORT_SSE=OFF \
   -DKTX_FEATURE_LOADTEST_APPS=ON -DVULKAN_INSTALL_DIR="${VULKAN_INSTALL_DIR}" \
   -DXCODE_CODE_SIGN_IDENTITY="${CODE_SIGN_IDENTITY}" \
   -DXCODE_DEVELOPMENT_TEAM="${DEVELOPMENT_TEAM}" \
   -DPRODUCTBUILD_IDENTITY_NAME="${PKG_SIGN_IDENTITY}"
 else # No secure variables means a PR or fork build.
   cmake -GXcode -Bbuild-macos \
   -DKTX_FEATURE_DOC=ON \
-  -DKTX_FEATURE_LOADTEST_APPS=ON -DVULKAN_INSTALL_DIR="${VULKAN_INSTALL_DIR}"
```
MarkCallow, line 32: Why is `-DVULKAN_INSTALL_DIR="${VULKAN_INSTALL_DIR}` removed from PR & fork builds but not from a origin branch builds?

B.t.w PR #427 will remove `VULKAN_INSTALL_DIR` from the build.

MarkCallow (changes requested):
Overall looks good except from several typos in `BUILDING.md`. There are also a couple of questions in the review. The one on `build_macos.sh` definitely needs an answer.

BUILDING.md
```diff
@@ -123,6 +127,37 @@ cmake -GXcode -Bbuild/mac -DKTX_FEATURE_LOADTEST_APPS=ON -DVULKAN_INSTALL_DIR="$
 # Compile the project
 cmake --build build/mac
 ```
+##### Apple Silicon and Universal Binaries
+
+Macs are either based on Intel or the newer Apple Silicon architecture. By default CMake configures to build for your hosts platform, whichever it is. If you want to cross compile universal binaries (that support both platforms), add the paramter `-DCMAKE_OSX_ARCHITECTURES="\$(ARCHS_STANDARD)"` to cmake.
+
+> **Known limitations:**
+> - Load tests apps are not supported on native Apple Silicon and cannot be cross compiled for Intel either
```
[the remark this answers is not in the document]
atteneder: TL;DR: While technically possible, it's not straightforward and I don't think it's of great use.

I briefly tried and faced obstacles (my system's libSDL2 was picked although it was arm64 and linking intel apps failed).

If someone deems this important, they can go ahead, figure it out and provide instructions, but I will not dedicate my resources on it.

ci_scripts/build_macos.sh
atteneder: It is not removed, but I split up those two in separate lines for readability. (see lines 36 and 37)

Thanks for pointing out `VULKAN_INSTALL_DIR` is removed.

MarkCallow (approved):

PR #419, opened 2021-05-10
Cleanup Vulkan SDK environment variables

DDoS:
KTX doesn't compile when Perl is found but this variable isn't set. The README states that Vulkan SDK is an optional dependency. This allows building without the Vulkan SDK.

Also this VULKAN_SDK variable seems to be redundant with VULKAN_INSTALL_DIR.


MarkCallow:
Thank you for these changes. Please also remove the redundant variable. `VULKAN_SDK` is what the SDK setup scripts create so that should be the remaining variable.

Appveyor has recently fairly often suffered these artifact upload errors. It's a nuisance.

MarkCallow:
The SDK installation on macOS has frameworks & libraries for both macOS and i{,Pad}OS. VULKAN_INSTALL_DIR was originally used to avoid having to specify the same location in 2 different places in the early days of the SDK. More recent SDKs include `setup-env.sh` which sets `VULKAN_SDK` to the `macOS` folder of the installation directory. We could do everything from `$VULKAN_SDK` by, in the iOS case, using `$VULKAN_SDK/../MoltenVK`. That will probably be simpler. As far I as know there is no plan to separate the 2 SDKs so I using a single variable shouldn't cause problems down the road. What do you think @DDoS?

DDoS:
@MarkCallow I've removed VULKAN_INSTALL_DIR as much as possible, but I came across this comment at the top of `libvulkan.gypi` that suggest it's not possible to entirely remove it:
```
# Note we can't just set VULKAN_INSTALL_DIR here to $(VULKAN_SDK)/..
# because gyp generators, at least the xcode one, will remove any leading
# component of a path that includes a '..'.
```
Although I don't have paths for `$(VULKAN_INSTALL_DIR)/MoltenVK/iOS` or `$(VULKAN_INSTALL_DIR)/MoltenVK/macOS` in my install of the SDK, so perhaps this file is outdated anyways?

MarkCallow:
> so perhaps this file is outdated anyways?

Yes the file is outdated. We don't use GYP anymore so the .. issue has gone. The VulkanSDK has also been changed. There is now a `MoltenVK/MoltenVK.xcframework` that has both macOS and iOS libraries in it. That's what the change I landed that caused a conflict with this one, was about.

tests/loadtests/vkloadtests/VulkanLoadTestSample.cpp
```diff
@@ -65,11 +65,11 @@ VulkanLoadTestSample::loadMesh(std::string filename,
 vk::PipelineShaderStageCreateInfo
 VulkanLoadTestSample::loadShader(std::string filename,
                                  vk::ShaderStageFlagBits stage,
-                                 std::string modname)
+                                 const char* modname)
```
MarkCallow, line 68: Why are you changing this?

tests/loadtests/vkloadtests/VulkanLoadTestSample.h
```diff
@@ -60,7 +60,7 @@ class VulkanLoadTestSample : public LoadTestSample {
     vk::PipelineShaderStageCreateInfo
     loadShader(std::string filename,
                vk::ShaderStageFlagBits stage,
-               std::string modname = "main");
+               const char* = "main");
```
MarkCallow, line 63: Why are you changing this?

tests/loadtests/vkloadtests/VulkanLoadTestSample.cpp
```diff
@@ -65,11 +65,11 @@ VulkanLoadTestSample::loadMesh(std::string filename,
...
 {
     vk::PipelineShaderStageCreateInfo shaderStage({}, stage);
     shaderStage.module = vkctx.loadShader(filename);
-    shaderStage.pName = modname.c_str();
+    shaderStage.pName = modname;
```
MarkCallow, line 72: And this.

BUILDING.md
```diff
@@ -106,7 +106,8 @@ You need to install the following
 - Xcode
 - [Doxygen](#doxygen) (only if generating documentation)
 
-For the load tests applications you need to install the [Vulkan SDK](#vulkan-sdk).
+For the load tests applications you need to install the [Vulkan SDK](#vulkan-sdk), and create an environment
```
MarkCallow, line 109: This should say something like

"For the load tests applications you need to install the [Vulkan SDK](#vulkan-sdk) and run the included `setup-env.sh` or otherwise create an environment variable `VULKAN_SDK` pointing to the `macOS` directory within the directory where you installed the SDK."

MarkCallow (changes requested):

MarkCallow:
Do you think we should make `VULKAN_SDK` a CMake option variable initially set from the environment variable? This is so that people who haven't set it already can tell CMake where to look? On macOS at least, the separate script to set the environment may not have been run or, very likely, people will not have jumped through the hoops necessary to get their environment variables into Launchpad so when they run CMake from there, it will be clueless about the environment variable.

Sorry this is turning into a lot of work for you.

MarkCallow:
> Do you think we should make VULKAN_SDK a CMake option variable initially set from the environment variable?

Definitely we should. It is as difficult as ever in macOS Big Sur to sync one's shell environment variables to Launchpad so that Launchpad-launched apps will see them. Having a CMake variable will be a big help to those using the CMake GUI on macOs.

DDoS:
> > so perhaps this file is outdated anyways?
> 
> Yes the file is outdated. We don't use GYP anymore so the .. issue has gone.

Can I just ignore the GYP files then?

tests/loadtests/vkloadtests.cmake
```diff
@@ -16,35 +16,27 @@ endif()
...
 if(APPLE)
-    # Try to locate Vulkan SDK install directory
-    if(NOT VULKAN_INSTALL_DIR)
-        message(SEND_ERROR "Please provide a valid path to your Vulkan SDK installation in CMake variable 'VULKAN_INSTALL_DIR'!")
-    endif()
-
     if(IOS)
-        set( VULKAN_SDK "${VULKAN_INSTALL_DIR}/MoltenVK" )
-        set( MOLTEN_VK_FRAMEWORK ${VULKAN_SDK}/iOS/framework/MoltenVK.framework )
+        set( MOLTEN_VK_FRAMEWORK $ENV{VULKAN_SDK}/iOS/framework/MoltenVK.framework )
```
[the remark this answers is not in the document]
DDoS: The `build_macos.sh` script changes the `VULKAN_SDK` path for the iOS build. To make sure I understand correctly, when building for iOS, the `VULKAN_SDK` variable will always be pointing to `VULKAN_INSTALL_DIRECTORY/macOS`, and the script should be responsible for changing the path to `MoltenVK/iOS`?

tests/loadtests/vkloadtests/VulkanLoadTestSample.cpp
DDoS: Because this function returns `vk::PipelineShaderStageCreateInfo,` which has a pointer to the data in the `std::string`, which is destroyed on return. This was making the load tests crash on my machine. There's in another implementation of `loadShader` in this project that also uses `const char*`, I assume to avoid this problem.

tests/loadtests/vkloadtests/VulkanLoadTestSample.h
DDoS: Because the string is used in the return value so it's undefined behaviour.

tests/loadtests/vkloadtests/VulkanLoadTestSample.cpp
DDoS: This pointer can't be returned because it's owned by an object on the stack.

MarkCallow:
> Can I just ignore the GYP files then?

Yes.

MarkCallow:
I've just been looking over the Vulkan SDK use globally, i.e. not just via the changed parts shown here. From that I learned that for Linux, Windows and macOS `find_package` is used from which the variables `Vulkan_INCLUDE_DIRS` and `Vulkan_LIBRARIES` are set. These are  used for building. For iOS there is a special case which uses, in this PR, `$ENV{VULKAN_SDK}/include` which is actually the same place as `Vulkan_INCLUDE_DIRS` will have been set to. The library to use has also been special cased for iOS to use the `MOLTEN_VK_FRAMEWORK` variable which is derived, in this PR, from `VULKAN_SDK`.

In short the only place `VULKAN_SDK` is needed is for setting `MOLTEN_VK_FRAMEWORK`.  I therefore propose that we stop using `VULKAN_SDK` and instead have a CMake variable `MOLTEN_VK_SDK` which is set by default to `$ENV{VULKAN_SDK/../MoltenVK}` and remove the special case for the include directory for iOS. This will avoid any confusion with the SDK's VULKAN_SDK. Use of that could make people think it is needed for other builds too.

Will you make these changes? Can you test iOS builds?

tests/loadtests/vkloadtests/VulkanLoadTestSample.cpp
MarkCallow: This is because the callers are using the default parameter, right? If the caller passed the modname parameter it would still be valid when the function returns and up until the caller exits, by which time the pipeline would be created and it wouldn't matter. Would making the parameter a reference, std::string&, help?

Where is this crashing? Windows? I've run the apps many times on macOs, iOS and Linux and never had a crash. I can't run on Windows because my Windows box doesn't support Vulkan.

DDoS: The pointer would still be invalid if the string is passed by value, like it was in the original version, because the parameter scope is the function IIRC. Passing by reference will extend the lifetime outside this function, but only until the string is destroyed. In this case, if the caller is using the default value, it should become invalid when the caller's scope ends.
Basically the string's lifetime must be as long or longer than that of the `vk::PipelineShaderStageCreateInfo`. By using a pointer to a literal, the string exists in memory for the life of the program, so it's not an issue.
Because this is undefined behaviour, a crash is not guaranteed. This will depend on the compiler optimizations and stdlib implementation of std::string. I had this crash with MSVC in debug, and it was happening later then the driver was trying to use the `vk::PipelineShaderStageCreateInfo` structure. I think MSVC in debug clobbers released memory to help identify these kinds of lifetime bugs, so it crashes were other compilers don't.

DDoS:
@MarkCallow I made some changes, let me know if this is what you have in mind.

DDoS:
Found another bug I'll lump into this PR: mkvk.cmake should use `CMAKE_HOST_WIN32` instead of `WIN32` otherwise building for emscripten on Windows won't work (it will use the wrong custom commands).

MarkCallow:
Please move the modname changes to another PR. I want them visible in `git log` later. While you are at it, please also fix the failure of `VulkanLoadTestSample::loadShader` to pass the modname parameter to the underlying VulkanContext::LoadShader. Once that is done the line `shaderStage.pName = modname` can be removed as VulkanContext::LoadShader will have already set it.

Please also move the CMAKE_HOST_WIN32 change to another PR.

I'll review the VULKAN_SDK removal changes later today.

MarkCallow:
I just landed PR #415 & #428 which have caused conflicts with this. Please resolve them.

BUILDING.md
```diff
@@ -106,7 +106,9 @@ You need to install the following
 - Xcode
 - [Doxygen](#doxygen) (only if generating documentation)
 
-For the load tests applications you need to install the [Vulkan SDK](#vulkan-sdk).
+For the load tests applications you need to install the Vulkan SDK and run the included setup-env.sh or
+otherwise create an environment variable `VULKAN_SDK` pointing to the macOS directory within the directory where you installed the SDK.
+For iOS you additionally need to set the CMake cache variable `MOLTEN_VK_SDK` to the root of MoltenVK inside the Vulkan SDK.
```
MarkCallow, line 111: Because when the CMake GUI is run from macOS Launchpad it will not see environment variables, VULKAN_SDK needs to be a CMake variable initialized from $ENV{VULKAN_SDK} if it is visible. The instructions on these lines need to reflect this. e.g set the CMake cache variable VULKAN_SDK if it is not already set.

cmake/mkvk.cmake
```diff
@@ -28,7 +36,7 @@ list(APPEND mkvkformatfiles_output
 if(WIN32)
     add_custom_command(OUTPUT ${mkvkformatfiles_output}
         COMMAND ${CMAKE_COMMAND} -E make_directory lib
-        COMMAND "${BASH_EXECUTABLE}" -c "VULKAN_SDK=${VULKAN_SDK} lib/mkvkformatfiles lib"
+        COMMAND "${BASH_EXECUTABLE}" -c "lib/mkvkformatfiles lib"
```
MarkCallow, line 39: VULKAN_SDK is still needed here. I overlooked this script the other day. sorry.

cmake/mkvk.cmake
```diff
@@ -40,7 +48,7 @@ if(WIN32)
 else()
     add_custom_command(OUTPUT ${mkvkformatfiles_output}
         COMMAND ${CMAKE_COMMAND} -E make_directory lib
-        COMMAND VULKAN_SDK=${VULKAN_SDK} lib/mkvkformatfiles lib
+        COMMAND lib/mkvkformatfiles lib
```
MarkCallow, line 51: Here also it is still needed.

tests/loadtests/vkloadtests.cmake
```diff
@@ -11,40 +11,32 @@ else()
             # Fallback for inconsistent variable name on Windows
             set(Vulkan_INCLUDE_DIRS ${Vulkan_INCLUDE_DIR})
         endif()
+        set(VULKAN_SDK "${Vulkan_INCLUDE_DIR}/..")
```
MarkCallow, line 14: Why did you add this?

DDoS: `find_package(Vulkan)` only gives include and lib paths. This code also needs the share and Frameworks paths, so I need the Vulkan root path (SDK path).

DDoS:
@MarkCallow New round of changes. I also found another bug, there are missing quotes in cputypetest.cmake, so I'm opening another PR for that.

.gitignore
```diff
@@ -32,3 +33,6 @@
 # version files
 version.h
 
+# Visual Studio
+.vs/
+CMakeSettings.json
```
MarkCallow, line 38: Why the 3 additions here. I haven't seen an issue with Visual Studio leaving files in the source when I've used it but perhaps that's because I do my builds in `build` which is already ignored.

tests/loadtests/vkloadtests.cmake
```diff
@@ -14,37 +14,28 @@ else()
     endif()
```
MarkCallow, line 14: Please add the following just above line 14

```
# Once we've moved on to CMake 3.20
#cmake_path(REMOVE_FILENAME ${Vulkan_LIBRARIES} OUTPUT VARIABLE Vulkan_LIBRARY_DIR)
# Until then
string(REGEX REPLACE lib/.*$ lib/ Vulkan_LIBRARY_DIR ${Vulkan_LIBRARIES})
```

CMakeLists.txt
```diff
@@ -54,6 +52,7 @@ if(APPLE)
     set(PRODUCTBUILD_KEYCHAIN_PATH "" CACHE FILEPATH "pkgbuild keychain file")
     if(IOS)
         set(XCODE_PROVISIONING_PROFILE_SPECIFIER "" CACHE STRING "Xcode provisioning profile specifier")
+        set(MOLTEN_VK_SDK "$ENV{VULKAN_SDK}/../MoltenVK" CACHE STRING "MoltenVK directory inside the Vulkan SDK")
```
MarkCallow, line 55: If VULKAN_SDK isn't set then MOLTEN_VK_SDK will end up being set to `../MoltenVK`. It might be better if it is set to an empty string in this case. I also found that if VULKAN_SDK is set incorrectly at the initial configure (using the CMake GUI) then MOLTEN_VK_SDK will be wrong, of course.  If you then fix VULKAN_SDK and configure again, MOLTEN_VK_SDK does not change. I would prefer if it picked up the new value of VULKAN_SDK but I don't know how to do that and I don't know if that is acceptable practice in CMake. Do you know @DDoS?

cmake/mkvk.cmake
```diff
@@ -2,6 +2,16 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Code generation scripts that require a Vulkan SDK installation
+
```
MarkCallow, line 5: Do you think we should change the code generation scripts to use Vulkan_INCLUDE_DIRS instead of VULKAN_SDK? If would mean moving find_package(Vulkan) from vkloadtests.cmake and probably changes to the scripts.

BUILDING.md
```diff
@@ -105,7 +105,11 @@ You need to install the following
 - Xcode
 - [Doxygen](#doxygen) (only if generating documentation)
 
-For the load tests applications you need to install the [Vulkan SDK](#vulkan-sdk).
+For the load tests applications you need to install the Vulkan SDK.
+To build for macOS you need set the CMake cache variable `VULKAN_SDK` to the root of the Vulkan SDK, if it is not already set.
```
MarkCallow, line 109: Delete this line. It is no longer necessary.

BUILDING.md
```diff
@@ -105,7 +105,11 @@ You need to install the following
...
+To build for iOS you need to set the CMake cache variable `MOLTEN_VK_SDK` to the root of MoltenVK inside the Vulkan SDK, if it is not already set.
```
MarkCallow, line 110: Add the following after this line:

Caution: `setup.env` in the macOS Vulkan SDK sets VULKAN_SDK to the macOS folder of the SDK, a sibling of the MoltenVK folder.

MarkCallow:
Please resolve the conflict in .gitignore, make the changes I have requested and provide your thoughts on the questions I raised.

.gitignore
DDoS: These are relevant when using Visual Studio's native CMake integration. [You can work with CMake in VS without having to use the VS project generator](https://docs.microsoft.com/en-us/cpp/build/cmake-projects-in-visual-studio?view=msvc-160). This means VS treats the project more like a solution, so it creates a `.vs` directory for its private files, and a `CMakeSettings.json` so you can configure CMake generation.
I added `/install*/` because that's a typical directory for the CMake install, and it's the default used by VS.

cmake/mkvk.cmake
DDoS: `mkvkformatfiles` still needs a `VULKAN_SDK` environment variable.

CMakeLists.txt
DDoS: I'm not sure if we could differentiate between `VULKAN_SDK` changing, and the user choosing a different `MOLTEN_VK_SDK` with the current scheme.

What might work is to use a different variable; initialized to `${VULKAN_SDK}/../MoltenVK` if `MOLTEN_VK_SDK` is empty, otherwise initialized to `${MOLTEN_VK_SDK}`. The default value of `MOLTEN_VK_SDK` could then be empty left empty by the user for the default path, and a change in `${VULKAN_SDK}` would change the MoltenVK SDK path.

MarkCallow: Let's leave it as is.

BUILDING.md
MarkCallow: Oh! I missed those 2. It is unfortunate that find_package(Vulkan) doesn't provide more information. The ICD is installed in /usr/local/share/ so could be found by deriving VULKAN_SDK from Vulkan_LIBRARIES as I did for Vulkan_LIBRARY_DIR. However the framework does not appear to be installed anywhere but inside the SDK. I will experiment with copying libvulkan.dylib to the app bundles instead.

cmake/mkvk.cmake
MarkCallow: Yes. Changes to the scripts will likely be necessary. mkvkformatfiles vulkan_core.h nothing else. I'm sure we could make it work from Vulkan_INCLUDE_DIRS. I will look at it later to confirm but IIRC you pass it the path of the file so it is the mkvk.cmake that is using $VULKAN_SDK to provide that path.

BUILDING.md
MarkCallow: `vulkan.framework` is not needed in the app bundle because the app is linked to the .dylib which is copied to the bundle by one of the lines you modified yesterday. So please delete lines 260 to 263 of vkloadtests.cmake.

VK_ICD is not used so you can also delete line 194. Since the libraries are in the app bundle we need a custom icd.json which is stored in other_lib/mac/resources and which is set in the variable MOLTEN_VK_ICD and is copied to the bundle. We do not need the icd.json from the SDK.

I have verified these changes. After them, the app still works both when run in Xcode and when run standalone.

Indeed the macOS build does not need VULKAN_SDK, except possibly to help cmake find the package.

cmake/mkvk.cmake
DDoS: It's OK to have more than one `find_package(Vulkan)` because the result is cached so later call don't search again.
Also `mkvkformatfiles` only needs `VULKAN_SDK` to find the include directory, so I changed it to use `Vulkan_INCLUDE_DIR` directly instead.
I think it's better to use the `Vulkan_INCLUDE_DIR` and `Vulkan_LIBRARY` (singular form) output variables from `find_package` because they won't be lists, so doing something like `cmake_path(REMOVE_FILENAME ${Vulkan_LIBRARY} OUTPUT VARIABLE Vulkan_LIBRARY_DIR)` is safer.
With these changes and those in `vkloadtests.cmake`, there are no more references to `VULKAN_SDK`, except for the default `MOLTEN_VK_SDK` value.

MarkCallow (approved):

MarkCallow:
I made 2 additional changes to this PR in my local workarea and merged the result to master, hence the conflicts. The 2 changes were

1. Print a status message in vkloadtests.cmake showing where Vulkan was found.
2. Since find_package isn't finding the Vulkan SDK when building for iOS (it was printing out an error in addition to the "skipping mkvk" message during config) I made the find_package in mkvk.cmake only run if not iOS. If it is iOS it just returns. There's no need to run mkvk in an iOS build.

I am sorry that my doing the merge has put my name on this work. Thank you very much for initiating this and for all your hard work cleaning it up. The result is much better. What we had evolved from a starting point before there even was a VulkanSDK for macOS and probably before there was a FindVulkan.cmake.

As the work is now merged I will close the PR.

PR #426, opened 2021-05-16
Update to upto date documentation

wasimabbas-arm:
There are no KTX2 and KTX branches in the repo. Since KTX2 work is now merged this could be removed.


MarkCallow:
Thanks.

PR #428, opened 2021-05-18
fix: Embed bitcode in iOS libraries

atteneder:
Some developers want to build apps for iOS that includes bitcode. Building an app for watchOS/tvOS even requires you to provide it. This means all dependencies and libraries in use have to provide bitcode as well. To make it easier for those devs, I'd enable bitcode by default. Only downside is an increased binary size, so one could argue to make this a CMake option.

Here's the [original request](https://github.com/atteneder/KtxUnity/issues/37)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,6 +43,8 @@ if(CPU_ARCHITECTURE STREQUAL x86_64 OR CPU_ARCHITECTURE STREQUAL x86)
     option( BASISU_SUPPORT_SSE "Compile with SSE support so applications can choose to use it" ON )
 endif()
 
+CMAKE_DEPENDENT_OPTION(KTX_EMBED_BITCODE "Embed bitcode in binaries" OFF "APPLE AND IOS" OFF)
+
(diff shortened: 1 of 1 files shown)


MarkCallow:
Do the library binaries we currently provide in the repo for macOS include bitcode? It probably doesn't matter as they're only used for the loadtests and those won't be shipped.

MarkCallow:
How much does it increase the size of the binaries?

atteneder:
Roughly 3 times as large 😕

Here's what I measured:

| Variant    |   ktx | ktx_read |
|------------|-------|----------|
| Debug      | 28.0M |     8.9M |
| Debug+BC   | 72.0M |    26.0M |
| Release    |  5.5M |     3.4M |
| Release+BC | 16.0M |     9.4M |

If this is a problem, then I'd suggest adding a CMake option. People use to inject compile flags via some global CMake variables, but I consider this bad practice and wouldn't recommend that.

atteneder:
> Do the library binaries we currently provide in the repo for macOS include bitcode?

No. I'm not sure if bitcode can even be created for dynamic macOS libs at all.

MarkCallow:
> If this is a problem

Yes this is a problem. Please add a CMake option.

atteneder:
> > If this is a problem
> 
> Yes this is a problem. Please add a CMake option.

Done

PR #430, opened 2021-05-19
Fix stale pointer in loadShader

DDoS:
There is a crash with MSVC in debug because this function returns vk::PipelineShaderStageCreateInfo, which has a pointer to the data in the std::string, which is destroyed on return.


MarkCallow:
Thank you.

PR #431, opened 2021-05-19
Use CMAKE_HOST_WIN32 in mkvk.cmake

DDoS:
mkvk.cmake should use CMAKE_HOST_WIN32 instead of WIN32 otherwise building for emscripten on Windows won't work (it will use the wrong custom commands).


MarkCallow:
And thanks for this.

PR #433, opened 2021-05-19
Add astc support

wasimabbas-arm:
This is the start of astc support. First the cli options required are provided.

Since ktx-software now supports multiple encoders this patch extracts
out the common options into '--encode <encoding>' and removes redundant options.
Also add enums for astc compressor into ktx.h.

PR #437, opened 2021-06-04
Removing builds for Android x86 ABI

atteneder:
Main reason: It blocks inclusion of ARM ASTC encoder #433 .

Also it's not considered very relevant (not much hardware exists)

If anyone requires this at some point, we can bring it back but have to make sure unsupported features are excluded.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -52,11 +52,14 @@ jobs:
         ANDROID_ABI: armeabi-v7a
         ANDROID_NDK: ${{ steps.setup-ndk.outputs.ndk-path }}
 
-    - name: android_x86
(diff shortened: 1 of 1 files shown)


MarkCallow:
@wasimabbas-arm has managed to compile the ASTC encoder for x86 so we could still keep this. However since the Windows build does not officially support 32-bit (x86) I think it is best not to try here either. Merging.

abbaswasim:
Should we just not get rid of the section instead of leaving it as comments?

atteneder:
@abbaswasim 

fair point. I thought I leave it in (commented) for now, in the unexpected case some users request Android x86. It can always be deleted later.

PR #438, opened 2021-06-09
Please consider making `ktxStream` public

UberLambda:
Hi,

This pull request moves `lib/stream.h`'s contents to `ktx.h`, and it makes all the previously-private types and functions therein public.
It also adds a `custom` stream type, with a user pointer and a `size_t` in its `data`.

### Motivation
I am currently creating a Rust wrapper for this library, and I wanted to read/write KTXs from Rust streams - anything <tt>dyn ([Read](https://doc.rust-lang.org/std/io/trait.Read.html) + [Write](https://doc.rust-lang.org/std/io/trait.Write.html) + [Seek](https://doc.rust-lang.org/std/io/trait.Seek.html))</tt>, like [std::fs::File](https://doc.rust-lang.org/std/fs/struct.File.html). So I had to:
1. Make `ktxStream` public.
2. [Create C-visible Rust functions for a custom `ktxStream`](https://github.com/UberLambda/libktx-rs/blob/cd25fdfc54aebe2450706b7bc335f7864873b19f/libktx-rs-sys/src/stream.rs#L116) (read, write, getpos, setpos...).
3. [Construct the ktxStream from Rust, wrapping a heap-allocated trait object](https://github.com/UberLambda/libktx-rs/blob/cd25fdfc54aebe2450706b7bc335f7864873b19f/libktx-rs-sys/src/stream.rs#L52).   

Custom `ktxStream`s would be useful in many other scenarios, though. Off the top of my head:
- Reading/writing KTXs from C++ `std::iostream`s
- Reading/writing KTXs from PhysicsFS' not-quite-`FILE *` streams
- Reading/writing KTXs from tinygltf filesystem callbacks

### Notes
- The `custom` struct contains both a `void *` and `size_t`. This would allow for direct storage of Rust's "fat pointers" (a pointer to a vtable + a size).
My current workaround is to add another layer of indirection (`ktxMem *` reinterpret-casted as a pointer to a Rust struct -> heap-allocated fat pointer -> heap-allocated trait object). Having the `size_t` in the `data` union would remove the need for heap-allocating the pointer.

Thanks!


MarkCallow:
Happy to hear you are creating a Rust wrapper. Please add a test (external to the library) that uses the stream functions. This is to ensure that linking will be successful on Windows, i.e. the all the necessary functions are properly exported.

include/ktx.h
```diff
@@ -755,13 +756,119 @@ enum ktxTextureCreateFlagBits {
...
+
+    enum streamType type;
+    union {
+        FILE* file;
+        ktxMem* mem;
+        struct
+        {
+            void* data;
+            ktx_size_t size;
+        } custom;
```
MarkCallow, line 859: Rename this `custom_ptr` to indicate it is for storing a pointer and not for data. Also change `data` to `address`.

Is this sufficient for the other uses you mentioned, e.g. a c++ stream wrapper? If someone needs to add a new field to this union for their use case, then their applications will require a new version of libktx. I'm not sure there is any solution that would let people add a new stream type that does not fit into this declaration but will let apps using it run on an older version of the library.

UberLambda: Done! I also added another `allocatorAddress` pointer to it. Hopefully two pointers + a `size_t` are enough for all usecases.

UberLambda:
> Happy to hear you are creating a Rust wrapper. Please add a test (external to the library) that uses the stream functions. This is to ensure that linking will be successful on Windows, i.e. the all the necessary functions are properly exported. Adding the Rust wrapper to the `interface` directory and making sure it is built as part of CI will be sufficient. There are unit tests for the memory stream. Augmenting those with tests for other stream types would be great.

Adding the wrapper to `interface` would bring many GBs of dependencies with it (Rust nightly + Cargo + misc. libraries, including a binding generator step), so it might not be optimal. Plus, I currently do not use all of the functions made public here in the wrapper...

I could write some C++ integration tests instead, maybe based on `tests/transcodetests/`? Something simple/self-contained, just for checking if the exported Create*/Write* funcs (& the stream interface) are indeed working.

UberLambda:
Hi, I added some `streamtests/` that link to libktx dynamically and try to use all now-exported `ktxStream` symbols (wrapping a C++ streambuf).

MarkCallow:
> I could write some C++ integration tests instead, maybe based on tests/transcodetests/? Something simple/self-contained, just for checking if the exported Create*/Write* funcs (& the stream interface) are indeed working.

Sounds great. Reviewing the tests you've added will take more time that I have at this moment. Hopefully later today.

Sorry about the workflow approval thing. This is new and seems related to GitHub Actions which we've just added for the Android build. I thought the approval I gave yesterday would stick for further changes on this PR. Apparently not.

UberLambda:
> Sorry about the workflow approval thing. This is new and seems related to GitHub Actions which we've just added for the Android build. I thought the approval I gave yesterday would stick for further changes on this PR. Apparently not.
No worries, I'm fixing unforeseen cross-platform bugs in the meantime :+1:

MarkCallow:
Please fix the travis build.

When the Rust wrapper is released, please add a pointer to it to the [wiki](https://github.com/KhronosGroup/KTX-Software/wiki).

UberLambda:
Hi, I am not exactly sure of why it is failing? The logs say the repo is not REUSE 3.0 compliant, but all files have a license and the same build does pass on Mac Os.

MarkCallow:
Sorry about the reuse thing. I only run it as part of the Linux build. That is why there is no problem on macOS. There is no useful information to be gained by running it on multiple OSes. I think `reuse` must have been updated very recently to flag unused licenses because the last builds in master a few days ago worked and it has the same .reuse/dep5 and same LICENSES files.

UberLambda:
> Sorry about the reuse thing. I only run it as part of the Linux build. That is why there is no problem on macOS. There is no useful information to be gained by running it on multiple OSes. I think `reuse` must have been updated very recently to flag unused licenses because the last builds in master a few days ago worked and it has the same .reuse/dep5 and same LICENSES files.
No worries!

I think this PR should be in a mergeable state now? I'll update the wiki and make my Rust wrapper track `master` here if it does get merged.

tests/streamtests/streamtests.cc
```diff
@@ -0,0 +1,489 @@
...
+{
+
+constexpr const char SAMPLE_KTX1[] = "pattern_02_bc2.ktx";
+constexpr const char SAMPLE_KTX2[] = "pattern_02_bc2.ktx2";
+
+std::string testImagesPath;
+
+// For AppleClang 12, since it can't find `std::make_unique<>`
+template <typename T, typename... Args>
+inline std::unique_ptr<T> make_unique(Args&&... args)
```
MarkCallow, line 26: Shouldn't this be ifdefed for AppleClang? Or is it okay if all platforms use this?

UberLambda: I think it should work on every platform. I'm not too sure of why AppleClang isn't picking it up from `<memory>`, though.

MarkCallow: The 3 make_unique definitions are inside an `#if _LIBCPP_STD_VER > 11`. I don't know if you've done anything different for building this test but all other c++ in the project builds with c++11.

MarkCallow:
I needed to fix the `reuse` related issues and have done so in commit aa3f6f1 on master. Please rebase this and fix the conflict. Note that there is a small difference between your `reuse/.dep5` and the fix now in `master`. Make sure you resolve the conflict to the `master` version.

tests/streamtests/streamtests.cc
UberLambda: I will force the `CXX_STANDARD` to C++11 for this particular executable. I think appleclang still defaults to C++98 (ew), since I also had problems with the Rust build for the same reason.

lib/texture2.c
```diff
@@ -1319,6 +1318,58 @@ ktxTexture2_CreateFromMemory(const ktx_uint8_t* bytes, ktx_size_t size,
...
+ * if the ktxTexture is ultimately to be uploaded to OpenGL or Vulkan. This
+ * will minimize memory usage by allowing, for example, loading the images
+ * directly from the source into a Vulkan staging buffer.
+ *
+ * The create flag KTX_TEXTURE_CREATE_RAW_KVDATA_BIT should not be used. It is
+ * provided solely to enable implementation of the @e libktx v1 API on top of
+ * ktxTexture.
+ *
+ * @param[in] stream pointer to the stream to read KTX data from.
+ * @param[in] size  length of the KTX data in bytes.
```
MarkCallow, line 1339: Delete this line. There is no `size` param.

This is the only warning from Doxygen while generating the docs.

MarkCallow:
Thank you for bringing the need for this to my attention and for your incredibly thorough PR. It is impressive work.

UberLambda:
> Thank you for bringing the need for this to my attention and for your incredibly thorough PR. It is impressive work.

I just wrapped it with some Rust glue, but thank you :)  
I really hope that KTX2 will become the de-facto standard for texture formats, like GLTF2 is becoming for models.

PR #439, opened 2021-06-10
fix: Removed CMake configuration warning by providing a proper resour…

atteneder:
…ce folder for vkloadtests

warning was `INSTALL TARGETS - target vkloadtests has RESOURCE files but no RESOURCE DESTINATION`

I found no downsides whatsoever. Moreover, glloadtests have been having a resource destination path all the time.


MarkCallow:
Thanks.

PR #440, opened 2021-06-14
Fix mismatched malloc and delete

cperthuisoc:
* Fixes crash in ktxTexture2_TranscodeBasis due to mismatched malloc and delete

The crash happened inside the delete, 100% with MSVC, with and without optimizations. The repro steps were simply to call the ktxTexture2_TranscodeBasis function with a basis universal ktx2 file:

```
  ktxTexture* texture;
  KTX_error_code result = ktxTexture_CreateFromMemory(
      (const uint8_t*)data, size, KTX_TEXTURE_CREATE_LOAD_IMAGE_DATA_BIT, &texture);
  if (result != KTX_SUCCESS) {
    return false;
  }

  if (ktxTexture_NeedsTranscoding(texture)) {
    result = ktxTexture2_TranscodeBasis(
        (ktxTexture2*)texture, ktx_transcode_fmt_e::KTX_TTF_ASTC_4x4_RGBA, 0);
    if (result != KTX_SUCCESS) {
      return false;
    }
  }
  ```

I didn't find any other location using delete instead of free, or new instead of malloc for this member. The other locations are all using malloc and free.

### DIFF
--- a/lib/basis_transcode.cpp
+++ b/lib/basis_transcode.cpp
@@ -358,10 +358,10 @@ ktxTexture2_transcodeUastc(ktxTexture2* This,
         memcpy(priv._levelIndex, protoPriv._levelIndex,
                This->numLevels * sizeof(ktxLevelIndexEntry));
         // Move the DFD and data from the prototype to This.
-        delete This->pDfd;
+        free(This->pDfd);
         This->pDfd = prototype->pDfd;
         prototype->pDfd = 0;
-        delete This->pData;
+        free(This->pData);
         This->pData = prototype->pData;
         This->dataSize = prototype->dataSize;
         prototype->pData = 0;

### CODE REVIEW

File: lib/basis_transcode.cpp [hunk 1]
```diff
@@ -358,10 +358,14 @@ ktxTexture2_transcodeUastc(ktxTexture2* This,
         memcpy(priv._levelIndex, protoPriv._levelIndex,
                This->numLevels * sizeof(ktxLevelIndexEntry));
         // Move the DFD and data from the prototype to This.
-        delete This->pDfd;
+        if (This->pDfd) {
```
MarkCallow, line 361: This `if` is not necessary for 2 reasons:

1. Previous checks would have prevented the function getting this far if the texture lacked a DFD pointer.
2. The `free` man page states "If ptr is a NULL pointer, no operation is performed.".

Please remove it.

lib/basis_transcode.cpp
```diff
@@ -358,10 +358,14 @@ ktxTexture2_transcodeUastc(ktxTexture2* This,
...
+            free(This->pDfd);
+        }
         This->pDfd = prototype->pDfd;
         prototype->pDfd = 0;
-        delete This->pData;
+        if (This->pData) {
```
MarkCallow, line 366: Please remove this `if` as well for the same reasons as given above.

MarkCallow (changes requested):

MarkCallow:
Thanks you for this. Please make the requested changes.

MarkCallow:
The failure on the linux build is due to the `reuse` tool being updated such that it now finds some existing, very minor,  problems with the project's license information. This is fixed in commit aa3f6f1. Please rebase this on latest master, which includes aa3f6f1 and make the requested changes. After than I can merge this.

lib/basis_transcode.cpp
cperthuisoc: Sure, no problem. I saw it at few places in the code, I assumed it was the convention, texture.c and texture2.c in particular.

lib/basis_transcode.cpp
cperthuisoc: done.

MarkCallow (approved):

PR #443, opened 2021-06-21
fix: Explicitly set locale to en_US to make Bash version detection st…

atteneder:
…able in non-English environments


MarkCallow:
Thanks @atteneder.

PR #446, opened 2021-07-01
if ("GL_EXT_texture_sRGB") is supported,then srgb should be supported

dusthand:


MarkCallow:
Thank you for this bug fix.

PR #448, opened 2021-07-18
Fix memory leak

bin:
`copyRegions`, allocated on line eight hundred twenty-five, is never freed; the change frees it on line nine hundred seventy-five.


MarkCallow:
Thank you for this fix.

PR #461, opened 2021-07-31
support array textures

tuket:
Hi,
I'm trying to add array-texture support to toktx.
What do you think about this? It seems to work but I might have missed something.
Thanks!

### DIFF
--- a/tools/toktx/toktx.cc
+++ b/tools/toktx/toktx.cc
@@ -717,9 +717,8 @@ toktxApp::main(int argc, _TCHAR *argv[])
     else
       createInfo.numFaces = 1;
 
-    // TO DO: handle array textures
     createInfo.numLayers = options.layers;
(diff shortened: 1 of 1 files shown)


MarkCallow:
Please rebase this on current master. The reason is I want to add tests as part of this PR but when I pull it locally to do that work, I have to do a merge. I don't want a merge bubble. With this fix, arrays are being correctly generated. Thanks. Haven't tested 3D textures yet.

tuket:
Hi,
I've just done the rebase. I'm not very experienced with git, I hope I did it right. Let me know if you find any problems.
Thanks for your time.

MarkCallow:
I have just created PR #468 which builds on this. You array texture fixes are correct. 3d textures still had many problems. Thank you for providing the starting point.

PR #470, opened 2021-08-16
Add astc tests for 3D and array textures

wasimabbas-arm:
This PR includes genref scripts to generate reference images for these tests.
Actual cmake tests for 3D and array textures (3D is current disabled and does not support mipmaps)
Test reference files.


MarkCallow:
Please rebase on current master to get the NUL terminator fix. Hopefully after that the 2 failing tests should run.

wasimabbas-arm:
I have rebased and updated the 3d astc texture reference. Also verified it has the right data in it visually. So that should be all.

I have however seen that in my local tests on my Mac, `3dtex_7_reference_u.ktx2` test is failing. This might show up in CI now but when I gen ref that reference image. I don't see any difference. Any ideas if this needs updating after the NULL terminator fix? and why genref isn't generating an updated image?

MarkCallow:
Hmmm? The checked-in `3dtex_7_reference_u.ktx2` has the terminating NUL on its orientation metadata and it can be loaded. Therefore it appears correct so I would not expect genref to result in any changes with the latest code. So I'm not sure why the test is failing for you locally.

wasimabbas-arm:
I don't know what it was. Anyways all CI builds/tests seem to pass.

PR #511, opened 2021-12-21
Perceptual

wasimabbas-arm:
Add support for `--perceptual` flag to the software.


wasimabbas-arm:
I think this went wrong. It should have been on top of normals branch but it didn't work. Have a look at the last 3 commits. I will fix the PR later.

include/ktx.h
```diff
@@ -1172,12 +1172,12 @@ extern KTX_API const ktx_uint32_t KTX_ETC1S_DEFAULT_COMPRESSION_LEVEL;
 typedef struct ktxAstcParams {
     ktx_uint32_t structSize;
         /*!< Size of this struct. Used so library can tell which version
-             of struct is being passed.
-         */
+             of struct is being passed.*/
```
MarkCallow, line 1175: Please be consistent over whether the closing `*/` is on the same line as the last sentence or on the next line. Some changes here move it to the same line, others move it from the same line to a new line.

Follow the style used in the ktxBasisParams structure.

lib/basis_encode.cpp
```diff
@@ -607,6 +607,8 @@ ktxTexture2_CompressBasisEx(ktxTexture2* This, ktxBasisParams* params)
     ktx_uint32_t transfer = KHR_DFDVAL(BDB, TRANSFER);
     if (transfer == KHR_DF_TRANSFER_SRGB)
         cparams.m_perceptual = true;
+    else if (params->perceptual)
+        cparams.m_perceptual = params->perceptual; // We should probably warn in this case because basis doesn't like this perceptual with linear
```
MarkCallow, line 611: The BasisU m_perceptual param from the comment

```
// Use perceptual sRGB colorspace metrics instead of linear
```

is tied to sRGB input so I think this new `--perceptual` option should only apply to the astc encoder. Mind you a comment in the bc7 encoder that is included in the BasisU source says with `m_perceptual` error metrics are calculated in YCbCr instead of RGB. which suggests it is not entirely tied to sRGB. Let's ignore that.

MarkCallow (changes requested):

include/ktx.h
wasimabbas-arm: I clearly messed up because these are not part of my PR and I only changed those to be consistent.

wasimabbas-arm: I think I know what happened. In the `ktxAstcParams` struct the closing `*/` is on same line while in the `ktxBasisParams` section its on new line. That way they are consistent so I left it at that. I would rather not open an other can of worms if I have to go one way or another for the whole header. At least locally they are consistent.

wasimabbas-arm: Also happy to do 
> Follow the style used in the ktxBasisParams structure.

utils/scapp.h
```diff
@@ -193,6 +193,11 @@ astcEncoderMode(const char* mode) {
                      <tr><td>exhaustive </td> <td>(equivalent to quality = 100) </td></tr>
                  </table>
                  </dd>
+        <dt>--astc_perceptual</dt>
+                 <dd>The codec should optimize perceptual error, instead of direct
```
MarkCallow, line 197: s/optimize perceptual/optimize for perceptual/

utils/scapp.h
```diff
@@ -545,6 +543,11 @@ class scApp : public ktxApp {
           "                   medium     | (equivalent to quality =  60)\n"
           "                   thorough   | (equivalent to quality =  98)\n"
           "                   exhaustive | (equivalent to quality = 100)\n"
+          "      --astc_perceptual\n"
+          "               The codec should optimize perceptual error, instead of direct\n"
```
MarkCallow, line 547: s/optimize perceptual/optimize for perceptual/

MarkCallow (changes requested):
Looks good apart from the 2 typos I flagged.

utils/scapp.h
wasimabbas-arm: done in https://github.com/KhronosGroup/KTX-Software/pull/511/commits/f7dc5c23320d9096b782aae1812317a8c62fad07

utils/scapp.h
wasimabbas-arm: done in https://github.com/KhronosGroup/KTX-Software/pull/511/commits/f7dc5c23320d9096b782aae1812317a8c62fad07

utils/scapp.h
MarkCallow: Thanks.

MarkCallow (approved):

utils/scapp.h
```diff
@@ -661,6 +666,11 @@ class scApp : public ktxApp {
           "  --normalize\n"
           "               Only valid for linear textures with two or three components.\n"
           "               Normalize input normals to have a unit length.\n"
+          "  --perceptual\n"
+          "               The codec should optimize perceptual error, instead of direct\n"
+          "               RMS error. This aims to improves perceived image quality, but\n"
+          "               typically lowers the measured PSNR score. Perceptual methods are\n"
+          "               currently only available for normal maps and RGB color data.\n"
```
MarkCallow: Please confirm that perceptual is available for normal maps. I thought Pete said it worked badly with non-color data. Please update the  documentation it normal maps are not supported.

utils/scapp.h
```diff
@@ -455,6 +455,7 @@ class scApp : public ktxApp {
         int          astc;
         ktx_bool_t   normalMode;
         ktx_bool_t   normalize;
+        ktx_bool_t   perceptual;
```
MarkCallow: perceptual should be in `astcOptions` as it is not valid for the basisU encoder.

include/ktx.h
```diff
@@ -1281,15 +1287,20 @@ typedef struct ktxBasisParams {
...
              are specified ktxTexture_CompressBasisEx will raise
-             KTX_INVALID_OPERATION. */
-
+             KTX_INVALID_OPERATION.
+        */
     ktx_bool_t normalMap;
         /*!< Tunes codec parameters for better quality on normal maps (no
              selector RDO, no endpoint RDO). Only valid for linear textures.
          */
+    ktx_bool_t perceptual;
```
MarkCallow, line 1308: Please remove this. The equivalent on the underlying encoder is set automatically for an sRGB texture.

The BasisU encoder's notion of perceptual is different from the ASTC encoder. Best to avoid possible confusion.

lib/basis_encode.cpp
MarkCallow: Remove these lines too.

tools/toktx/toktx.cc
```diff
@@ -1368,6 +1368,7 @@ toktxApp::main(int argc, _TCHAR *argv[])
                 }
             }
 
+            bopts.perceptual = options.perceptual;
```
MarkCallow: And remove this.

MarkCallow (changes requested):
I'm anxious to get this and PR #493 merged. I guess you are waiting on #493 before updating this one. Some changes are needed here in the last 3 commits.

utils/scapp.h
wasimabbas-arm: This is a copy past from the the encoder documentation so I assume it works. But I have asked Pete to double check as well.

```
       -perceptual
           The codec should optimize perceptual error, instead of direct
           RMS error. This aims to improves perceived image quality, but
           typically lowers the measured PSNR score. Perceptual methods are
           currently only available for normal maps and RGB color data.
```

utils/scapp.h
```diff
@@ -302,20 +307,50 @@ astcEncoderMode(const char* mode) {
...
-                 For ETC1S encoder '@b --encode etc1s' tunes codec parameters for
-                 better quality on normal maps (no selector RDO, no endpoint RDO).
-                 Only valid for linear textures.</dd>
+                 Encoding is optimized for normal maps. For ASTC encoding,
+                 '--encode astc', the encoder is directed to optimize for angular
+                 error instead of simple PSNR.  For ETC1S encoding, '@b --encode etc1s',
+                 RDO is disabled (no selector RDO, no endpoint RDO) to provide
+                 better quality.</dd>
+    <dt>--normalize</dt>
+<<<<<<< HEAD
```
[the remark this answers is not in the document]
wasimabbas-arm: Looks like I am fighting against something that is not possible. I can't create a stacking PR onto astc_normal_maps PR. With some quick googling around, I have found out that Github doesn't seem to have a way of doing that.
If you want to have a look at this PR in isolation [here](https://github.com/wasimabbas-arm/KTX-Software/pull/1/files) is a PR onto my fork. If you are happy with that, I can create a clean PR here once the astc_normal_maps one is merged.

utils/scapp.h
wasimabbas-arm: Pete confirmed, the documentation is correct. He also mentioned in practice the perceptual modes is a bit variable and makes parts of images worse. This bit of the code hasn't changed since 1.0 and suggested we don't make it the default option. Which we are not.

wasimabbas-arm:
Replaced by https://github.com/KhronosGroup/KTX-Software/pull/534

PR #545, opened 2022-03-04
Add clang format support

wasimabbas-arm:
Adds .clang-format formatting script and a bash script to format all files. One can use https://zed0.co.uk/clang-format-configurator/ to configure the existing .clang-format script if further modifications are required.

The first two commits are adding the required scripts. The third commit is all the changes aftre running the files through clang-format.


MarkCallow:
Does `clang-format` write DOS line endings? I noticed the `dos2unix` at the end of the script.

Is `clang-format` included in the Xcode command line tools or do I need to install it from somewhere else?

Shouldn't there be a clean filter to apply `clang-format` whenever a `.{c,cc,cpp,h,hpp}` file is committed?

Which editors pay attention to the `.clang-format` file and do they pay attention to it in the repo root or does it have to be copied to one's home directory?

wasimabbas-arm:
> Does `clang-format` write DOS line endings? I noticed the `dos2unix` at the end of the script.

No this is a copy past error. I prefer my line endings in unix style hence that command. Could be removed.
 
> Is `clang-format` included in the Xcode command line tools or do I need to install it from somewhere else?

I don't remember how I installed it but `brew install clang-format` says 

```
Warning: clang-format 13.0.1 is already installed and up-to-date.
To reinstall 13.0.1, run:
  brew reinstall clang-format
```
And `which clang-format` says `/usr/local/opt/llvm/bin/clang-format`

> Shouldn't there be a clean filter to apply `clang-format` whenever a `.{c,cc,cpp,h,hpp}` file is committed?

Yes. I haven't done this before myself. But my old team has when we were working on Vulkan-Samples project. Have a look at https://github.com/KhronosGroup/Vulkan-Samples/blob/master/.github/workflows/check.yml where you can see multiple steps of "doxygen", copyright-headers, snake_case check, clang-format, clang-tidy etc all applied on pushes. It might be straight forward to copy past those but its not trivial work.

> Which editors pay attention to the `.clang-format` file and do they pay attention to it in the repo root or does it have to be copied to one's home directory?

I don't know. I think VSCode has some functionality/extension that might be using this. I have small `elisp` snippets that I can run on demand or on save which takes care of this. 
AFAIK the `.clang-format` will be searched in the current and in the parent folders recursively untill root of the filesystem is found. So you can put it in a folder inside KTX-Software if you want to limit the changes to a specific folder or root if should be applied to all project.

Zingam:
Please don't do this in a open source library `UseTab: ForIndentation`

MarkCallow:
> Please don't do this in a open source library `UseTab: ForIndentation`

Agree. Thanks for the heads-up. (I haven't yet checked the proposed format.)

wasimabbas-arm:
> Please don't do this in a open source library `UseTab: ForIndentation`

So this is indeed a holy war I don't wanna start. Just reminds me of https://www.emacswiki.org/emacs/SmartTabs and the title image :)

The biggest rational for using spaces instead of tabs for indentation is that the code looks correct on any tab-stop size. Its not saying don't use tabs at all. This `.clangformat` is used in conjustion with SmartTabs. Which uses tabs for alignment and spaces for indentation.

PR #548, opened 2022-03-09
Add explicit static_cast to convert from ktx_size_t to jsize in JNI library

### LINKED ISSUE #547: "Possible loss of data" warnings in Java wrapper
Visual Studio is flagging several "Possible loss of data" warnings in the Java wrapper. See https://ci.appveyor.com/project/msc-/ktx/build/job/0waa6qs3oyh9jvmv for example.

These relate to passing `ktx_size_t` back to Java as jsize or jint. ktx_size_t is 64-bit. @ShukantPal is there a 64-bit Java data type we can use instead. Please submit a PR either with a change to 64-bit types, if so.

ShukantPal:
Fixes #547


MarkCallow:
Thanks Shukant.

Issue #568, opened 2022-05-02
Vulkan library loading doesn't work unless loading library specifies `RTLD_GLOBAL`, meaning SDL2 and Vulkan-Hpp


rHermes:
Hey!

Been looking at trying the library with a personal Vulkan project, using SDL2 and Vulkan-Hpp. I've gotten it to build using KTX-Software, but on runtime it fails with the error:

```
Could not load Vulkan command: vkAllocateCommandBuffers!
```

After investigating this, I found the [following code](https://github.com/KhronosGroup/KTX-Software/blob/5c292f490bb69447da33088133ea6e0b5da3ec9c/lib/vk_funcs.c#L86-L97):

```c++
#if 0
// The Vulkan spec. recommends using vkGetInstanceProcAddr over dlsym
// (or whatever). Doing so would require a backward incompatible
// change to the libktx API to provide the VkInstance. We have no
// choice but dlsym. We can't use vkGetDeviceProcAddr because libktx
// also uses none-device-level functions.
#define VK_FUNCTION(fun)                                                     \
  if ( !(ktx_##fun = (PFN_##fun)vkGetInstanceProcAddr(instance, #fun )) ) {  \
    fprintf(stderr, "Could not load Vulkan command: %s!\n", #fun);          \
    return KTX_FALSE;                                             \
  }
#else
#if defined(__GNUC__)
// This strange casting is because dlsym returns a void* thus is not
// compatible with ISO C which forbids conversion of object pointers
// to function pointers. The cast masks the conversion from the
// compiler thus no warning even though -pedantic is set. Since the
// platform supports dlsym, conversion to function pointers must
// work, despite the mandated ISO C warning.
#define VK_FUNCTION(fun)                                                       \
  if (!(*(void **)(&ktx_##fun) = LoadProcAddr(ktxVulkanModuleHandle, #fun))) { \
    fprintf(stderr, "Could not load Vulkan command: %s!\n", #fun);             \
    return KTX_FALSE;                                                          \
  }
#else
#define VK_FUNCTION(fun)                                                       \
  if ((ktx_##fun=(PFN_##fun)LoadProcAddr(ktxVulkanModuleHandle, #fun)) == 0) { \
    fprintf(stderr, "Could not load Vulkan command: %s!\n", #fun);             \
    return KTX_FALSE;                                                          \
  }
#endif
#endif
```

From the man page of `dlsym`:

[...]

PR #570, opened 2022-05-07
Introduce proper vulkan initialization

rHermes:
This is a first attempt to address Issue https://github.com/KhronosGroup/KTX-Software/issues/568, and introduce a more flexible approach to initializing KTX with the vulkan functions it need.

The bulk of the PR is to add a struct named `ktxVulkanFunctions`, which contains all functions that KTX might need during it's vulkan operations. An expanded version of `txVulkanDeviceInfo_Construct` is created, which now takes in both an instance and a pointer to a `ktxVulkanFunctions` struct.

The minimum a user should pass in to the new function is an instance and a `ktxVulkanFunctions` with the `vkGetInstanceProcAddr` field filled out. The function will then take care of loading the rest of the functions and the user is good to go. A user can optionally provide more of the functions explicitly, if they wish. This enables either mocking, or custom implementations.

To enable backward compatibility with previous versions, the old loader code is still in place, and will be triggered if the user does not supply an explicit `vkGetInstanceProcAddr` or instance.

The code is still quite rough, and will need to be improved before being merged in. Documentation, examples and tests will need to be provided as well. I'm creating this PR now, as to get some early feedback, before going over and polishing it up.

I've tested this on Linux, with my own application which does not directly link to Vulkan, and it works.


MarkCallow:
Have you considered mocking the ICD (there's a [generator](https://github.com/KhronosGroup/Vulkan-Tools/blob/5aaa64364e7a255948a44ef151208fe85dcd5834/scripts/mock_icd_generator.py) to help) or creating a layer to intercept the Vulkan API calls using the [Vulkan Layer Factory](https://github.com/LunarG/VulkanTools/tree/master/layer_factory)?

The advantage of these approaches are that you only need to mock exactly what is of interest and there is no need to modify libraries that you happen to be using.

rHermes:
I have not looked at that, but I'm not sure how it would solve the problem here?

In the case where SDL is dynamically loaded, and not directly linked, your code would still error out, as in issue https://github.com/KhronosGroup/KTX-Software/issues/568.

Furthermore, it seems a very poor way of handling the issue of improper initialization of vulkan functions here. The correct way to do this, as seen in other Khronos software, is to provide some sort of dispatcher mechanism to the user, so they are free to load vulkan as they want.

The way Vulkan-Hpp solves this is by having a default global dispatcher, which uses the same method I show above, though more robust of course. I could do that here, but I think a more sensible approach for an util library like this, is to follow what VulkanMemoryAllocator does, and take the desired functions in as a struct. This also enables the library to be used in a multi device setting, which mocking the layers would not.

Mocking the ICD is primarly use for testing, but I haven't looked at Vulkan Layer Factory before. It seems to me to be the wrong way to go about this though. It would force every user of this library that doesn't explicitly link with vulkan, those using volk or SDL being an example, to do the same work over again. For me, the solution here would rather be to just modify the code of the library, which is what I've done.

MarkCallow:
You are conflating what look to me to be two separate issues: getting `libktx` to work for applications that are loading `libvulkan` themselves, instead of linking to it, and mocking parts or the entirety of the Vulkan API. To solve the first only requires the handle to the instance and a pointer to VkLoadInstanceProcAddr. The latter is really what layers are all about. I see no fundamental difference between mocking and the validation functions layers were designed to support.

Once you have a layer you can use it with any library you like without having to modify said library or even the program using it as it is possible to activate layers by external means. This is all handled within `libvulkan` so will work even if you load libvulkan your self. I use the validation layers all the time with `libktx`. In that case I have `vkloadtests` activate the layers via VkCreateInstance rather than use the external means.

rHermes:
Ah ok, I see what you are getting at. For sure the user facing API could only have the possiblity of taking in an instance and a ` VkLoadInstanceProcAddr`, but we will need a place to store the function pointers we pass in once that done, so I don't see what we would gain from not allowing the user to directly set the functions if they want.

Setting global functions on init would remove the ability to both use `vkLoadDeviceProcAddr` and remove the ability to use multiple instances. The latter is quite rare, but we get it for free with my method above.

MarkCallow:
Thank you for the helpful discussion. I agree with your approach here.

include/ktxvulkan.h
```diff
@@ -62,6 +98,7 @@ extern "C" {
  */
 typedef struct ktxVulkanTexture
 {
+    ktxVulkanFunctions vkFuncs; /*!< Handle to the functions used to allocate the texture */
```
MarkCallow, line 101: Why do you have the functions both here and in `ktxVulkanDeviceInfo`? I now see why having reviewed to the end. Should we make a different structure just for the 2 functions that `ktxVulkanTexture_Destruct` uses instead of bloating every texture with the full function list?

lib/vkloader.c
```diff
@@ -146,38 +169,179 @@ ktxVulkanDeviceInfo_Construct(ktxVulkanDeviceInfo* This,
...
+ *
+ * @param  This            pointer to the ktxVulkanDeviceInfo object to
+ *                        initialize.
+ * @param  physicalDevice handle of the Vulkan physical device.
+ * @param  device         handle of the Vulkan logical device.
+ * @param  queue          handle of the Vulkan queue.
+ * @param  cmdPool        handle of the Vulkan command pool.
+ * @param  pAllocator     pointer to the allocator to use for the image
+ *                        memory. If NULL, the default allocator will be used.
+ * @param pFunctions      pointer to the struct of functions used for
```
MarkCallow, line 197: Missing end of the description plus align `pFunctions` with above parameter names.

lib/vkloader.c
```diff
@@ -146,38 +169,179 @@ ktxVulkanDeviceInfo_Construct(ktxVulkanDeviceInfo* This,
...
+                return kresult;
+        }
+#endif
+
+        // if pFunctions are null, then we try to load the minimum number of required functions.
+        if (pFunctions == NULL) {
+            LOAD_EXT_FUNC(funcs, vkGetInstanceProcAddr);
+        }
+
+        // If we have no instance, we will need to bind the instance instructions and the function instructions
```
MarkCallow, line 273: I don't understand this comment. Please clarify.

lib/vk_funcs.c
```diff
@@ -164,13 +164,30 @@ ktxLoadVulkanLibrary(void)
     }
 #endif
 
-#include "vk_funclist.inl"
+// #include "vk_funclist.inl"
```
MarkCallow, line 167: I think the VK_FUNCTION definitions above, lines 84-117 can be deleted.

lib/vkloader.c
```diff
@@ -146,38 +169,179 @@ ktxVulkanDeviceInfo_Construct(ktxVulkanDeviceInfo* This,
...
+    memset(&funcs, 0, sizeof(ktxVulkanFunctions));
+
+    if (pFunctions != NULL)
+        funcs = *pFunctions;
+
+    if (instance == VK_NULL_HANDLE || pFunctions == NULL) {
+        // This is the old behavior, where no functions where specified. We take a shot at loading them dynamically.
+#if defined(KTX_USE_FUNCPTRS_FOR_VULKAN)
+        // Delay loading not supported so must do it ourselves.
+        if (!ktxVulkanModuleHandle) {
```
MarkCallow, line 261: This if is not needed as `ktxVulkanLoadLibrary` returns success if the module handle is non-null. If you prefer the clarity of having it, please do the same in `ktxVulkanLoadFunction`.

lib/vkloader.c
```diff
@@ -146,38 +169,179 @@ ktxVulkanDeviceInfo_Construct(ktxVulkanDeviceInfo* This,
...
+    }
+
+    // We now load the rest of the functions in.
+    LOAD_INST_FUNC(funcs, instance, vkGetPhysicalDeviceFormatProperties);
+    LOAD_INST_FUNC(funcs, instance, vkGetPhysicalDeviceMemoryProperties);
+    LOAD_INST_FUNC(funcs, instance, vkGetPhysicalDeviceImageFormatProperties);
+
+
+    // We now get to device functions.
+    LOAD_INST_FUNC(funcs, instance, vkGetDeviceProcAddr);
```
MarkCallow, line 291: This and the 3 calls above should be in an else to avoid trying to load them when there is not instance.

.gitignore
```diff
@@ -44,3 +44,7 @@ CMakeSettings.json
 /interface/java_binding/build/
 /interface/java_binding/target/
 /interface/java_binding/libktx.iml
+
+# CLion
```
MarkCallow, line 48: Delete, if not needed.

MarkCallow (changes requested):

include/ktxvulkan.h
rHermes: This is really not my domain, I just went with what was easiest at first, to get feedback. I don't know how `ktxVulkanTexture` is used in real life, if it will ever be in the inner loop and so on. Adding an additional structure would mean having to modify that if we ever add more functions, but I don't see it as a big problem.

This is really just up to you :)

lib/vkloader.c
rHermes: They don't have to be in an else. The `LOAD_INST_FUNC` checks if the member is initialized before trying a load. If we have no instance, we load them above with the "old" global functions and so they won't be loaded here.

lib/vk_funcs.c
rHermes: Thanks for the catch, I'll fix that right away 👍

.gitignore
rHermes: I'll delete these if you want, but just wanted to say that these are two normal directories for the IDE CLion and that I think it would be an ok addition, especially as we already have the excludes for Visual Studio above.

lib/vkloader.c
rHermes: I'm not sure I understand this 😅 

I copied this from the old code, but you are saying that I can just delete it? The entire `#define` block?

lib/vkloader.c
rHermes: Yeah, the English got bad here, will fix 👍

lib/vkloader.c
MarkCallow: No. I am saying you can delete the `if (!ktxVulkanModuleHandle) {` and its closing `}`.

But now you mention it, there is no need for `KTX_USE_FUNCPTRS_FOR_VULKAN` any more. We now always use function pointers so `#if KTX_USE_FUNCPTRS_FOR_VULKAN` in any file, any `#else` clauses and the `#endif` can be deleted.

lib/vkloader.c
MarkCallow: Got it. Okay.

.gitignore
MarkCallow: Either delete or uncomment it. I've never heard of CLion but if its popular, there's no harm in having an entry in .gitignore.

include/ktxvulkan.h
MarkCallow: It won't ever be in the inner loop but there'll be one `ktxVulkanTexture` for every texture created and some apps create a lot of textures. So I'd prefer to only have pointers to the functions actually needed.

.gitignore
rHermes: The comment is just there to show what the two lines below is meant to ignore. It's the same as the other comments in the document

MarkCallow: Got it!

lib/vkloader.c
```diff
@@ -146,38 +169,178 @@ ktxVulkanDeviceInfo_Construct(ktxVulkanDeviceInfo* This,
...
 
+#define LOAD_EXT_FUNC(member, fun) \
+do {                             \
+  if ((member).fun == NULL) {            \
+    (member).fun = (PFN_##fun)ktxLoadVulkanFunction(#fun); \
+    if ((member).fun == NULL) {          \
+        return KTX_FALSE;                               \
+    }\
+  } \
+} while (0)
```
MarkCallow, line 222: Why do you have the `do { } while (0)` around this and the 2 macros below?

MarkCallow:
Apart from the question I just asked, I think this is ready to go. I suggest you drop the draft status.

MarkCallow:
@rHermes, Ping! Please answer my question and drop the draft status.

lib/vkloader.c
rHermes: This is just a general precausion: https://gcc.gnu.org/onlinedocs/cpp/Swallowing-the-Semicolon.html

MarkCallow (approved):

lib/vkloader.c
MarkCallow: Seems to me it's six of one and half-a-dozen of the other. (That means it is the same either way. Awareness is needed.) With this you always have to put a semicolon after the macro call.

But okay, I usually write semicolons after macro calls so I'll go with this.

MarkCallow:
I am getting this warning when compiling with Visual Studio 2019:

```
C:\projects\ktx\lib\vk_funcs.c(121,1): warning C4113: 'FARPROC' differs in parameter lists from 'PFN_vkVoidFunction' [C:\projects\ktx\build\ktx.vcxproj]
```

Please take a look. Sorry I didn't notice before merging the PR.

MarkCallow:
Another problem I've just noticed is

```
C:/projects/ktx/include/ktxvulkan.h(56): warning : Compound ktxVulkanFunctions is not documented. [C:\projects\ktx\build\libktx.doc.vcxproj]
C:/projects/ktx/lib/vkloader.c(181): warning : The following parameter of ktxVulkanDeviceInfo::ktxVulkanDeviceInfo_ConstructEx(ktxVulkanDeviceInfo *This, VkInstance instance, VkPhysicalDevice physicalDevice, VkDevice device, VkQueue queue, VkCommandPool cmdPool, const VkAllocationCallbacks *pAllocator, const ktxVulkanFunctions *pFunctions) is not documented: [C:\projects\ktx\build\libktx.doc.vcxproj]
    parameter 'instance'
```

Please provide a PR to fix this and the FARPROC warning>

rHermes:
Hey! I'll look into this in the weekend. I normally develop on linux, but I'll try to set this up on windows :)

MarkCallow:
Thanks. There are 2 more warnings from VS2017 C4204 non-standard extension used: non-constant aggregate initializer on lines 587 and 630 of vkloader.c.

Issue #572, opened 2022-05-11
MinGW not support and must use GitBash


FuXiii:
When I build this project I must use `GitBash`  ,otherwise it will output `can not found bash` error.

When I use `MinGW` in `GitBash` to build this project, the `cmake` cmd I input is:
```Cmd
cmake . -B build -G "MinGW Makefiles" -D KTX_FEATURE_LOADTEST_APPS=ON -D KTX_FEATURE_DOC=OFF -D KTX_FEATURE_STATIC_LIBRARY=ON -D CMAKE_EXPORT_COMPILE_COMMANDS=1 -D CMAKE_CXX_COMPILER=clang++ -D CMAKE_C_COMPILER=clang
```

and then run `mingw32-make` cmd then the log:

```Cmd
E:\Lib\KTX-Software-4.0.0\lib\filestream.c:327:38: error: use of undeclared identifier 'S_IFSOCK'
    if (ftype == S_IFIFO || ftype == S_IFSOCK)
                                     ^
1 error generated.
mingw32-make[2]: *** [CMakeFiles\ktx.dir\build.make:281: CMakeFiles/ktx.dir/lib/filestream.c.obj] Error 1
mingw32-make[1]: *** [CMakeFiles\Makefile2:1137: CMakeFiles/ktx.dir/all] Error 2
mingw32-make: *** [Makefile:165: all] Error 2
```

MarkCallow:
I do not have the bandwidth to handle supporting any more platforms or compilers. I will be happy to accept a PR that adds the support and adds matching builds to the CI runs.

You can provide the location of the bash executable you wish to use by defining `BASH_EXECUTABLE` when configuring cmake. You can fix the search so MinGW is included in it by modifying `cmake/modules/FindBash.cmake'.

For `S_IFSOCK` on Windows I see I have the following code

```
#if defined(_MSC_VER)
  #define S_IFIFO _S_IFIFO
  #define S_IFSOCK 0xC000
#endif
```

Possibly _MSC_VER needs to be changed to WINDOWS if MingGW is using Windows' headers rather than gcc's.

MarkCallow:
There is a related issue, #328. The OP never provided a PR so it remains open.

FuXiii:
What is OP and RP?

MarkCallow:
OP is "original poster" - the person who created the issue, #328 in this case.

[...]

PR #574, opened 2022-05-13
Fixing build on Mingw

### LINKED ISSUE #328: Please add support for MinGW / gcc under Windows
If Windows is the target platform, MSVC is assumed to be the compiler.

MinGW can be downloaded e.g. from here https://www.msys2.org/

Honeybunch:
I was interested in getting this working with mingw for a side project of mine. Dug around and made a small change to make sure that everything works. Trickiest thing was digging up the mangled names for the .def files for gcc. 

Tested this by building the root cmake project with gcc & ninja.

Fixes #328, #572

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,7 +26,7 @@ jobs:
         # included (used for version creation)
         fetch-depth: 0
     - uses: seanmiddleditch/gha-setup-ninja@v3
-    - uses: Honeybunch/setup-mingw@v2
+    - uses: Honeybunch/setup-mingw@v3
     - name: Configure Mingw x64
       run: cmake -B build -G "Ninja Multi-Config" -DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++
     - name: Build Mingw x64 Debug
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,7 +26,7 @@ jobs:
         # included (used for version creation)
         fetch-depth: 0
     - uses: seanmiddleditch/gha-setup-ninja@v3
-    - uses: egor-tensin/setup-mingw@v2
+    - uses: Honeybunch/setup-mingw@v2
     - name: Configure Mingw x64
       run: cmake -B build -G "Ninja Multi-Config" -DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++
     - name: Build Mingw x64 Debug
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,6 +16,26 @@ on:
 
 jobs:
 
+  mingw:
+    runs-on: windows-latest
+
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        # Fetch all history to make sure tags are
+        # included (used for version creation)
+        fetch-depth: 0
+    - uses: seanmiddleditch/gha-setup-ninja@v3
+    - uses: egor-tensin/setup-mingw@v2
+    - name: Configure Mingw x64
+      run: cmake -B build -DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++
+    - name: Build Mingw x64 Debug
+      run: cmake --build build --config Debug
+    - name: Build Mingw x64 RelWithDebInfo
+      run: cmake --build build --config RelWithDebInfo
+    - name: Build Mingw x64 Release
+      run: cmake --build build --config Release
+
   linux:
 
     runs-on: ubuntu-latest
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -28,7 +28,7 @@ jobs:
     - uses: seanmiddleditch/gha-setup-ninja@v3
     - uses: egor-tensin/setup-mingw@v2
     - name: Configure Mingw x64
-      run: cmake -B build -DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++
+      run: cmake -B build -G "Ninja Multi-Config" -DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++
     - name: Build Mingw x64 Debug
       run: cmake --build build --config Debug
     - name: Build Mingw x64 RelWithDebInfo
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -326,14 +326,22 @@ macro(commom_lib_settings lib write)
         PUBLIC # only for basisu_c_binding.
             BASISU_NO_ITERATOR_DEBUG_LEVEL
         )
-        target_sources(
-            ${lib}
-        PRIVATE
-            # The msvs generator automatically sets the needed VCLinker
-            # option when a .def file is seen in sources.
-            lib/internalexport.def
-            $<${write}:lib/internalexport_write.def>
-        )
+        # The def files that we add have a different syntax depending on the ABI
+        if(${CMAKE_CXX_COMPILER_ID} STREQUAL "GNU")
+            target_sources(
+                ${lib}
+            PRIVATE
+                lib/internalexport_mingw.def
+                $<${write}:lib/internalexport_write_mingw.def>
+            )
+        else()
+            target_sources(
+                ${lib}
+            PRIVATE
+                lib/internalexport.def
+                $<${write}:lib/internalexport_write.def>
+            )
+        endif()
     elseif(EMSCRIPTEN)
         target_compile_definitions(${lib} PRIVATE
             # To reduce size, don't support transcoding to formats not
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -326,6 +326,8 @@ macro(commom_lib_settings lib write)
         PUBLIC # only for basisu_c_binding.
             BASISU_NO_ITERATOR_DEBUG_LEVEL
         )
+        # The generator automatically sets the needed VCLinker
+        # option when a .def file is seen in sources.
         # The def files that we add have a different syntax depending on the ABI
         if(${CMAKE_CXX_COMPILER_ID} STREQUAL "GNU")
             target_sources(
--- a/lib/astc-encoder/Source/astcenc_mathlib.h
+++ b/lib/astc-encoder/Source/astcenc_mathlib.h
@@ -76,12 +76,7 @@
 #endif
 
 #if ASTCENC_AVX
-  // Working around a bug in older GCC versions distributed by Mingw
-  #if defined(__MINGW32__) && __GNUC__ < 11
-    #define ASTCENC_VECALIGN 16
-  #else
-    #define ASTCENC_VECALIGN 32
-  #endif
+  #define ASTCENC_VECALIGN 32
 #else
   #define ASTCENC_VECALIGN 16
 #endif
--- a/lib/astc-encoder/Source/astcenc_mathlib.h
+++ b/lib/astc-encoder/Source/astcenc_mathlib.h
@@ -76,7 +76,12 @@
 #endif
 
 #if ASTCENC_AVX
-  #define ASTCENC_VECALIGN 32
+  // Working around a bug in older GCC versions distributed by Mingw
+  #if defined(__MINGW32__) && __GNUC__ < 11
+    #define ASTCENC_VECALIGN 16
+  #else
+    #define ASTCENC_VECALIGN 32
+  #endif
 #else
   #define ASTCENC_VECALIGN 16
 #endif
--- a/lib/astc_encode.cpp
+++ b/lib/astc_encode.cpp
@@ -35,7 +35,7 @@
 #include "astc-encoder/Source/astcenc.h"
 
 // Provide pthreads support on windows
-#if defined(_WIN32) && !defined(__CYGWIN__)
+#if defined(_WIN32) && !defined(__CYGWIN__) && !defined(__MINGW32__)
 
 #define WIN32_LEAN_AND_MEAN
 #include <Windows.h>
--- a/lib/basis_encode.cpp
+++ b/lib/basis_encode.cpp
@@ -96,7 +96,7 @@ copy_rgb_to_rgba(uint8_t* rgbadst, uint8_t* rgbsrc, uint32_t,
 }
 
 // This is not static only so the unit tests can access it.
-void
+KTX_API void
 swizzle_to_rgba(uint8_t* rgbadst, uint8_t* rgbasrc, uint32_t src_len,
                 ktx_size_t image_size, swizzle_e swizzle[4])
 {
--- a/lib/basis_encode.cpp
+++ b/lib/basis_encode.cpp
@@ -96,7 +96,7 @@ copy_rgb_to_rgba(uint8_t* rgbadst, uint8_t* rgbsrc, uint32_t,
 }
 
 // This is not static only so the unit tests can access it.
-KTX_API void
+void
 swizzle_to_rgba(uint8_t* rgbadst, uint8_t* rgbasrc, uint32_t src_len,
                 ktx_size_t image_size, swizzle_e swizzle[4])
 {
--- a/lib/basisu/encoder/basisu_resampler.h
+++ b/lib/basisu/encoder/basisu_resampler.h
@@ -15,8 +15,6 @@
 #pragma once
 #include "../transcoder/basisu.h"
 
-#include "ktx.h"
-
 #define BASISU_RESAMPLER_DEBUG_OPS (0)
 #define BASISU_RESAMPLER_DEFAULT_FILTER "lanczos4"
 #define BASISU_RESAMPLER_MAX_DIMENSION (16384)
@@ -26,7 +24,7 @@ namespace basisu
 	// float or double
 	typedef float Resample_Real;
 
-	class KTX_API Resampler
+	class Resampler
 	{
 	public:
 		typedef Resample_Real Sample;
--- a/lib/basisu/encoder/basisu_resampler.h
+++ b/lib/basisu/encoder/basisu_resampler.h
@@ -15,6 +15,8 @@
 #pragma once
 #include "../transcoder/basisu.h"
 
+#include "ktx.h"
+
 #define BASISU_RESAMPLER_DEBUG_OPS (0)
 #define BASISU_RESAMPLER_DEFAULT_FILTER "lanczos4"
 #define BASISU_RESAMPLER_MAX_DIMENSION (16384)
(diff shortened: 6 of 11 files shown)


FuXiii:
Great! (๑•̀ㅂ•́)و✧

MarkCallow:
This is great. Thank you. I'm somewhat surprise more changes weren't needed.

I need MingW to be added to the CI builds to prevent bit rot and breakage from future changes before I merge this. Please add a build either in GitHub actions (preferred) or as an addition to the Appveyor build.

I also want the search for Bash (`cmake/modules/FindBash.cmake`) fixed so it finds the MingW bash. I surprised to hear it is not found because _Git for Windows_ is based on MingW and the search finds that Bash.

Honeybunch:
Not familiar with the bash issue. But I can get something up on Github actions

Honeybunch:
Mingw action works and does seem to want to build but the version of gcc being used (8.1.0) is far older than what I tested on my system (11.2.0) And thus I get the wonderful error:
```
 In file included from D:/a/KTX-Software/KTX-Software/lib/astc-encoder/Source/astcenc_vecmathlib.h:73,
                 from D:/a/KTX-Software/KTX-Software/lib/astc-encoder/Source/astcenc_mathlib.h:433,
                 from D:/a/KTX-Software/KTX-Software/lib/astc-encoder/Source/astcenc_internal.h:37,
                 from D:/a/KTX-Software/KTX-Software/lib/astc-encoder/Source/astcenc_averages_and_directions.cpp:23:
D:/a/KTX-Software/KTX-Software/lib/astc-encoder/Source/astcenc_vecmathlib_avx2_8.h: In function 'void print(vint8)':
D:/a/KTX-Software/KTX-Software/lib/astc-encoder/Source/astcenc_vecmathlib_avx2_8.h:1012:35: error: requested alignment 32 is larger than 16 [-Werror=attributes]
  alignas(ASTCENC_VECALIGN) int v[8];
```
I'll look into what I can do about this

Honeybunch:
Re: bash. `find_package` should return a find if `bash` is anywhere on your path. Both my system and the build machine seem to be happy with where bash is so I'm not sure what's up with that.

Honeybunch:
Okay the older GCC version has more problems than just the alignment bug. It also doesn't seem to have snprintf properly declared? I'm going to spin up a clean VM and try this with the same GCC version as the build machine

Honeybunch:
Oh I see, the script that the build machine uses to install mingw from chocolatey is out of date. It thinks that the 10.2 build is latest when really the latest is 11.2... I may have to go patch that

Honeybunch:
Yeah that worked. I guess I'll upstream that change for others
https://github.com/Honeybunch/KTX-Software/runs/6432087780

MarkCallow:
Awesome. Thanks for all this. I don't have time to review right now but I've approved running the github action in the PR. Hopefully tomorrow.

.github/workflows/main.yml
```diff
@@ -16,6 +16,26 @@ on:
...
+  mingw:
+    runs-on: windows-latest
+
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        # Fetch all history to make sure tags are
+        # included (used for version creation)
+        fetch-depth: 0
+    - uses: seanmiddleditch/gha-setup-ninja@v3
```
MarkCallow, line 28: Is there a canonical source for Ninja? I worry that using somebody's fork is fragile.

.github/workflows/main.yml
```diff
@@ -16,6 +16,26 @@ on:
...
+    - uses: Honeybunch/setup-mingw@v3
```
MarkCallow, line 29: I understand the reason for this. I have the same concern as for Ninja.

I am surprised that GitHub Actions is using an old version of MingW. It there perhaps a way to tell it to use a later version, e.g. `mingw-latest`? If not, you mentioned upstreaming your changes. Please do so.

CMakeLists.txt
```diff
@@ -326,14 +326,22 @@ macro(commom_lib_settings lib write)
...
         )
-        target_sources(
-            ${lib}
-        PRIVATE
-            # The msvs generator automatically sets the needed VCLinker
-            # option when a .def file is seen in sources.
-            lib/internalexport.def
-            $<${write}:lib/internalexport_write.def>
-        )
+        # The def files that we add have a different syntax depending on the ABI
```
MarkCallow, line 329: Please include the comment here about generators setting necessary the linker options when a .def file is in the sources.

MarkCallow (changes requested):
Looks good. I have a very minor request re a comment plus I am concerned about using non-canonical sources for ninja and mingw. See my review comments.

.github/workflows/main.yml
Honeybunch: Most github actions for setting these things up are created by users like this. I've used this particular action for setting up ninja a number of times and it's pretty widely depended on. I see the concern but I'd argue that this is as stable as it can get for github actions

.github/workflows/main.yml
Honeybunch: This is the PR I put up to upstream to the setup-mingw action I was going to use: https://github.com/egor-tensin/setup-mingw/pull/3

User egor-tensin is the only person I've found who has setup an action for setting up mingw. If my PR gets accepted I can go back and update this to use egor-tensin's version rather than mine. Either way the action packages themselves are hosted by Github so they're not going anywhere. 

A better PR for this might have been to modify the script to enable the action user to specify a version. The only complication I saw for that was that this action can also setup mingw on linux or through cygwin so it's interacting with a few different package managers.

CMakeLists.txt
Honeybunch: :+1:

MarkCallow (approved):

MarkCallow:
@FuXiii is `F:\\Git\\bin` in your %PATH% environment variable? If so `find_program` should have found it.

To determine the version number, FindBash has a RE match on the version string output by bash which includes "version". Before executing `bash --version` it does `set(ENV{LANG} "en_US.UTF-8")` to ensure the version is printed in English. If `F:\\Git\\bin` is in your path, please try setting LANG to en_US.UTF-8 and then `bash version` to see if the output is in English.

MarkCallow:
@Honeybunch how is progress on upstreaming your change to the script for installing mingw? If it is close I'd like to wait for it and update this PR accordingly.

As for the ninja install, since it is coming from a location on GitHub I'm okay with using that fork if that is the best alternative.

I'm looking into the build problems on Travis. Of the 2 jobs that filed, one succeeded on retrying. I'm retrying the other as I write this. After that I will try the ones that errored. I can see no reason for the errors.

Honeybunch:
That change is just kind of sitting there. Worried that the owner may not be active. I'll try to chase it down after my vacation this week. Otherwise the options are to just rely on my fork or for KhronosGroup to fork my patched copy and rely on that. That way you don't have to rely on mine; but I want this change alive for my own uses so it should be alive for as long as github will host the fork

FuXiii:
@MarkCallow Yes you are right.
In my `systtem %PATH% environment variable` only the  `F:\Git\cmd` had in it, that make some kind weird, then I add the `F:\Git\bin` in `%PATH%`. the `bash --version` output:

image

Previously used `git-bash.exe` is in git path root, it will automatic load `./bin/bash.exe`. 

image

MarkCallow:
@FuXiii `git-bash.exe` starts an MSYS2 terminal running bash. It is not itself bash and `git-bash.exe --version` does not produce any output. There is no way to make FindBash use this to find the actual bash executable needed to run the script in the build. Therefore everything needed to support mingw builds in already in this PR. I will merge shortly.

FuXiii:
Ok, thanks :)

FuXiii:
When I build KTX with `mingw32-make`:
```
E:\Turbo\thirdparty\KTX-Software\lib\astc_encode.cpp:437:5: error: unknown type name 'pthread_t'
    pthread_t threadHandle;
    ^
1 error generated.
mingw32-make[2]: *** [thirdparty\KTX-Software\CMakeFiles\ktx.dir\build.make:500: thirdparty/KTX-Software/CMakeFiles/ktx.dir/lib/astc_encode.cpp.obj] Error 1
mingw32-make[1]: *** [CMakeFiles\Makefile2:3057: thirdparty/KTX-Software/CMakeFiles/ktx.dir/all] Error 2
mingw32-make: *** [Makefile:155: all] Error 2
The terminal process "C:\WINDOWS\System32\WindowsPowerShell\v1.0\powershell.exe -Command mingw32-make" terminated with exit code: 1.
```
the vscode clangd can find the define of `pthread_t`
image
but the compiler can't find it...
What should I do?
ε(┬┬﹏┬┬)3

MarkCallow:
When you create a new cmake configuration, i.e. when you select the generator and the compiler and they differ from any existing config in the output directory, cmake runs a test to see if pthread is available. This is the output I see when configuring for Xcode & clang:

```
Looking for pthread.h
Looking for pthread.h - found
Performing Test CMAKE_HAVE_LIBC_PTHREAD
Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
```

I see the same in our CI build under Mingw which specifies Ninja and gcc.

What compiler are you specifying in the configure step and what do you see in the configure output with regards to pthread?

FuXiii:
I use clang/clang++
```
C:\Users\g1018>clang++ --version
clang version 14.0.0 (https://github.com/llvm/llvm-project.git 329fda39c507e8740978d10458451dcdb21563be)
Target: x86_64-w64-windows-gnu
Thread model: posix
InstalledDir: F:/llvm-mingw-20220323-msvcrt-x86_64/bin
```
---
`CMake` output:
```
$ cmake . -B build -G "MinGW Makefiles" -D KTX_FEATURE_LOADTEST_APPS=ON -D KTX_FEATURE_DOC=OFF -D KTX_FEATURE_STATIC_LIBRARY=ON -D CMAKE_EXPORT_COMPILE_COMMANDS=1 -D CMAKE_CXX_COMPILER=clang++ -D CMAKE_C_COMPILER=clang
-- Found Bash: F:/Git/usr/bin/bash.exe
-- The C compiler identification is Clang 14.0.0
-- The CXX compiler identification is Clang 14.0.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: F:/llvm-mingw-20220323-msvcrt-x86_64/bin/clang.exe - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: F:/llvm-mingw-20220323-msvcrt-x86_64/bin/clang++.exe - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Looking for CL_VERSION_2_2
-- Looking for CL_VERSION_2_2 - not found
-- Looking for CL_VERSION_2_1
-- Looking for CL_VERSION_2_1 - not found
-- Looking for CL_VERSION_2_0
-- Looking for CL_VERSION_2_0 - not found
-- Looking for CL_VERSION_1_2
-- Looking for CL_VERSION_1_2 - not found
-- Looking for CL_VERSION_1_1
-- Looking for CL_VERSION_1_1 - not found
-- Looking for CL_VERSION_1_0
-- Looking for CL_VERSION_1_0 - not found
-- Could NOT find OpenCL (missing: OpenCL_LIBRARY OpenCL_INCLUDE_DIR)
-- Found Vulkan: F:/VulkanSDK/1.3.204.1/Lib/vulkan-1.lib
-- Found Perl: F:/Git/usr/bin/perl.exe (found version "5.26.2")
--   AVX2 backend     - ON
--   SSE4.1 backend   - OFF
--   SSE2 backend     - OFF
--   NEON backend     - OFF
--   NONE backend     - OFF
--   NATIVE backend   - OFF
--   Decompressor     - OFF
--   No invariance    - OFF
--   Diagnostics      - OFF
--   ASAN             - OFF
--   Unit tests       - OFF
-- Found OpenGL: opengl32
OPENGL_ES_EMULATOR not set. Will not build OpenGL ES load tests applications.
-- Found Vulkan at F:/VulkanSDK/1.3.204.1/Include & F:/VulkanSDK/1.3.204.1/Lib/vulkan-1.lib
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Failed
-- Looking for pthread_create in pthreads
-- Looking for pthread_create in pthreads - not found
-- Looking for pthread_create in pthread
-- Looking for pthread_create in pthread - found
-- Found Threads: TRUE
-- Configuring done
-- Generating done
-- Build files have been written to: E:/Lib/KTX-Software/build
```

MarkCallow:
@FuXiii, it is not finding the declaration in the screenshot you posted from astc_encode.cpp because there is a `!defined(__MINGW32__)` just before it. I am not familiar with MingW but it looks like that code should only be excluded when gcc is being used with Mingw.

On the other hand since CMake is finding thread support without these declarations, there must be some other way to get the thread support in Mingw & Clang. @Honeybunch do you know?

Honeybunch:
I realized after much prodding that I can't repro this because of how my copy of LLVM was compiled with the default triple & sysroot being MSVC rather than MinGW:

```
λ clang --version
clang version 14.0.3
Target: x86_64-pc-windows-msvc
Thread model: posix
InstalledDir: C:\Users\Arsine\scoop\apps\llvm\current\bin
```

I just configured with ` cmake -B build-clang-mingw -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -G "MinGW Makefiles" -DCMAKE_MAKE_PROGRAM=make -DKTX_FEATURE_LOADTEST_APPS=ON -DKTX_FEATURE_DOC=OFF -DKTX_FEATURE_STATIC_LIBRARY=ON`

and build with `cmake --build build-clang-mingw` and it just worked fine (after tweaking an .rc file to be UTF8 rather than UTF16)

I tried getting my copy of LLVM to build with my mingw sysroot but I think scoop's packaging of mingw is not suitable for this as this test: `clang++ --target=x86_64-w64-windows-gnu --sysroot=%USERPROFILE%/scoop/apps/gcc/11.2.0 -o test test.cpp -v`

Fails with: `ld.exe: cannot find -lgcc_s`

I'll have to spin up a VM real quick to get a clean environment where I can repro this

Honeybunch:
I have a mingw environment set up in msys2 now and I'm actually still having a hard time reproducing this. There are still some things I can fix here; it doesn't build. But I'm not getting hung up on the same issue. My version of clang and gcc both compile with `__MINGW32__` defined.

```
$ clang --version
clang version 14.0.3
Target: x86_64-w64-windows-gnu
Thread model: posix
InstalledDir: C:/msys64/mingw64/bin
```

```
$ gcc --version
gcc.exe (Rev1, Built by MSYS2 project) 12.1.0
Copyright (C) 2022 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

The packages here I installed from msys2 are `mingw-w64-x86_64-gcc` and mingw-w64-x86_64-clang`

Honeybunch:
I got something building but still didn't run into a problem with threads. I configured with: `cmake . -B build -G "MinGW Makefiles" -DCMAKE_BUILD_TYPE=Debug -D KTX_FEATURE_LOADTEST_APPS=OFF -D KTX_FEATURE_DOC=OFF -D KTX_FEATURE_STATIC_LIBRARY=ON -D CMAKE_EXPORT_COMPILE_COMMANDS=1 -D CMAKE_CXX_COMPILER=clang++ -D CMAKE_C_COMPILER=clang`

I had to turn off the FEATURE_LOADTEST_APPS because some binary support libs (like SDL2) are built against the MSVC ABI and will not link with the mingw version. I didn't run into this because that define is off by default. Maybe a task for the future.

Note that in my msys2 environment that the built-in cmake from pacman didn't support the "Mingw Makefiles" generator so I had to install cmake from the web and add it to my msys2 bash path. I also had to use `mingw-w64-x86_64-make`. Kind of a weird setup. 

I'm going to put up a PR to use the built in `MINGW` var that CMake provides which, in my experience, will detect MinGW usage properly in and outside of MSYS2 with either GCC or Clang but will properly return false when using clang with MSVC.

Honeybunch:
I did manage to come up with something that resembles FuXiii's issue. 

Building from a standard windows powershell prompt with:
```
cmake -B build-clang-mingw -G "Ninja Multi-Config" -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_FLAGS="--target=x86_64-w64-windows-gnu --sysroot=C:/ProgramData/chocolatey/lib/mingw/tools/install/mingw64" -DCMAKE_CXX_FLAGS="--target=x86_64-w64-windows-gnu --sysroot=C:/ProgramData/chocolatey/lib/mingw/tools/install/mingw64"
```
Here I'm using the msvc based LLVM from chocolately with the sysroot provided by chocolately. Unlike the version of gcc provided by scoop, this one from chocolatey is the same version but seems to include some of the libraries scoop was missing (gcc_s). 

```
PS C:\Users\User\KTX-Software> clang --version
clang version 14.0.3
Target: x86_64-pc-windows-msvc
Thread model: posix
InstalledDir: C:\Program Files\LLVM\bin
```

This isn't the same issue that was seen before. I get linker errors like `C:/Users/User/KTX-Software/lib/astc_encode.cpp:486: undefined reference to `pthread_join'`

If I'm reading FuXiii's post correctly I see `InstalledDir: F:/llvm-mingw-20220323-msvcrt-x86_64/bin` and that `msvcrt` gives me pause. 

@FuXiii could you do me a favor: 
Create a test file with `touch test.c` (the file can be blank)
Run `clang -dM -E test.c > defines.h`
This will report all of the defines that the compiler provides to the preprocessor by default. Do you see `__MINGW32__` in there?

If I force my MSVC built clang to use the gnu triple with `clang --target=x86_64-w64-windows-gnu --sysroot=C:/ProgramData/chocolatey/lib/mingw/tools/install/mingw64 -dM -E test.c > defines.h` the defines still don't report `__MINGW32__`

This is a strange config but means that `__MINGW32__` is a pretty unreliable define when caring about Clang on MinGW.

I'll have to figure out a good way to get CMake to report pthreads support

Honeybunch:
Put up an update to https://github.com/KhronosGroup/KTX-Software/pull/579/ which should result in much better detection of pthreads support on Windows

I just built for windows with:
LLVM + Mingw in MSYS2
LLVM + Mingw with that fucked-up sysroot nonsense in powershell
GCC + Mingw in powershell
CL + MSVC in VS2022 dev env

All of them worked

FuXiii:
```CMake
        if(MINGW)
            # Check if the Threads package is provided; if using Mingw it MIGHT be
            find_package(Threads)
            if(Threads_FOUND)
                target_compile_definitions(ktx PRIVATE WIN32_HAS_PTHREADS)
                target_link_libraries(ktx PRIVATE Threads::Threads)
            endif()
        endif()
```
the `find_package(Threads)`, Should I download the `Threads` lib from some where?

PR #584, opened 2022-06-03
Release memory before early exit.

kacprzak:
The `prototype` texture is released only before last `return` statement. Probably should be released before all returns.


MarkCallow:
Thank you for this.

PR #594, opened 2022-06-25
Regularize Tools

MarkCallow:
This does many things:

- Makes the stragglers, _ktx2ktx2_ and _ktxinfo_, use `ktxApp` base class
- Makes all tools use a common `stdafx.h` replacing a copy per app.
- Adds `fopen_write_if_not_exists` function to `ktxApp` which fixes the failure to close an existing file after the check that was in the _ktxsc_-specific code it replaces.
- Makes _ktx2ktx2_ use `fopen_write_if_not_exists` so it too will work on mingw or with very old versions of msvcrt.
- Adds _ktx2ktx2_ tests.
- Fixes documentation errors and formatting.
- Fixes a warning in `vkloader.cpp` raised by older Visual Studio versions.
- Fixes warning about an unknown warning disable from older versions of Xcode.
- Fixes _ktx2ktx2_ and _ktxsc_ to check input file is the correct ktx version and remove output file on any error after its creation.
- Fixes _ktx2check_, _ktx2ktx2_ and _ktxsc_ tests to actually fail when the tools print incorrect error strings.


MarkCallow:
@Honeybunch I need your assistance again. Very sorry. In this PR I added tests of the `ktx2ktx2` tool. The ones that involve reading a file are failing on Mingw. An error is occurring on an`fread` of the input file. It is returning less than the number of items requested and the error indicator is being set. I put some debugging into the internal filestream read function to get this far but to progress really needs being able to run the tool in a debugger on Mingw. Please take a look.

See [here](https://github.com/KhronosGroup/KTX-Software/runs/7089564649?check_suite_focus=true) for a failing build.

Here's a relevant section from the `ctest` test log:
```
166/320 Testing: ktx2ktx2-cnvrt-2d-uncompressed
166/320 Test: ktx2ktx2-cnvrt-2d-uncompressed
Command: "C:/Program Files/Git/bin/bash.exe" "-c" "D:/a/KTX-Software/KTX-Software/build/Release/ktx2ktx2.exe --test -f -o ktx2ktx2.orient-down-metadata-u.ktx2 orient-down-metadata.ktx && diff orient-down-metadata-u.ktx2 ktx2ktx2.orient-down-metadata-u.ktx2 && rm ktx2ktx2.orient-down-metadata-u.ktx2"
Directory: D:/a/KTX-Software/KTX-Software/tests/testimages
"ktx2ktx2-cnvrt-2d-uncompressed" start time: Jun 28 09:26 Coordinated Universal Time
Output:
----------------------------------------------------------
Error reading filestream: Invalid argument          <<<<<<<=====================================
ktx2ktx2 failed to create ktxTexture from orient-down-metadata.ktx: File read error.
<end of output>
Test time =   0.06 sec
----------------------------------------------------------
Test Failed.
"ktx2ktx2-cnvrt-2d-uncompressed" end time: Jun 28 09:26 Coordinated Universal Time
"ktx2ktx2-cnvrt-2d-uncompressed" time elapsed: 00:00:00
----------------------------------------------------------
```

The highlighted line is from the debugging I added. It confirms the "File read error" is due to the error indicator rather than the eof indicator being set on the stdio stream. Ignore the `Invalid argument`.  It is a printout of the `errno` value but `fread` does not set `errno` so the value shown is not relevant.

Another tool `ktxsc` has virtually identical code for opening its input files and all its tests are working on Mingw so it is very difficult to understand what is happening here. The only difference I can see is that `ktx2ktx2` is opening KTX v1 files while `ktxwc` is opening v2 files but that wouldn't make any difference to the file reading code.

Honeybunch:
Woah yeah that's weird. I'll try to get some time to look at that soon. Mingw does use a different C++ ABI IIRC but for cdecl linkage I'm pretty sure it just links against the MSVC CRT. I need to verify that though.

Honeybunch:
Yeah every source distribution of mingw that I can find just distributes .def files for assembling what I assume to be a reference DLL that winds up pointing to the standard msvcrt. I can't seem to find anything that suggests a full custom fread implementation anywhere so I'm really confused. Something seems afoot since clearly we're not getting the same behavior.

MarkCallow:
> just distributes .def files for assembling what I assume to be a reference DLL that winds up pointing to the standard msvcrt.

.def files are used to list the functions to be exported from a DLL that you are building. If there are other uses, I am unaware of them.

> Something seems afoot since clearly we're not getting the same behavior.

Yes.

Honeybunch:
So doing some more digging into the way that mingw handles the C std lib. It seems to link against msvcrt/ucrt rather than provide a full glibc: https://stackoverflow.com/questions/6394512/standard-c-library-in-mingw

I think I'm correct that the dll that's built my mingw mostly just exports linkages that point to the actual microsoft crt. Similarly to how you might see something like `#pragma comment(linker, "/EXPORT:Direct3DCreate9=C:\\Windows\\System32\\d3d9.Direct3DCreate9")` in a project that wants to hook D3D9's DLLmain but still provide an export to the original functionality.

I've been pouring over the Mingw-w64 code hosted here: https://sourceforge.net/p/mingw-w64/mingw-w64/ci/master/tree/ and haven't been able to find any evidence of a custom `fread` or `fwrite` implementation. I'm *very* confused.

I wrote up a small test program to compile with mingw just to ensure that I'm not losing my mind: 
```
#include <stdio.h>
#include <stdint.h>
#include <stdlib.h>

int main()
{
    FILE *file = fopen("pattern_02_bc2.ktx", "rb");

    fseek(file, 0, SEEK_END);
    size_t len = (size_t)ftell(file);
    fseek(file, 0, SEEK_SET);

    uint8_t *buffer = malloc(len);

    size_t read = fread(buffer, 1, len, file);
    int32_t err = ferror(file);

    fclose(file);
    free(buffer);

    return 0;
}
```
This *does* open the same file and reads the entire contents without issue. I was wondering if it was some UNICODE nonsense but it doesn't look like any of the KTX code is compiled with UNICODE even though there are some places where functions from tchar.h are used.

Honeybunch:
Other thing to note: ktx2ktx2 is failing after reading 4092 of 11202397 bytes. That's consistent between runs and between Debug and Release.

Honeybunch:
Did some debug logging and here's what it looks like when a run succeeds (built with llvm for windows) 
image

But here's what a broken run with mingw looks like:
image

From this: `fprintf(stdout, "(%x) reading %d from readpos: %d\n", str->data.file, count, str->readpos);` in `ktxFileStream_read`

MarkCallow:
@Honeybunch, thanks for the investigative work. It looks like you can reproduce the problem locally which is good.

I did a side-by-side diff of `tools/ktx2ktx2/ktx2ktx2.cpp` and `tools/ktxsc/ktxsc.cpp` because the latter works and the code handling setting up the input and output files and creating a texture from the input file is virtually identical. In fact they are identical except for 2 things:

1. `ktxsc.cpp` includes windows.h before any other include files.
2. The inputs to`ktx2ktx` tests are KTX v1 files while those for the ktxsc tests are KTX v2 files.

I suspect no 1. In the process of doing the diff I found several things in the code I wanted to clean up. The most important is that both tools were using the general purpose `ktxTexture_CreateFromStdioStream` and not checking if the created texture was of the relevant type. I've fixed them to use the `ktxTexture1_` and `ktxTexture2_` variants as appropriate and I've added tests with the incorrect input file type. I will commit these changes and see what happens. If, as I expect, the problem still occurs I will try adding `#include <windows.h>` at the top of `ktx2ktx2.cpp` and see what happens.

MarkCallow:
I included a printf when the error condition triggered, similar to yours, in my latest commit and tried a CI run. The only consistency is that it always fails after reading 4092 bytes but the reported size of the request (nitems) is very strange in 2 cases. Here is the test log:

```
168/324 Testing: ktx2ktx2-cnvrt-2d-uncompressed
168/324 Test: ktx2ktx2-cnvrt-2d-uncompressed
Command: "C:/Program Files/Git/bin/bash.exe" "-c" "D:/a/KTX-Software/KTX-Software/build/Release/ktx2ktx2.exe --test -f -o ktx2ktx2.orient-down-metadata-u.ktx2 orient-down-metadata.ktx && diff orient-down-metadata-u.ktx2 ktx2ktx2.orient-down-metadata-u.ktx2 && rm ktx2ktx2.orient-down-metadata-u.ktx2"
Directory: D:/a/KTX-Software/KTX-Software/tests/testimages
"ktx2ktx2-cnvrt-2d-uncompressed" start time: Jun 30 06:38 Coordinated Universal Time
Output:
----------------------------------------------------------
Error reading filestream. Attempted to read -1 bytes, only 4092 read.
ktx2ktx2 failed to create ktxTexture from orient-down-metadata.ktx: File read error.
<end of output>
Test time =   0.05 sec
----------------------------------------------------------
Test Failed.
"ktx2ktx2-cnvrt-2d-uncompressed" end time: Jun 30 06:38 Coordinated Universal Time
"ktx2ktx2-cnvrt-2d-uncompressed" time elapsed: 00:00:00
----------------------------------------------------------

169/324 Testing: ktx2ktx2-cnvrt-2d-bc2
169/324 Test: ktx2ktx2-cnvrt-2d-bc2
Command: "C:/Program Files/Git/bin/bash.exe" "-c" "D:/a/KTX-Software/KTX-Software/build/Release/ktx2ktx2.exe --test -f -o ktx2ktx2.pattern_02_bc2.ktx2 pattern_02_bc2.ktx && diff pattern_02_bc2.ktx2 ktx2ktx2.pattern_02_bc2.ktx2 && rm ktx2ktx2.pattern_02_bc2.ktx2"
Directory: D:/a/KTX-Software/KTX-Software/tests/testimages
"ktx2ktx2-cnvrt-2d-bc2" start time: Jun 30 06:38 Coordinated Universal Time
Output:
----------------------------------------------------------
Error reading filestream. Attempted to read 11202397 bytes, only 4092 read.
ktx2ktx2 failed to create ktxTexture from pattern_02_bc2.ktx: File read error.
<end of output>
Test time =   0.05 sec
----------------------------------------------------------
Test Failed.
"ktx2ktx2-cnvrt-2d-bc2" end time: Jun 30 06:38 Coordinated Universal Time
"ktx2ktx2-cnvrt-2d-bc2" time elapsed: 00:00:00
----------------------------------------------------------

170/324 Testing: ktx2ktx2-cnvrt-2d-array-astc
170/324 Test: ktx2ktx2-cnvrt-2d-array-astc
Command: "C:/Program Files/Git/bin/bash.exe" "-c" "D:/a/KTX-Software/KTX-Software/build/Release/ktx2ktx2.exe --test -f -o ktx2ktx2.texturearray_astc_8x8_unorm.ktx2 texturearray_astc_8x8_unorm.ktx && diff texturearray_astc_8x8_unorm.ktx2 ktx2ktx2.texturearray_astc_8x8_unorm.ktx2 && rm ktx2ktx2.texturearray_astc_8x8_unorm.ktx2"
Directory: D:/a/KTX-Software/KTX-Software/tests/testimages
"ktx2ktx2-cnvrt-2d-array-astc" start time: Jun 30 06:38 Coordinated Universal Time
Output:
----------------------------------------------------------
Error reading filestream. Attempted to read -1839259522 bytes, only 4092 read.
ktx2ktx2 failed to create ktxTexture from texturearray_astc_8x8_unorm.ktx: File read error.
<end of output>
Test time =   0.05 sec
----------------------------------------------------------
Test Failed.
"ktx2ktx2-cnvrt-2d-array-astc" end time: Jun 30 06:38 Coordinated Universal Time
"ktx2ktx2-cnvrt-2d-array-astc" time elapsed: 00:00:00
----------------------------------------------------------

171/324 Testing: ktx2ktx2-cnvrt-implied-out-2d-bc2
171/324 Test: ktx2ktx2-cnvrt-implied-out-2d-bc2
Command: "C:/Program Files/Git/bin/bash.exe" "-c" "cp pattern_02_bc2.ktx ktx2ktx2.ip.pattern_02_bc2.ktx && D:/a/KTX-Software/KTX-Software/build/Release/ktx2ktx2.exe --test -f ktx2ktx2.ip.pattern_02_bc2.ktx && rm ktx2ktx2.ip.pattern_02_bc2.ktx && diff pattern_02_bc2.ktx2 ktx2ktx2.ip.pattern_02_bc2.ktx2 && rm ktx2ktx2.ip.pattern_02_bc2.ktx2"
Directory: D:/a/KTX-Software/KTX-Software/tests/testimages
"ktx2ktx2-cnvrt-implied-out-2d-bc2" start time: Jun 30 06:38 Coordinated Universal Time
Output:
----------------------------------------------------------
Error reading filestream. Attempted to read 11202397 bytes, only 4092 read.
ktx2ktx2 failed to create ktxTexture from ktx2ktx2.ip.pattern_02_bc2.ktx: File read error.
<end of output>
Test time =   0.11 sec
----------------------------------------------------------
Test Failed.
"ktx2ktx2-cnvrt-implied-out-2d-bc2" end time: Jun 30 06:38 Coordinated Universal Time
"ktx2ktx2-cnvrt-implied-out-2d-bc2" time elapsed: 00:00:00
----------------------------------------------------------
```

I'm going to try including `<windows.h>` next.

MarkCallow:
> I'm going to try including `<windows.h>` next.

It still fails. Will have to look more closely at KTX v1 loading.

MarkCallow:
@Honeybunch, regarding the results you showed earlier from your tests, the reads are as follows:

1. The file identifier (12 bytes)
2. The rest of the header (52 bytes)
3. The metadata (32 bytes)
4. A uint32_t giving size of the first mip level (4 bytes)
5. The first mip level using the size read in step 4 as the `count` argument to `filestream_read`. The actual size in the subject file is 1048576 which you see in the working llvm version. It looks like something is going wrong in reading the size in step 4 since it is attempting to read 11202397.

Try putting some prints of the size at lines 1308 and 1328 in `lib/texture1.c`. However this doesn't explain the error. If the size was simply too large I'd expect `feof` to return true not `ferror`.

MarkCallow:
I found the problem. Only the KTX v1 case calls `filestream_getsize`.

https://github.com/KhronosGroup/KTX-Software/blob/efe14a06ba01d3c4158f644756431d792826a810/lib/filestream.c#L323-L331

What is the appropriate define to add to this for the Mingw case?

Honeybunch:
Good catch! You can detect mingw with `__MINGW32__`: `(defined(_MSC_VER) && _MSC_VER < 1900) || defined(__MINGW32__)` worked locally for me

image
image

MarkCallow:
> Good catch! You can detect mingw with `__MINGW32__`: `(defined(_MSC_VER) && _MSC_VER < 1900) || defined(__MINGW32__)` worked locally for me

Thanks. I'll try that. IT seems Mingw is using a very old version of msvcrt as the bug being worked around was in VS2013.

PR #606, opened 2022-07-28
Build-system fixes

pierremoreau:
* Fix missing header when building toktx while `KTX_FEATURE_VULKAN` is disabled.
* Do not build glloadtests and vkloadtests when their corresponding loader is not enabled, otherwise it results in missing symbols. For example, do not build vkloadtests when `KTX_FEATURE_LOADTEST_APPS` is enabled but `KTX_FEATURE_VULKAN` is not.

### DIFF
--- a/tests/loadtests/CMakeLists.txt
+++ b/tests/loadtests/CMakeLists.txt
@@ -196,9 +196,11 @@ set( LOAD_TEST_COMMON_RESOURCE_FILES
     ${MODELS}
 )
 
-include(glloadtests.cmake)
+if(KTX_FEATURE_GL_UPLOAD)
+    include(glloadtests.cmake)
+endif()
 
-if(NOT EMSCRIPTEN)
+if(KTX_FEATURE_VULKAN AND NOT EMSCRIPTEN)
     include(vkloadtests.cmake)
 endif()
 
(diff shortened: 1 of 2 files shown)


MarkCallow:
I see `pngimage.cc` includes dfd.h. How come it managed to be  found when KTX_FEATURE_VULKAN was enabled?

pierremoreau:
> I see `pngimage.cc` includes dfd.h. How come it managed to be found when KTX_FEATURE_VULKAN was enabled?

That is due to:

1. The include directories for libktx being modified to include the following when `KTX_FEATURE_VULKAN` is enabled:
    ```CMake
        target_include_directories(
            ${lib}
        PRIVATE
            $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/lib/dfdutils>
            $<INSTALL_INTERFACE:lib/dfdutils>
        )
    ```
2. And toktx adding the following to its own include directories:
    ```CMake
    $<TARGET_PROPERTY:ktx,INCLUDE_DIRECTORIES>
    ```

MarkCallow:
Thank you for the explanation and for this fix.

pierremoreau:
No worries! Thank you for the quick replies and work on the project.

PR #621, opened 2022-08-14
Fix JNI package names in KtxTexture2.cpp

Illithidek:
The name of the functions in the JNI KtxTexture2.cpp didn't match
the java package which cause linker error.

### DIFF
--- a/interface/java_binding/src/main/cpp/KtxTexture2.cpp
+++ b/interface/java_binding/src/main/cpp/KtxTexture2.cpp
@@ -7,17 +7,17 @@
 #include <iostream>
 #include "libktx-jni.h"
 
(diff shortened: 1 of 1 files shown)


MarkCallow:
How come our tests were able to link and run?

Illithidek:
Hey, tests just don't use these functions so JNI doesn't call them so the linker doesn't have a problem :)

Illithidek:
Basically, JNI works on the promise that somewhere out there we will add a library that will have these functions. Like an "extern" in C. If something is not invoked then the promise doesn't have to be fulfilled.
All other functions works quite well, I found these by luck because I needed to use "needsTranscoding" and a linker error appeared.

MarkCallow:
Thanks for the contribution.

PR #622, opened 2022-08-15
New function: ktxTexture_WriteToMemory2

AnonN10:
Implementation of #614. Includes new types of memory streams, needs a thorough review.


AnonN10:
I have no idea why AppVeyor build failed. I don't think I can fix this one.

MarkCallow:
> I have no idea why AppVeyor build failed. I don't think I can fix this one.

Don't worry about it. A [different build of the same commits](https://ci.appveyor.com/project/msc-/ktx/builds/44469737), also on Appveyor, worked. When I look at the failed build log, it does not give me the opportunity to `rerun failed jobs` so there is nothing I can do. I also don't understand why Appveyor does 2 builds for every PR only one of which shows up in the project build history. I'm not minded to investigate because I'm working on moving the build to GitHub Actions.

I noticed some compile warnings when looking through the log. Please fix them.

MarkCallow:
I've not had time for a deep review yet. On a brief look some of the new functions in memstream.c could likely be combined, possibly using flags to control behavior. Any internal function can be freely changed, e.g. those declared `static`.

Also we'll need tests for this new functionality.  They can be added to `tests/texturetests/texturetests.cc` (the WriteToStream stuff) and to `tests/unittests/unittests.cc` (tests of the memstream stuff).

AnonN10:
Can't add tests, but will try to fix the warnings

MarkCallow:
I can't merge without tests for the new functionality. "If it isn't tested it's broken." I an confident that these functions are working for you now. Having tests is about keeping them working in future.

There are some memstream tests in `tests/unittests/unittests.cc`. You can probably clone and modify one or more of those to test your additions. There are no explicit tests for `ktxtexture?_WriteToMemory` because `tests/texturetes/texturetest.cc` relies heavily on those functions but it won't be much work to take some pieces from some of the other tests to make tests for the `ktxTexture?_WriteToMemory2` functions.

lib/memstream.c
```diff
@@ -258,6 +349,26 @@ KTX_error_code ktxMemStream_read(ktxStream* str, void* dst, const ktx_size_t cou
...
+*                          specifies an input stream.
+* @param [in,out] dst      pointer to memory where to copy read bytes.
+* @param [in,out] count    pointer to number of bytes to read.
+*
+* @return      KTX_INVALID_OPERATION.
+*/
+static
+KTX_error_code ktxMemStream_disable_read(ktxStream* str, void* dst, const ktx_size_t count)
+{
+    UNUSED(str);
```
MarkCallow, line 366: What is the point of this function? It does nothing.

A simpler fix  for unused parameters is to remove the parameter name leaving only the type. `UNUSED` is intended for situations such as when there is a variable only used for an assert so unused in a Release build.

AnonN10: that's quite literally your idea
https://github.com/KhronosGroup/KTX-Software/discussions/614#discussioncomment-3354432
> The read function for such a stream should return KTX_INVALID_OPERATION.

AnonN10: > A simpler fix for unused parameters is to remove the parameter name leaving only the type. `UNUSED` is intended for situations such as when there is a variable only used for an assert so unused in a Release build.

Okay I rushed things a bit and edited files without trying them out first. Turns out the alternative you suggest is invalid in C and fails to build with MSVC (error code C2055). I'll be reverting the changes.

AnonN10:
I added the tests for the new memory stream types. As for `WriteToMemory2`, I don't know how and where exactly to add tests for it because `texturetests.cc` seems to be using a different approach to unit testing which I'm not familiar with. I'd like you to elaborate a bit on what you mean by "no explicit tests".

lib/memstream.c
MarkCallow: > that's quite literally your idea

I was confused by the function name which implies that it should be doing something, such as setting a flag, to disable read. Maybe `ktxMemStream_no_read. I hope to do a full review of this early next week.

Sorry for the bad steer on the parameter names.

MarkCallow:
> I'd like you to elaborate a bit on what you mean by "no explicit tests".

Both `unittests.cc` and `texturetests.cc` are gtest based. `unittests` is for testing various functions not exposed in the API. `texturetests` is for testing the ktxTexture{,[12]}_* APIs. What I mean by no explicit test is that there is no, e.g.

```
TEST_F(ktxTexture1WriteToMemoryTest , ...)
```

This is because all of the 

```
TEST_F(ktxTexture?WriteTest*, ...)
```

use `ktxTexture1_WriteToMemory` underneath via the `helper`, which you can find in `tests/unittests/wthelper.h`, so the functions are thoroughly tested.

You'll have to write new test fixture template similar to  `ktxTexture1WriteTestBase` as a base for the new tests. ktxTexture1WriteTestBase` is the base class for all the `TEST_F(ktxTexture?WriteTest*, ...)` tests.

AnonN10:
I'm really struggling to understand where to put the test, I did try to look for instances of `WriteToMemory` in `texturetests.cc`, but all of them seem to not be testing the call to the function itself.

MarkCallow:
> I'm really struggling to understand where to put the test, I did try to look for instances of `WriteToMemory` in `texturetests.cc`, 

That is what I meant by no explicit test for WriteToMemory.

> but all of them seem to not be testing the call to the function itself.

All of the ktxTexture1WriteTest* tests uses WriteToMemory to write the files that are later compared to an expected result. Thus they all indirectly verify that WriteToMemory is working. `ktxTextureTestBase` which is the base for the other tests also uses WriteToMemory at the end of its constructor to create the file later tested for correctness.

For WriteToMemory2 I would create a test fixture template that accepts numComponents, etc like the other fixtures and creates a file into allocated memory. Tests can use that to create the file then open it file and compares it with the expected result using a compare function in the test fixture. This can likely be derived from `ktxTextureTestBase`. Only the constructor needs to be different.

AnonN10:
Why does the build fail? What did I miss?

AnonN10:
I give up, wasted too much time on this, I'll just use a custom branch with my additions.

MarkCallow:
I regret that you are giving up. I think this will be a useful addition.

After the most recent commit this PR builds fine.

The Windows builds on Appveyor and the macOS, Linux & Emscripten builds on Travis-CI all worked.  The most recent GitHub Actions Mingw and Windows Arm64 builds failed but they look to be before the commit updating `internalexport.def` probably because I hadn't yet authorized a build of that commit*. The mingw build failed due to the linker not finding the functions you added to `internalexport.def`. The Windows Arm64 build failed due to an ongoing issue with Chocolatey when trying to install Doxygen that can only be resolved by re-running the failed build. It would eventually have failed for the same reason as the Mingw build had the Doxygen install worked.

* Until a first PR has been accepted from a contributor Actions requires me to authorize each build of the initial PR.

AnonN10:
Sorry really can't afford to spend my time on tests, they are too convoluted for my small brain. And I lost 2 weeks without making much progress on my own matters as a result already! But if you really want, I may eventually create another PR with yet another addition this time, which is a chunk based `CreateFromMemory`, however I won't be touching the tests.

PR #623, opened 2022-08-17
Cherry-pick stability fix from astcenc 4.1.0 release

solidpixel:
This PR is a cherry-pick of a fix for the astcenc compressor backend that we made in the astcenc 4.1.0 release.

Without the fix SSE2 and SSE4.1 builds that use the decompressor APIs can generate faults on masked stores, as `_mm_maskmoveu_si128()` doesn't suppress faults on any lanes that are disabled. AVX2 builds are not impacted by this - the AVX2 masked store does suppress faults on masked lanes.


MarkCallow:
Thanks @solidpixel.

PR #634, opened 2022-09-27
fix(cmake): Fallback to astcenc-static for fat libraries (multiple ar…

atteneder:
…chitectures). An example would be `CMAKE_OSX_ARCHITECTURES="x86_64;arm64"`.

The ideal solution would be to use different astcenc versions per architecture. This is a mere workaround avoid CMake configuration from failing.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -788,8 +788,9 @@ add_subdirectory(interface/basisu_c_binding)
 # On arm based systems ISA_NEON is default
 
 list(FIND CMAKE_OSX_ARCHITECTURES "$(ARCHS_STANDARD)" ASTC_BUILD_UNIVERSAL)
+list(LENGTH CMAKE_OSX_ARCHITECTURES ARCHITECTURE_COUNT)
 
-if(${ASTC_BUILD_UNIVERSAL} EQUAL -1)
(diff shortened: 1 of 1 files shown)


MarkCallow:
The proper fix is for the encoders (both astcenc and basisu) to use runtime detection or compiler-defined pre-processor macros. For the former macOS has various queries to determine supported CPU features. The latter makes a version per architecture during the compilation.

Because of these issues our releases do not use fat libraries. We build architecture-specific installers. We do not want to lose the SIMD performance advantage.

atteneder:
@MarkCallow Thanks for the explanation

> The proper fix is for the encoders (both astcenc and basisu) to use runtime detection or compiler-defined pre-processor macros. For the former macOS has various queries to determine supported CPU features. The latter makes a version per architecture during the compilation.

That makes sense.

> Because of these issues our releases do not use fat libraries. We build architecture-specific installers. We do not want to lose the SIMD performance advantage.

Ok. I understand that this makes sense in the context of the installers.

Feel free to reject this proposed fix. Assuming your ideal solution won't get done soon, we need a way the configuration does not fail for projects integrating libktx via CMake (like the Unity integration).

Alternatives that come to mind

- Option to force `astcenc-static` (opt-in)
- Option to disable ASTC enc altogether

What do you say?

MarkCallow:
> Assuming your ideal solution won't get done soon

It won't. This fix is is a fine interim step.

Please rebase with the latest master which has a fix for the timex.h build error in Emscripten.

PR #635, opened 2022-09-29
Fix static build on Linux

06393993:
On Linux, it is very likely that a GNU libtool is used, and different parameters are needed for libtool to combine libraries.


MarkCallow:
Thanks for this fix.

Issue #641, opened 2022-10-20
MingW build broke with latest MingW


MarkCallow:
The MingW build has been broken by an update to the MingW installation on GitHub Actions.

The issue is at https://github.com/KhronosGroup/KTX-Software/blob/f4feff2a56f652a705209be2651712e810532997/lib/filestream.c#L310-L317

The GCC version in the MingW where this works is 11.2. That in the version where it is broken is 12.2. I do not know the MingW version as cmake config does not report it. Presumably the later version is no longer using the msvcrt that had the problem. The obvious fix is to remove the `|| defined(__MINGW32__)` but that will break running, but not building, in earlier versions of MingW.

@Honeybunch please provide a PR to fix this. I do not have a MingW installation so can't test a fix especially not on multiple versions. Presumably a fix that will not break earlier versions will involve using the `__MINGWXX_MAJOR_VERSION__` and ``__MINGWXX_MAJOR_VERSION__` defines.

Why are we using `__MINGW32__` in the above and not `__MINGW64__` We should be building 64-bit libraries and applications so the latter seems more appropriate.

Honeybunch:
Oh no! I'll take a look soon

MarkCallow:
Great. I am ready to make Release 4.1.0 and this cropped up as I was updating the code-signing part of the build.

I'm also getting warnings about node.js 12 actions being deprecated. This affects the seanmiddleditch ninja action. Probably the reference needs updating to a newer version. [This build](https://github.com/KhronosGroup/KTX-Software/actions/runs/3280945044) shows the issues.

Honeybunch:
I have a repro and a local build fix but it seems like there are some serious test failures. I'll need to attach a debugger and hunt those down. This seems to happen regardless of whether or not I try to specify linking the ucrt or msvcrt. Though I may be messing that bit up. 

```
[ctest] 69% tests passed, 101 tests failed out of 331
[ctest] 
[ctest] Total Test time (real) =   4.40 sec
[ctest] 
[ctest] The following tests FAILED:
[ctest] 	 59 - texturetestktxTexture2_IterateLevelFacesTest.InvalidValueOnNullCallback (Exit code 0xc0000409
[ctest] )
[ctest] 	 60 - texturetestktxTexture2_IterateLevelFacesTest.IterateImages (Exit code 0xc0000409
[ctest] )
[...]

PR #642, opened 2022-10-21
Fixing support for mingw toolchains that target the newer ucrt

Honeybunch:
Attempting to fix: https://github.com/KhronosGroup/KTX-Software/issues/641

### DIFF
--- a/lib/filestream.c
+++ b/lib/filestream.c
@@ -307,13 +307,13 @@ KTX_error_code ktxFileStream_getsize(ktxStream* str, ktx_size_t* size)
 
     assert(str->type == eStreamTypeFile);
 
-    // Need to flush so that fstat will return the current size.
-    // Can ignore return value. The only error that can happen is to tell you
-    // it was a NOP because the file is read only.
-#if (defined(_MSC_VER) && _MSC_VER < 1900) || defined(__MINGW32__)
-    // Bug in VS2013 msvcrt. fflush on FILE open for READ changes file offset
-    // to 4096.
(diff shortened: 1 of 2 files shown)


lib/filestream.c
```diff
@@ -307,13 +307,13 @@ KTX_error_code ktxFileStream_getsize(ktxStream* str, ktx_size_t* size)
...
-    // Can ignore return value. The only error that can happen is to tell you
-    // it was a NOP because the file is read only.
-#if (defined(_MSC_VER) && _MSC_VER < 1900) || defined(__MINGW32__)
-    // Bug in VS2013 msvcrt. fflush on FILE open for READ changes file offset
-    // to 4096.
-    if (str->data.file->_flag & _IOWRT)
+  // Need to flush so that fstat will return the current size.
+  // Can ignore return value. The only error that can happen is to tell you
+  // it was a NOP because the file is read only.
+#if (defined(_MSC_VER) && _MSC_VER < 1900) || defined(__MINGW32__) && !defined(_UCRT)
```
MarkCallow, line 313: Should we change `__MINGW32__` to `__MINGW64__`?

Honeybunch: Oh yeah I forgot to reply to that part of the issue. I don't think we really need to. It'll work the same either way but I can change it if you think it's a good idea.

MarkCallow: Please change it to `__MINGW64__`. Since KTX-Software only supports 64-bit builds, it will be less confusing.

MarkCallow: That was quick! Thanks.

PR #654, opened 2022-12-05
git fallback to commit hash version ID

danhedron:
Background:
This is an issue when `KTX-Software` is included as a git submodule in a Rust crate (a Rust project), and that crate is being built as a dependency specified as a git URL in another crate. The implementation within the Cargo package manager results in a very sparse checkout (i.e. without any tags), which cannot be associated with a version number via git.

The result is a completely empty version number macro. This fails silently on GCC & clang, but does not compile under MSVC.

Change:
The git path checks one additional time for an empty version name, before combining the default version with the current short hash to identify the code. This ensures the `VERSION` macro is not left empty if the tag information is missing.

### DIFF
--- a/mkversion
+++ b/mkversion
@@ -39,6 +39,10 @@ function genversion() {
         # No containing tag. Get version number a previous tag.
         VN=$(git describe --match "v[0-9]*" $(git rev-list -1 HEAD $1) 2>/dev/null)
       fi
+      if [ -z "$VN" ]; then
+        # No tags at all, best we can do is tag the commit.
+        VN="$DEF_VER-$(git rev-parse --short HEAD 2>/dev/null)"
+      fi
     fi
     case "$VN" in
     *$LF*) (exit 1) ;;


MarkCallow:
Could the Cargo package manager be changed to do a non-sparse checkout?

Will a version created with this patch still show if the product was built from a modified  tree?

danhedron:
> Could the Cargo package manager be changed to do a non-sparse checkout?

I don't know, I'm only a user of Cargo. I assume there is a performance reason it checks out submodules as sparsely as possible.

This is only an issue when the crate using KTX-Software as a submodule is built via a git URL. Consider the `libktx-rs` crate. This can be built in 3 ways, depending on circumstances:

- A developer working on `libktx-rs`:
  - A full git checkout, including in submodules
  - :heavy_check_mark:  the version is determined when building
- Another crate using `libktx-rs` downloads a published version of `libktx-rs` from crates.io:
  - git is not used to download the package
  - :heavy_check_mark: the version is already present
- Another crate using an _unpublished_ version of `libktx-rs` specified by git URL:
  - Cargo will fetch `libktx-rs` using git and sparse checkout the `KTX-Software` submodule
  - :heavy_multiplication_x: The git repository is missing version tags

The empty version issue only occurs in the last case, where a git repository is present, but doesn't contain the tag information.

> Will a version created with this patch still show if the product was built from a modified tree?

When built this way the version ID looks like: `v4.0-bffa38f5-dirty`

MarkCallow:
Thank you very much @danhedron for the fix and the clear explanation of the circumstances causing the need for it.

PR #655, opened 2022-12-06
fix libwinpthread when use mingw

FuXiii:


FuXiii:
in `KTX-Software CI / mingw (pull_request) : Build Mingw x64 Debug`
```
D:/a/KTX-Software/KTX-Software/lib/astc_encode.cpp:44:16: error: conflicting declaration 'typedef void* pthread_t'
   44 | typedef HANDLE pthread_t;
      |                ^~~~~~~~~
In file included from c:\programdata\chocolatey\lib\mingw\tools\install\mingw64\include\c++\12.2.0\x86_64-w64-mingw32\bits\gthr-default.h:35,
                 from c:\programdata\chocolatey\lib\mingw\tools\install\mingw64\include\c++\12.2.0\x86_64-w64-mingw32\bits\gthr.h:[148](https://github.com/KhronosGroup/KTX-Software/actions/runs/3625601765/jobs/6113819568#step:8:149),
                 from c:\programdata\chocolatey\lib\mingw\tools\install\mingw64\include\c++\12.2.0\ext\atomicity.h:35,
                 from c:\programdata\chocolatey\lib\mingw\tools\install\mingw64\include\c++\12.2.0\bits\ios_base.h:39,
                 from c:\programdata\chocolatey\lib\mingw\tools\install\mingw64\include\c++\12.2.0\ios:42,
                 from c:\programdata\chocolatey\lib\mingw\tools\install\mingw64\include\c++\12.2.0\ostream:38,
                 from c:\programdata\chocolatey\lib\mingw\tools\install\mingw64\include\c++\12.2.0\iostream:39,
                 from D:/a/KTX-Software/KTX-Software/lib/astc_encode.cpp:21:
c:\programdata\chocolatey\lib\mingw\tools\install\mingw64\x86_64-w64-mingw32\include\pthread.h:197:19: note: previous declaration as 'typedef uintptr_t pthread_t'
  197 | typedef uintptr_t pthread_t;
      |                   ^~~~~~~~~
D:/a/KTX-Software/KTX-Software/lib/astc_encode.cpp:45:13: error: conflicting declaration 'typedef int pthread_attr_t'
   45 | typedef int pthread_attr_t;
      |             ^~~~~~~~~~~~~~
c:\programdata\chocolatey\lib\mingw\tools\install\mingw64\x86_64-w64-mingw32\include\pthread.h:238:31: note: previous declaration as 'typedef struct pthread_attr_t pthread_attr_t'
  238 | typedef struct pthread_attr_t pthread_attr_t;
      |                               ^~~~~~~~~~~~~~
D:/a/KTX-Software/KTX-Software/lib/astc_encode.cpp:49:1: error: 'int pthread_create(pthread_t*, const pthread_attr_t*, void* (*)(void*), void*)' was declared 'extern' and later 'static' [-fpermissive]
   49 | pthread_create(pthread_t* thread, const pthread_attr_t* attribs,
      | ^~~~~~~~~~~~~~
c:\programdata\chocolatey\lib\mingw\tools\install\mingw64\x86_64-w64-mingw32\include\pthread.h:301:26: note: previous declaration of 'int pthread_create(pthread_t*, const pthread_attr_t*, void* (*)(void*), void*)'
  301 | int       WINPTHREAD_API pthread_create(pthread_t *th, const pthread_attr_t *attr, void *(* func)(void *), void *arg);
      |                          ^~~~~~~~~~~~~~
D:/a/KTX-Software/KTX-Software/lib/astc_encode.cpp: In function 'int pthread_create(pthread_t*, const pthread_attr_t*, void* (*)(void*), void*)':
D:/a/KTX-Software/KTX-Software/lib/astc_encode.cpp:53:27: error: invalid conversion from 'HANDLE' {aka 'void*'} to 'pthread_t' {aka 'long long unsigned int'} [-fpermissive]
   53 |     *thread = CreateThread(nullptr, 0, func, thread_arg, 0, nullptr);
      |               ~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                           |
      |                           HANDLE {aka void*}
D:/a/KTX-Software/KTX-Software/lib/astc_encode.cpp: At global scope:
D:/a/KTX-Software/KTX-Software/lib/astc_encode.cpp:59:1: error: 'int pthread_join(pthread_t, void**)' was declared 'extern' and later 'static' [-fpermissive]
   59 | pthread_join(pthread_t thread, void** value) {
      | ^~~~~~~~~~~~
c:\programdata\chocolatey\lib\mingw\tools\install\mingw64\x86_64-w64-mingw32\include\pthread.h:302:26: note: previous declaration of 'int pthread_join(pthread_t, void**)'
  302 | int       WINPTHREAD_API pthread_join(pthread_t t, void **res);
      |                          ^~~~~~~~~~~~
D:/a/KTX-Software/KTX-Software/lib/astc_encode.cpp: In function 'int pthread_join(pthread_t, void**)':
D:/a/KTX-Software/KTX-Software/lib/astc_encode.cpp:61:25: error: invalid conversion from 'pthread_t' {aka 'long long unsigned int'} to 'HANDLE' {aka 'void*'} [-fpermissive]
   61 |     WaitForSingleObject(thread, INFINITE);
      |                         ^~~~~~
      |                         |
      |                         pthread_t {aka long long unsigned int}
```
It looks like before in to :
```
#define WIN32_LEAN_AND_MEAN
#include <Windows.h>

typedef HANDLE pthread_t;
typedef int pthread_attr_t;
```
had include `pthread.h` previously. (｡ŏ_ŏ)

MarkCallow:
The non-LLVM MingW with `ucrt` obviously defines pthread in an include that comes via the include of `<iostream>`. I don't know if the same is true for the earlier non-LLVM MingW that used msvcrt. Unfortunately you can't undefine a typedef.

We can do one of the following:
1. Drop this PR so always use pthreads when available bringing in the pthread library on LLVM MingW.
2. Find a way (preferably via compiler pre-defined macros) to test for LLVM MingW so we can avoid using pthreads there.
3. Don't use the names `pthread_t`, `pthread_attr_t`, `pthread_create` and `pthread_join`. Create our own types and defines which are set to the pthread definitions when non-Windows and to the Windows definitions when Windows.

@alvinhochun, @FuXiii what do you think?

alvinhochun:
Well, 1 (always use pthreads) is obviously the easy way. As a non-user of KTX I don't know if it is worth the effort to fix this properly on your side. If you do want to, then 3 -- basically creating a thin threading abstraction around pthread and win32 API -- sounds good to me.

I don't think 2 is a good idea, but if you _really_ want to you can test for the `_LIBCPP_VERSION` macro (defined by libc++) after including any c++ std headers.

MarkCallow:
@FuXiii thanks for looking creating this PR and finding `WIN_PTHREADS_H`. We all agree that is not a good idea to use this macro.

@alvinhochun thanks for your thoughts. 

I suspect that the implementations of pthreads in MingW are thin wrappers on top of Windows threading similar to what we would do in 3. Therefore using 1 is the obvious and easy choice. So I will close this PR.

alvinhochun:
Good. Using pthreads means that packagers will need to ship `libwinpthread-1.dll`, but it can be avoided by linking to winpthread statically. Adding the link flags `-Wl,-Bstatic -lwinpthread -Wl,-Bdynamic` should do it. There is also a hacky way to get this by deleting `libwinpthread.dll.a` from the lib dir so that it will link to the static lib `libwinpthread.a` instead. Just an additional note that might be useful.

(I do believe that winpthreads is a bit heavier than you may think, because it has to fully match the semantics of pthreads. Looking at the source, `pthread_create` has 102 lines of code, and there is a 72-line function that wraps around the user thread function.)

PR #657, opened 2022-12-10
fix typo in lib/info.c

simi:


MarkCallow:
Thank you for this fix.

The build failure appears to be because LunarG has dropped support for the Vulkan SDK version CI tries to install. I need to make some enquiries. Please wait a while.

Issue #652, opened 2022-11-28
Linkage error on Linux Garuda


adihodos:
Building the latest version from master fails on my machine with this linkage error:
`[ 91%] Building CXX object tools/ktxsc/CMakeFiles/ktxsc.dir/ktxsc.cpp.o
[ 91%] Linking CXX executable ktxsc
/usr/bin/ld: ../../libktx.a(astc_encode.cpp.o): in function `ktxTexture2_CompressAstcEx':
astc_encode.cpp:(.text+0x288): undefined reference to `astcenc_config_init(astcenc_profile, unsigned int, unsigned int, unsigned int, float, unsigned int, astcenc_config*)'
/usr/bin/ld: astc_encode.cpp:(.text+0x2b7): undefined reference to `astcenc_context_alloc(astcenc_config const*, unsigned int, astcenc_context**)'
/usr/bin/ld: astc_encode.cpp:(.text+0x31d): undefined reference to `astcenc_context_free(astcenc_context*)'
/usr/bin/ld: astc_encode.cpp:(.text+0x4bb): undefined reference to `astcenc_compress_image(astcenc_context*, astcenc_image*, astcenc_swizzle const*, unsigned char*, unsigned long, unsigned int)'
/usr/bin/ld: astc_encode.cpp:(.text+0x584): undefined reference to `astcenc_compress_reset(astcenc_context*)'
/usr/bin/ld: astc_encode.cpp:(.text+0x5d5): undefined reference to `astcenc_get_error_string(astcenc_error)'
/usr/bin/ld: astc_encode.cpp:(.text+0x6ba): undefined reference to `astcenc_context_free(astcenc_context*)'
/usr/bin/ld: ../../libktx.a(astc_encode.cpp.o): in function `compressionWorkloadRunner(int, int, void*)':
astc_encode.cpp:(.text+0x2ccb): undefined reference to `astcenc_compress_image(astcenc_context*, astcenc_image*, astcenc_swizzle const*, unsigned char*, unsigned long, unsigned int)'
clang-14: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [tools/ktxsc/CMakeFiles/ktxsc.dir/build.make:103: tools/ktxsc/ktxsc] Error 1
make[1]: *** [CMakeFiles/Makefile2:1456: tools/ktxsc/CMakeFiles/ktxsc.dir/all] Error 2
make: *** [Makefile:166: all] Error 2`

Configuration used:
[...]

PR #664, opened 2023-01-04
Fix static build on Linux

kachkov98:
Fix for #652 - GNU libtool incorrectly links static libraries archives, so do not use it on Linux


MarkCallow:
Thank you very much for this and the investigation you reported in #652.

Issue #663, opened 2023-01-04
Python bindings for libktx


ShukantPal:
Hello there from #479. Hopefully I was good to work with before :-)

I've learned Python and wrote some Python bindings for libktx. Do you think they should be implemented here as well?

MarkCallow:
Hi @ShukantPal, I think Python bindings would be a great addition. I have no more bandwidth. I haven't even been able make time to add your Java bindings implementation to  Maven Central. Therefore I would have to rely on you to completely take care of this. 

This means

- Writing code and tests
- If there is anything to build, ensuring it builds it CI
- Ensuring the tests run and pass in CI on all supported platforms
- Providing code in CI to deploy to whatever the usual Python distribution channel is.

If you are willing to do all this, then go for it.

I ended up spending a lot of time getting the Java bindings to build in CI on multiple platforms. I want to avoid that in this case.

ShukantPal:
Thanks @MarkCallow. I've completed the Python bindings. Now I'm working on adding it to the build process of this repository.

Can you point me to where the workflow for releases is (where would I add the config / code to upload the bindings to PyPI - the Python package index)?

ShukantPal:
@MarkCallow I found a way to do this − I created a "publish" workflow that you can manually trigger after a release if you want to publish to PyPI. It will fetch the latest release from this GitHub repositories and download the pyktx-related artifacts.

The only input needed from you is:
* register at https://pypi.org/account/register/ (no verification shenanigans this time :-))
* go to https://pypi.org/manage/account/#api-tokens & generate an API token
* create a secret that can be used in the workflow in GitHub called PYPI_API_TOKEN
(steps described [here](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#saving-credentials-on-github))

Voila, then you can go to Actions page in GitHub and trigger the workflow to publish to PyPI.

PR #665, opened 2023-01-08
pyktx (Python bindings for libktx)

ShukantPal:
#663

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -1,0 +1,117 @@
+# Copyright 2015-2020 The Khronos Group Inc.
+# SPDX-License-Identifier: Apache-2.0
+name: KTX-Software Android CI
+
+# Seems no way to avoid duplicating this on logic in each .yml file.
+# See https://github.com/actions/starter-workflows/issues/245.
+on:
+  # Trigger the workflow on a pull request,
+  pull_request:
+
+  push:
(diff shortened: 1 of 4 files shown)


ShukantPal:
Screenshot 2023-01-08 at 3 55 02 PM

Been pullin' my hair for a while and it turns out 64-bit Windows ships with 32-bit versions of Python. How convenient!

ci_scripts/build_win.ps1
```diff
@@ -20,7 +20,7 @@ function Set-Config-Variable {
     if ($res -eq $null) {
         $res = $DefaultValue
     }
-  } 
+  }
```
ShukantPal, line 23:
```suggestion
  } 
```
(revert)

MarkCallow:
I'm mystified by the VS2015 build issue. It shows as red but in the log it appears to have completed successfully. Don't worry about it. I shall be removing the VS2015 build from CI as soon as I've been able to release 4.1.0 and maybe sooner. Since I will not be able to merge this until after the release, there is no need to figure out what is wrong with VS2015.

How did you get the VS2017 build to find Python3? We've had issues  with that in another branch.

ShukantPal:
I added unit tests to the Python bindings and binded them to the CTest suite 🙌. I believe this round of work (PR) is now complete from my side.

ShukantPal:
Thanks for the update, will look into this.

PR #674, opened 2023-03-01
Fix build when Vulkan SDK path contains whitespaces

dBagrat:
Build failed on windows with Vulkan SDK installed in C:/Program Files/VulkanSDK
```
...
Generating VkFormat-related source files /usr/bin/bash: Files/VulkanSDK/1.3.216.0/Include: No such file or directory
...
```


MarkCallow:
Thank you for the fix.

PR #697, opened 2023-05-07
Remove pinned buffer list in JNI wrapper to avoid segmentation faults

### LINKED ISSUE #690: Crash when using Java interface on Windows
Hello,

First off, many thanks for providing toktx and the Java interface. I'm using this to compress about 12,000 GIS related textures to ASTC format, but I'm seeing a persistent crash in KtxTexture2.destroy().

I've done my best to reduce this to a simple testcase. Here it is:

```
package uk.co.abraded.toktx;

import org.khronos.ktx.KtxAstcParams;
import org.khronos.ktx.KtxCreateStorage;
import org.khronos.ktx.KtxErrorCode;
import org.khronos.ktx.KtxPackAstcBlockDimension;
import org.khronos.ktx.KtxPackAstcEncoderMode;
import org.khronos.ktx.KtxPackAstcQualityLevel;
import org.khronos.ktx.KtxTexture2;
import org.khronos.ktx.KtxTextureCreateInfo;
import org.khronos.ktx.VkFormat;

public class ToKTXBug {

   private static boolean nativeLoaded = false;

   public static void main(String[] args) {
      System.out.println("Main() >");

      loadNative();

      // Repeatedly create a compress an image. On my machine this segfaults
      // after about 500 iterations
      for (int i = 0; i < 10000; i++) {
         final int size = convertToASTC();
         System.out.println("Iteration: " + i + ", compressed data size is " + size);
      }

      System.out.println("Main() <");
   }

   private synchronized static void loadNative() {
      if (!nativeLoaded) {
         System.loadLibrary("ktx-jni");
         nativeLoaded = true;
      }
   }

   public static int convertToASTC() {
      loadNative();

      final int w = 1024;
      final int h = 1024;

      // Create Uncompressed texture
      final KtxTextureCreateInfo info = new KtxTextureCreateInfo();
      info.setBaseWidth(w);
      info.setBaseHeight(h);
      info.setVkFormat(VkFormat.VK_FORMAT_R8G8B8_SRGB); // Uncompressed
      final KtxTexture2 t = KtxTexture2.create(info, KtxCreateStorage.ALLOC);

      // Pass the uncompressed data
      int bufferSize = w * h * 3;
      final byte[] rgbBA = new byte[bufferSize];
      t.setImageFromMemory(0, 0, 0, rgbBA);

      // Compress the data
      final KtxAstcParams p = new KtxAstcParams();
      p.setBlockDimension(KtxPackAstcBlockDimension.D8x8);
      p.setMode(KtxPackAstcEncoderMode.LDR);
      p.setQualityLevel(KtxPackAstcQualityLevel.EXHAUSTIVE);
      final int rc = t.compressAstcEx(p);
      if (rc != KtxErrorCode.SUCCESS) {
         throw new RuntimeException("ASTC error " + rc);
      }
      final int retDataLen = (int) t.getDataSize();
      
      // Free things up - segfault usually occurs inside this destroy() call
      t.destroy();
      
      return retDataLen;
   }
}
```

On my Windows x64 box, this reliably crashes after about 400-500 iterations:

```
Iteration: 437, compressed data size is 262144
Iteration: 438, compressed data size is 262144
Iteration: 439, compressed data size is 262144
Iteration: 440, compressed data size is 262144
Iteration: 441, compressed data size is 262144
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  EXCEPTION_ACCESS_VIOLATION (0xc0000005) at pc=0x00007ffc9557403d, pid=2008, tid=3568
#
# JRE version: Java(TM) SE Runtime Environment (17.0.1+12) (build 17.0.1+12-LTS-39)
# Java VM: Java HotSpot(TM) 64-Bit Server VM (17.0.1+12-LTS-39, mixed mode, sharing, tiered, compressed oops, compressed class ptrs, g1 gc, windows-amd64)
# Problematic frame:
# V  [jvm.dll+0x3e403d]
#
# No core dump will be written. Minidumps are not enabled by default on client versions of Windows
#
# An error report file with more information is saved as:
# R:\svn\abraded\rob\code\trunk\src\modules\hs_err_pid2008.log
#
# If you would like to submit a bug report, please visit:
#   https://bugreport.java.com/bugreport/crash.jsp
#
```

The stack trace in the hotspot dump file is:

```
Java frames: (J=compiled Java code, j=interpreted, Vv=VM code)
J 738  org.khronos.ktx.KtxTexture.destroy()V libktx@4.1.0 (0 bytes) @ 0x00000260ac920fe3 [0x00000260ac920fa0+0x0000000000000043]
J 761 c1 uk.co.abraded.toktx.ToKTXBug.convertToASTC()I uk.co.abraded.toktx (136 bytes) @ 0x00000260a55b8e1c [0x00000260a55b85a0+0x000000000000087c]
j  uk.co.abraded.toktx.ToKTXBug.main([Ljava/lang/String;)V+20 uk.co.abraded.toktx
v  ~StubRoutines::call_stub
```

I managed to put some printfs in the native code, and it seems to be crashing inside the call to ReleaseByteArrayElements in KtxTexture.free_buffer_list(). The argment passed to ReleaseByteArrayElements look good and match those from the corresponding getByteArrayElements() so I suspect storage is getting trampled during execution of the encoder.

Two other comments:

1) My real code also fails in this way on Windows/ARM64, but I can't get this reduced testcase to fail.
2) Commenting out the call to destroy in this testcase avoids the issue, but the memory leakage means that I can't process all my 12,000 textures ina single run.

Many thanks in advance if you are able to look into this.
Regards,
Rob

ShukantPal:
Fixes #690 

After going through the C++ code in libktx, I realized that ktxTexture{1,2}_setImageFromMemory copy from the source buffers anyway. That meant pinning the source byte arrays in Java didn't do anything useful; hence, I removed that and the associated code keeping track of pinned buffers.

I have also added a unit test to make sure the JNI wrapper is stable for ~1000 iterations of usage.

### DIFF
--- a/interface/java_binding/src/test/java/org/khronos/ktx/test/KtxParallelTest.java
+++ b/interface/java_binding/src/test/java/org/khronos/ktx/test/KtxParallelTest.java
@@ -45,7 +45,7 @@ public class KtxParallelTest {
 
         public void run() {
             // Repeatedly create a compress an image.
-            for (int i = 0; i < 1000; i++) {
+            for (int i = 0; i < 300; i++) {
                 final int w = (testRandomizer.nextInt() % 512) + 1024;
                 final int h = w;
                 final int size = convertToASTC(w, h);
--- a/interface/java_binding/src/test/java/org/khronos/ktx/test/KtxParallelTest.java
+++ b/interface/java_binding/src/test/java/org/khronos/ktx/test/KtxParallelTest.java
@@ -1,0 +1,88 @@
+/*
+ * Copyright (c) 2023, Shukant Pal, robnugent, and Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.khronos.ktx.test;
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.khronos.ktx.*;
+import java.util.Random;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+@ExtendWith({ KtxTestLibraryLoader.class })
+public class KtxParallelTest {
+    private static final int NUM_THREADS = 2;
+    private static final Logger logger = Logger.getLogger(KtxParallelTest.class.getCanonicalName());
+
+    @Test
+    public void testParallelAstcConversion() throws InterruptedException {
+        final Thread[] runThreads = new Thread[NUM_THREADS];
+
+        for (int i = 0; i < NUM_THREADS; i++) {
+            final KtxTestRun run = new KtxTestRun(i);
+            final Thread runThread = new Thread(run);
+            runThread.setDaemon(false);
+            runThread.start();
+
+            runThreads[i] = runThread;
+        }
+
+        for (Thread thread : runThreads) {
+            thread.join();
+        }
+    }
+
+    private static class KtxTestRun implements Runnable {
+        private final int id;
+        private final Random testRandomizer = new Random();
+
+        public KtxTestRun(int id) {
+            this.id = id;
+        }
+
+        public void run() {
+            // Repeatedly create a compress an image.
+            for (int i = 0; i < 1000; i++) {
+                final int w = (testRandomizer.nextInt() % 512) + 1024;
+                final int h = w;
+                final int size = convertToASTC(w, h);
+
+                // Change level to INFO for logging
+                logger.log(Level.FINE,id + " iteration: " + i + ", size: " + w + "x" + h +  ", compressed data size is " + size);
+            }
+        }
+
+        public int convertToASTC(int w, int h) {
+            // Create Uncompressed texture
+            final KtxTextureCreateInfo info = new KtxTextureCreateInfo();
+            info.setBaseWidth(w);
+            info.setBaseHeight(h);
+            info.setVkFormat(VkFormat.VK_FORMAT_R8G8B8_SRGB); // Uncompressed
+            final KtxTexture2 t = KtxTexture2.create(info, KtxCreateStorage.ALLOC);
+
+            // Pass the uncompressed data
+            int bufferSize = w * h * 3;
+            final byte[] rgbBA = new byte[bufferSize];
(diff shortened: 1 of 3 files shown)


interface/java_binding/src/test/java/org/khronos/ktx/test/KtxParallelTest.java
```diff
@@ -0,0 +1,88 @@
...
+        private final int id;
+        private final Random testRandomizer = new Random();
+
+        public KtxTestRun(int id) {
+            this.id = id;
+        }
+
+        public void run() {
+            // Repeatedly create a compress an image.
+            for (int i = 0; i < 1000; i++) {
```
ShukantPal, line 48: Reducing this to ~300 because the tests are timing out in CI. The crashes I saw locally happened after ~250 textures (when around ~1GB of heap memory was accumulated, triggering a GC I think)

```suggestion
            for (int i = 0; i < 300; i++) {
```

MarkCallow:
Thank you @ShukantPal. You are correct that `ktxTexture{1,2}_SetImageFromMemory` copy from source.

robnugent:
@ShukantPal - first off, I just wanted to say many thanks again for looking into this problem.

FYI - as a heads-up for you, I think there is still a related issue. I built the latest code (with your fix) on Windows/ARM and Windows/x64 and when I take my multithreaded test from issue #690 and increase the number of iterations, then it looks like it's leaking memory. (This can be seen from Windows Task Manager).

The test eventually crashes the JVM on both platforms, where it should run OK forever, Note that this is *not* normal Java Heap memory variation, as if I limit that with e.g. the JVM flag -Xmx512m, then it still fails with a JVM crash. even consuming all the main memory and swap on my Windows/x64 box, which has 128GB RAM and 128GB swap.

I'm not confident in my own build of toktx (not sure I specified all the right build options), so I'm wont raise a bug at this stage, but I will do if the problem persists when I get an official 4.1.1 (or whatever) toktx build with your fix.

Thanks again
Rob

PR #698, opened 2023-05-08
pyktx (Python bindings for libktx)

ShukantPal:
Sorry re-opening #665 after I force pushed & accidentally deleted that PR. The history hasn't been changed


MarkCallow:
@ShukantPal please fix the latest conflicts. I'm almost ready to review and merge this.

I send you a direct e-mail about another matter. I'm not sure how effective you direct e-mail is so repeating here. I've set up an arm64 build on Ubuntu on Travis-CI that includes the Java wrapper. While CI seems to be running successfully, when I run it locally the Java wrapper test fails sometimes. Right now it is doing it every time. The error says ""The forked VM terminated without properly saying goodbye. VM crash or System.exit called?”. I’ve tried both JDK 11 (the default for Ubuntu 22.04) and JDK 17. At first the latter seemed to work but the problem has reappeared on my local machine though not (yet) on the CI machine.

Please investigate.

.appveyor.yml
```diff
@@ -20,11 +20,13 @@ environment:
   OPENGL_ES_EMULATOR_WIN: C:\Imagination\Windows_x86_64
   PACKAGE: NO
   PVR_SDK_HOME: https://github.com/powervr-graphics/Native_SDK/raw/master/lib/Windows_x86_64/
+  PYTHON: C:\Python311-x64\python.exe
```
MarkCallow, line 23: We stopped using appveyor in favor of workflows. You can delete this file.

.github/workflows/main.yml
```diff
@@ -48,6 +48,7 @@ jobs:
```
MarkCallow, line 48: main.yml has been replaced by separate files: windows.yml, mingw.yml and android.yml. Please add these changes to those files.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,31 @@
+# Copyright 2023 Shukant Pal
+# SPDX-License-Identifier: Apache-2.0
+name: Publish to PyPI
+
+on:
+  workflow_dispatch:
```
MarkCallow, line 6: We only want to publish when we make a release.

When is workflow_dispatch set?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,31 @@
...
+
+jobs:
+  pypi:
+    runs-on: ubuntu-latest
```
MarkCallow, line 10: The python bindings can be used on any platform, right? Ubuntu is just the choice for publishing, correct?

.gitignore
```diff
@@ -9,6 +9,7 @@
 /CMakeUserPresets.json
 /build*/
 /install*/
+/out/
```
MarkCallow, line 12: What goes in `out`. Why not put whatever it is under `build`. `out` is a very generic name. Can we use something more specific, e.g. `python-out`?

CMakeLists.txt
```diff
@@ -466,6 +468,16 @@ if(KTX_FEATURE_JNI)
     add_subdirectory(interface/java_binding)
 endif()
 
+# This is put here so CMake doesn't warn that "PYTHON" is unused if KTX_FEATURE_PY=OFF
+# PYTHON_EXECUTABLE is resolved in python_binding/CMakeLists.txt
```
MarkCallow, line 472: Where is PYTHON set?

Can we put this as an `else()` to the `if(KTX_FEATURE_PY)`?

interface/python_binding/LICENSE
```diff
@@ -0,0 +1,178 @@
+Apache License
```
MarkCallow, line 1: Why are you including the license here? Is this another GitHub repo that has been copied here?

interface/python_binding/pyktx/gl_internalformat.py
```diff
@@ -0,0 +1,102 @@
+# Copyright (c) 2021, Shukant Pal and Contributors
+# SPDX-License-Identifier: Apache-2.0
+
+from enum import IntEnum
+
+
+class GlInternalformat(IntEnum):
```
MarkCallow, line 7: Is this a complete list of all the OpenGL internal formats?

interface/python_binding/pyktx/vk_format.py
```diff
@@ -0,0 +1,195 @@
+# Copyright (c) 2023, Shukant Pal and Contributors
+# SPDX-License-Identifier: Apache-2.0
+
+from enum import IntEnum
+
+
+class VkFormat(IntEnum):
```
MarkCallow, line 7: Must have been painful creating this. It should probably be generated from the same mkvkformatfiles awk script that generates `lib/vkformat_enum.h`.

interface/python_binding/tests/test_ktx_texture.py
```diff
@@ -0,0 +1,26 @@
+# Copyright (c) 2023, Shukant Pal and Contributors
+# SPDX-License-Identifier: Apache-2.0
+
+from pyktx import *
+from test_config import __test_images__
+import unittest
+
+
+class TestKtxTexture(unittest.TestCase):
```
MarkCallow, line 9: How are the tests run? Are they always run as part of the build? I didn't notice anything in the CI additions.

MarkCallow (commented):
Please review my comments and make changes as appropriate. They're mostly in the changes to existing files. I know nothing about Python wrappers so can't comment on the new files.

Is there documentation for this wrapper? I don't recall noticing anything in the new files.

Note that when the `ktxtools` branch is merged this wrapper will require changes. For example `ktxtools` has new error codes and supports an additional supercompression option.

.github/workflows/publish.yml
ShukantPal: You can manually trigger this action here: https://github.com/KhronosGroup/KTX-Software/actions

.github/workflows/publish.yml
ShukantPal: That's right!

interface/python_binding/LICENSE
ShukantPal: The LICENSE file is published to [Python Package Index](https://pypi.org).

interface/python_binding/tests/test_ktx_texture.py
ShukantPal: They are automatically run as a part of the build. See `python_binding/build.py`.  These tests are added to CTest in the python_binding/CMakeLists.txt file.

.github/workflows/main.yml
ShukantPal: I updated the Windows build.

CMakeLists.txt
ShukantPal: I removed that snippet - seems like it's unnecessary now.

interface/python_binding/pyktx/gl_internalformat.py
ShukantPal: There is a 1:1 match between these formats and the ones defined in the Java interface.

interface/python_binding/pyktx/vk_format.py
ShukantPal: I copied the Java VkFormats file and did some find/replace. Is it necessary to use awk, since these are static files?

.gitignore
ShukantPal: I removed this from .gitignore - not necessary anymore.

ShukantPal:
@MarkCallow I followed up on everything I believe. And the build is passing as far as I see.

ShukantPal:
@MarkCallow I think you need to remove the AppVeyor webhook.

.travis.yml
```diff
@@ -41,36 +41,40 @@ env:
     # them during CI. CI runs on x86_64 and there is a PostBuild command
     # that attempts to execute the compiled-for-m1 tests to have gtest
     # add the list of tests to the cmake test runner.
+    #
+    # FEATURE_PY is off for arm64 macOS because we'd need a cross-compiled version
```
MarkCallow, line 45: Does anything need to be included in the install package for Python or is the wrapper installed from the Python package manager after KTX-Software is installed?

If the former I'd hate to leave it out of the macOS/arm package. Is a cross-compiling version of Python available? If the latter will the extensions be compiled on first use? When I was using GYP, it came as Python source code. On first use it was compiled and saved.

ci_scripts/build_linux.sh
```diff
@@ -25,6 +25,8 @@ for i in $@; do
   eval $i
 done
 
+echo which python $(which python) and path is $PATH
```
MarkCallow, line 28: Do we need to keep this? If it is for debugging the CI build, comment it out.

MarkCallow:
> @MarkCallow I think you need to remove the AppVeyor webhook.

I removed it a couple of days ago. I think it is too late for this PR.

MarkCallow:
You haven't answered my question about the documentation for this wrapper.

.travis.yml
ShukantPal: It's the latter. We're publishing a source distribution of the Python package, which means that our package should compile itself when it's installed.

ci_scripts/build_linux.sh
```diff
@@ -25,6 +25,8 @@ for i in $@; do
...
+
```
ShukantPal, line 29:
```suggestion
```

ci_scripts/build_linux.sh
ShukantPal: I removed the line. It's no longer needed; I also have a Linux machine now so I find the standard Python path myself :)

MarkCallow:
> > @MarkCallow I think you need to remove the AppVeyor webhook.
> 
> I removed it a couple of days ago. I think it is too late for this PR.

I just created a new PR and it still shows appveyor builds which, of course, fail as there is no {.,}appveyor.yml. I do not understand why. Further investigation needed.

MarkCallow:
> Further investigation needed.

Found the issue. There is an AppVeyor GitHub app from which access to KTX-Software needs to be removed. The Khronos admin will have to do it. Request sent.

MarkCallow:
We've now merged the last in-progress major work. Sorry to make you to rebase again. This will be the last time.

> Where would you like me to add a note in the Doxygen-based documentation?

I'll get back to you on that.

> It's the latter. We're publishing a source distribution of the Python package, which means that our package should compile itself when it's installed.

Does the source distribution need to be included in our installer packages or will people get it from the python package manager to which, if I'm understanding correctly, that workflow_dispatch job uploads the source.

.travis.yml
MarkCallow: I replied earlier in the wrong place. Here is my question again:

Does the source distribution need to be included in our installer packages or will people get it from the python package manager to which, if I'm understanding correctly, that workflow_dispatch job uploads the source?

What does FEATURE_PY=ON do?

MarkCallow (commented):
Looks fine. Still need to resolve, for me anyway, documentation and packaging.

ShukantPal:
Sorry was extra busy these past couple of weeks. I'll finish this over the weekend.

.travis.yml
ShukantPal: > Does the source distribution need to be included in our installer packages or will people get it from the python package manager to which, if I'm understanding correctly, that workflow_dispatch job uploads the source?

That’s correct. The package manager “pip” will automatically pull the source distribution from PyPI and compile the Python C bindings.

> What does FEATURE_PY=ON do?

* Build the source distribution
* Compile the bindings for the host architecture
* Build a binary distribution
* Run unit tests

.github/workflows/publish.yml
MarkCallow, line 6: Please add

```
push:
    # Also trigger on push of release tags to any branch. Useful
    # for testing release builds before merging to main.
    tags:
      - 'v[0-9]+\.[0-9]+\.[0-9]+'
      - 'v[0-9]+\.[0-9]+\.[0-9]+-*'
```

The last line is for alphas, betas and release candidates. Do you think we should avoid deploying pre-releases?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,31 @@
...
+    steps:
```
MarkCallow, line 11: Add

```
# Hack to workaround "on:" not supporting AND condition.
- name: Exit if not on main branch
  if: endsWith(github.ref, 'main') == false
  run: exit 0
```

MarkCallow (changes requested):
I've requested a couple of additions to automate deployment when a release tag is created in main.

.travis.yml
MarkCallow: What is the purpose of building a binary distribution? We are deployng source to PyPl.

MarkCallow:
> There isn't any special documentation for the wrapper as it follows the Java (naming) patterns pretty closely, but in a Pythonic way. Where would you like me to add a note in the Doxygen-based documentation?

If you add documentation to the .py files in the standard Python way, can you then use the Python tools to generate a single .html doc file or at least a set of files with a .index.html? If so we can add a link to that documentation by editing the `navindex` element of the `*DoxyLayout` files in `pkgdoc`.

.github/workflows/publish.yml
ShukantPal: Done

ShukantPal: I think it’s ok to deploy pre-releases. You will have to update the version in [this file](https://github.com/ShukantPal/KTX-Software/blob/5505ac4e902d5fdf7675c82457fa37217bcc7c46/interface/python_binding/setup.py#L7).

.github/workflows/publish.yml
ShukantPal: Done. I changed it to `exit 1` so it fails instead of succeeds.

.travis.yml
ShukantPal: When available, binary distributions save time.

.github/workflows/publish.yml
MarkCallow: `exit 0` is correct. The idea is to make it as if the build was never run. Please change it back.

.github/workflows/publish.yml
MarkCallow: Can `setup.py` get the version from the libktx version file, `lib/version.h`? So far I've managed to avoid having to manually update any versions prior to making releases. Everything comes automatically from the git tag name. I even managed to coerce Maven's `pom.xml` into doing this for the JNI build.

.travis.yml
MarkCallow: How are binary distributions distributed?

ShukantPal: By pip? It's the same package but you can provide both source and binary dists

.github/workflows/publish.yml
ShukantPal: `exit 0` will cause the step to exit successfully and it will not stop the GitHub action from continuing.

.travis.yml
MarkCallow: So should we be deploying binary distributions to PyPl too?

.github/workflows/publish.yml
ShukantPal: I think we can split it into 2 jobs. The first one skips a step if not on main branch, and the second one checks if that step ran in the first job using the `if` syntax.

https://stackoverflow.com/questions/60589373/how-to-force-to-exit-in-github-actions-step#61832535

.travis.yml
ShukantPal: Sorry "pip" is the executable that uses the "PyPI" website to get the source/binary distributions.

We are uploading both dists to PyPI and people will use pip to install them.

.github/workflows/publish.yml
ShukantPal: Did you do that in the CMake build?

.travis.yml
MarkCallow: Ahh! Which binaries is CI building and uploading. Are there more we should add?

.github/workflows/publish.yml
MarkCallow: Let's try the 2 jobs idea. However step outputs have been deprecated since that SO answer was written. You now have to write environment variables into the file `$GITHUB_ENV` and look for those variables in subsequent steps or jobs.

ShukantPal: Hi @MarkCallow, after taking a fresh look; it seems like a single `if` condition on the branch at the job-level should suffice. Please see my latest change.

(The 2 step workflow made sense if you can’t directly resolve the “if” true/false in without running a command)

.github/workflows/publish.yml
ShukantPal: Thanks for the pointer. Since `setup.py` is a Python script, it can read environment variables. I injected `LIBKTX_VERSION` using the existing CMake build.

.travis.yml
ShukantPal: It’s basically building the glue C code for the Python bindings. It saves compilation time when installing the Python package. This isn’t a necessity but if this package becomes popular, it might make sense to invest in uploading binaries for all platforms.

interface/python_binding/CMakeLists.txt
```diff
@@ -0,0 +1,103 @@
...
+add_custom_command(
+    TARGET pyktx
+    PRE_BUILD
+    COMMAND
+        ${PYTHON_EXECUTABLE} -m pip install -r ${CMAKE_SOURCE_DIR}/interface/python_binding/requirements.txt
+    COMMENT
+        "Install dependencies for pyktx build"
+)
+
+if (NOT WIN32) # Heap corruption on Windows somehow
```
MarkCallow, line 79: Does this mean it can't be built on Windows?

interface/python_binding/setup.py
```diff
@@ -0,0 +1,28 @@
+# Copyright (c) 2023, Shukant Pal and Contributors
+# SPDX-License-Identifier: Apache-2.0
+
+import os
+from setuptools import setup
+
+__name__    = 'pyktx'
+__version__ = os.environ['LIBKTX_VERSION']
```
MarkCallow, line 8: Is KTX_VERSION set in the environment by the CMake build or is it only in CI builds?

interface/python_binding/build.py
```diff
@@ -0,0 +1,197 @@
...
+    if LIBKTX_INCLUDE_DIR is None:
+        LIBKTX_INCLUDE_DIR = LIBKTX_INSTALL_DIR + '\\include'
+    if LIBKTX_LIB_DIR is None:
+        LIBKTX_LIB_DIR = LIBKTX_INSTALL_DIR + '\\lib'
+elif os.name == 'osx':
+    if LIBKTX_INCLUDE_DIR is None:
+        LIBKTX_INCLUDE_DIR = '/usr/local/include'
+    if LIBKTX_LIB_DIR is None:
+        LIBKTX_LIB_DIR = '/usr/local/lib'
+
```
MarkCallow, line 25: What about defaults GNU/Linux?

MarkCallow (changes requested):
Thanks for adding the documentation. How is a document created from the documentation comments? I don't see anything in `build.py` to do it.

I have a couple of questions and spotted some minor issues that need fixing.

interface/python_binding/pyktx/ktx_texture.py
```diff
@@ -80,23 +120,59 @@ def kv_data_raw(self) -> Optional[c_buffer]:
...
 
     @property
     def data_size_uncompressed(self) -> int:
+        """Byte length of the texture's uncompressed image data."""
+
         return lib.ktxTexture_GetDataSizeUncompressed(self._ptr)
 
     def row_pitch(self, level: int) -> int:
+        """
+        Return pitch betweeb rows of a texture image level in bytes.
```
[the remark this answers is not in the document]
ShukantPal:
```suggestion
        Return pitch between rows of a texture image level in bytes.
```

interface/python_binding/setup.py
ShukantPal: CMake build.

ci_scripts/build_linux.sh
```diff
@@ -37,7 +37,9 @@ else
   # No Vulkan SDK yet for Linux/arm64.
   FEATURE_LOADTESTS=${FEATURE_LOADTESTS:-OpenGL}
 fi
+FEATURE_PY=${FEATURE_PY:-OFF}
 FEATURE_TESTS=${FEATURE_TESTS:-ON}
+FEATURE_LOADTESTS=${FEATURE_LOADTESTS:-ON}
```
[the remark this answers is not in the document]
ShukantPal:
```suggestion
```

interface/python_binding/build.py
ShukantPal: libktx installs itself the system header/library directory’s on GNU/Linux. That’s why explicitly specifying these isn’t necessary (and why the build passes on Linux CI).

.github/workflows/publish.yml
```diff
@@ -0,0 +1,42 @@
...
+
+jobs:
+  pypi:
+    runs-on: ubuntu-latest
+    if: endsWith(github.event.base_ref, 'main')
+    steps:
+      # Hack to workaround "on:" not supporting AND condition.
+      - name: Exit if not on main branch
+        if: endsWith(github.event.base_ref, 'main') == false
+        run: exit 1
```
ShukantPal, line 22:
```suggestion
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,42 @@
...
+    # Trigger on push of release tags. There is no way to limit the trigger
+    # to a specific branch. A later build step checks for the main branch.
+    tags:
+      - 'v[0-9]+\.[0-9]+\.[0-9]+'
+      - 'v[0-9]+\.[0-9]+\.[0-9]+-*'
+
+jobs:
+  pypi:
+    runs-on: ubuntu-latest
+    if: endsWith(github.event.base_ref, 'main')
```
ShukantPal, line 17: @MarkCallow Here is the change re: running only on main

.github/workflows/publish.yml
ShukantPal: Sorry, I didn’t delete the old code. I pinged you on the new line added (it’s above).

interface/python_binding/pyktx/ktx_supercmp_scheme.py
```diff
@@ -5,6 +5,13 @@
...
+    """Enumerators identifying the supercompression scheme."""
+
     NONE = 0
+    """No supercompression."""
+
     BASIS_LZ = 1
+    """Basis LZ supercompression."""
+
     ZSTD = 2
+    """ZStd supercompression."""
```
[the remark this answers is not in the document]
ShukantPal: Added.

interface/python_binding/CMakeLists.txt
ShukantPal: I think this was due to the previous CI setup. I’ll see if the Python tests can run on Windows CI now.

ShukantPal:
> Thanks for adding the documentation. How is a document created from the documentation comments? I don't see anything in build.py to do it.

Yeah I haven’t completed that yet. Next thing to do!

interface/python_binding/CMakeLists.txt
ShukantPal: They are now passing, and I removed the “if not win32 check”.

ShukantPal:
@MarkCallow Ok, the CMake build now will generate a static website using Sphinx with pyktx documentation.

Screenshot 2023-07-31 at 9 41 12 PM

Screenshot 2023-07-31 at 9 41 23 PM

MarkCallow:
Sorry for the long delay over this. I went to LA to attend Siggraph and soon after my return I had to attend a funeral.

The output html file(s?) for the documentation need to be installed in a subdirectory of `${CMAKE_CURRENT_BINARY_DIR}/docs/html` and a line like

```
    <tab type="user"  url="libktx/index.html" title="libktx Reference"/>
```

must be added to `pkgdoc/packageDoxyLayout.xml` to point at the index.html for the new pages.

Other than that, this is looking good.

MarkCallow:
ping @ShukantPal Is the documentation being installed in the right place or is work needed? The place I am asking for it to be installed is the directory from which the on-line documentation is deployed.

ShukantPal:
Hi Mark, I will work on this soon. I just moved across the country.

ShukantPal:
@MarkCallow I added the pyktx documentation to the Doxygen website.

Screenshot 2023-09-04 at 3 07 32 PM

MarkCallow:
> I added the pyktx documentation to the Doxygen website.

The change looks good, except that I forgot to tell you that all the `*DoxyLayout.xml` files in `pkgdoc` need to be modified in a similar way. This is so the top menu bar maintains a consistent appearance across documents.

I am a concerned about the usability implications of clicking on the pyktx link and getting a completely different document and layout structure without the top menu bar to go to of the other documents. Users will have to use the browser's back feature to return to the other documents. Do you have any thoughts about how to make the transition less brutal?

The Windows build failures are caused by a new warning in a clang compiler update in GitHub's Windows CI runners. I am working on a fix. I'll let you know when it is ready. You'll have to rebase again once I've posted it.

The Emscripten build failure (Travis) looks like it may have been triggered by network issues. I've kicked off another try to build the job.

ShukantPal:
The Python documentation is not generated by Doxygen. The Python docs look like most other Python library docs, e.g. https://cffi.readthedocs.io/en/latest/

Since most Python devs would expect this layout, I'm not sure this is a problem tbh.

MarkCallow:
> The Python documentation is not generated by Doxygen. The Python docs look like most other Python library docs, e.g. https://cffi.readthedocs.io/en/latest/

Yes I am aware of that.

> 
> Since most Python devs would expect this layout, I'm not sure this is a problem tbh.

My concern is about not having an obvious way to get back to the top-level package document not the different layout. Anyway, please update the other `*DoxyLayout.xml` files then I will make a branch from the PR in my workarea and see what the docs look like in reality.

MarkCallow:
The fix for clang 16 is now in `main`.

ShukantPal:
@MarkCallow I've added the DoxyLayout changes.

MarkCallow:
Thanks. I'll pull your branch and check out the documentation as soon as I can.

Does `interface/python_binding/_static/.keep` really need to be checked into the git repo?

I hope you are settling in after your move. I moved not long ago. It is very stressful.

harelfar2:
Hi @ShukantPal. 
Thanks for all your work and effort on this, I see that you've been on this for a long time.

I just encountered in my work for the need to process .ktx images. 
I didn't get from this PR if you will publish a library to pypi and I (and everyone else) will have the ability to open those images and convert them to a more familiar format. If not, how exactly will I be able to use this?

ShukantPal:
Hi @harelfar2, this library doesn't provide direct conversion. But it will let you get the raw pixels from .ktx images. You can use those to save the image into another format (using another library for that format).

harelfar2:
@ShukantPal.
Sounds good, you mean raw numpy pixels, right? 

Still, I didn't get from your PR how can I use it.. Will it be available in pypi?
Can you add some usage example on the readme or something? Perhaps on a side readme in a subdirectory?

Thanks!

MarkCallow:
> Still, I didn't get from your PR how can I use it.. Will it be available in pypi? Can you add some usage example on the readme or something? Perhaps on a side readme in a subdirectory?

This PR is a python wrapper for libktx and exposes the functionality of libktx. Until this PR is merged and a release built, you can look at the [libktx c API documentation](https://github.khronos.org/KTX-Software/libktx/index.html) to see the functionality. The wrapper is for Python programmers who wish to write tools to manipulate KTX files. If you just want to manipulate images from a KTX file the existing command line tools may meet your needs. Look at the [v4.3.0 tools documentation](https://github.khronos.org/KTX-Software/ktxtools/index.html) to see what kinds of manipulation are supported today. These are available for GNU/Linux, macOS and Windows.

I'll leave @ShukantPal to answer the question about pypl.

> Sounds good, you mean raw numpy pixels, right?
> 

A KTX file is a container that may, and in all likelihood will, contain multiple images. All pixels in all images will have the same format chosen from a large selection of possibilities: most Vulkan VkFormats  plus a couple of transcodable formats. (The file provides information about the format via one or both of the VkFormat enum and a detailed Data Format Descriptor.) An image read from a KTX file via the API is delivered in its described format. Conversion of an image to another format is up to the application. `libktx` does provide functionality to transcode images in transcodable formats to an R8G8B8A8 format simplifying the application's work in this case. v4.3.0's `ktx extract` tool can extract all or specific images from a KTX file and save them as PNG or EXR files, depending on the format.

harelfar2:
got you.
thanks a lot!

MarkCallow:
Hi @ShukantPal, I am seeing the following warnings during the python build in the logs for the Windows vs2019 image on GitHub, probably for the vs2022 image as well. Please fix.

```
EXEC : warning : The script wheel.exe is installed in 'C:\hostedtoolcache\windows\Python\3.11.4\x64\Scripts' which is not on PATH. [D:\a\KTX-Software\KTX-Software\build\interface\python_binding\pyktx.vcxproj]
    Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.
    Attempting uninstall: setuptools
      Found existing installation: setuptools 65.5.0
      Uninstalling setuptools-65.5.0:
        Successfully uninstalled setuptools-65.5.0
EXEC : warning : The script pygmentize.exe is installed in 'C:\hostedtoolcache\windows\Python\3.11.4\x64\Scripts' which is not on PATH. [D:\a\KTX-Software\KTX-Software\build\interface\python_binding\pyktx.vcxproj]
    Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.
EXEC : warning : The script normalizer.exe is installed in 'C:\hostedtoolcache\windows\Python\3.11.4\x64\Scripts' which is not on PATH. [D:\a\KTX-Software\KTX-Software\build\interface\python_binding\pyktx.vcxproj]
    Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.
EXEC : warning : The script pybabel.exe is installed in 'C:\hostedtoolcache\windows\Python\3.11.4\x64\Scripts' which is not on PATH. [D:\a\KTX-Software\KTX-Software\build\interface\python_binding\pyktx.vcxproj]
    Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.
EXEC : warning : The scripts sphinx-apidoc.exe, sphinx-autogen.exe, sphinx-build.exe and sphinx-quickstart.exe are installed in 'C:\hostedtoolcache\windows\Python\3.11.4\x64\Scripts' which is not on PATH. [D:\a\KTX-Software\KTX-Software\build\interface\python_binding\pyktx.vcxproj]
    Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.
```
See https://github.com/KhronosGroup/KTX-Software/actions/runs/6218234270/job/16874416648.

I haven't had time to review the documentation integration yet.

MarkCallow:
Hi @ShukantPal please fix the warnings.

ShukantPal:
@MarkCallow Have you time for reviewing the documentation?

MarkCallow:
The warnings are still there in both the windows-2019 and windows-latest builds. Please fix them.

What do I need to install on my MacBook in order to build the documentation so that I can build and review it?

ShukantPal:
You don’t need to install anything. If you enable the documentation flag in CMake, the docs should be auto-generated. Any libraries used in generating the documentation are automatically installed using `pip`

ShukantPal:
Seems like my commit didn't get pushed from my workstation. I'll check what happens when I'm at work tomorrow.

MarkCallow:
> Seems like my commit didn't get pushed from my workstation. I'll check what happens when I'm at work tomorrow.

Ping!

interface/python_binding/pyktx/vk_format.py
```diff
@@ -0,0 +1,197 @@
...
+    VK_FORMAT_B12X4G12X4R12X4G12X4_422_UNORM_4PACK16 = 1000156021
+    VK_FORMAT_PVRTC1_2BPP_UNORM_BLOCK_IMG = 1000054000
+    VK_FORMAT_PVRTC1_4BPP_UNORM_BLOCK_IMG = 1000054001
+    VK_FORMAT_PVRTC2_2BPP_UNORM_BLOCK_IMG = 1000054002
+    VK_FORMAT_PVRTC2_4BPP_UNORM_BLOCK_IMG = 1000054003
+    VK_FORMAT_PVRTC1_2BPP_SRGB_BLOCK_IMG = 1000054004
+    VK_FORMAT_PVRTC1_4BPP_SRGB_BLOCK_IMG = 1000054005
+    VK_FORMAT_PVRTC2_2BPP_SRGB_BLOCK_IMG = 1000054006
+    VK_FORMAT_PVRTC2_4BPP_SRGB_BLOCK_IMG = 1000054007
+    VK_FORMAT_ASTC_4x4_SFLOAT_BLOCK_EXT = 1000066000
```
MarkCallow, line 182: The formats from here, line 182, to the end of the file, line 197 have been added to Vulkan core. Therefore please remove the _EXT from each of these lines and, of course, anywhere that is using them.

ShukantPal: I've removed the `_EXT` here.

MarkCallow:
Let's get this over the finish line. Please fix the remaining warnings. I will review the documentation within the next cojple of days.

The recent build failure seemed to be the result of a network issue on the GitHub CI runner. I've started a re-run.

MarkCallow:
I made a local branch with this and tried to build on my M2 MacBook Pro. I got

```
ld: warning: ignoring file '/Users/mark/Projects/khronos/github/KTX-Software/build/macos/Debug/libktx.4.3.0.dylib': found architecture 'arm64', required architecture 'x86_64'
```

Why is it requiring x86_64? It needs to support either without any warning.

I also got

```
Creating file docs/modules.rst.
/bin/sh: sphinx-build: command not found
```

which is presumably why nothing showed up in my output docs directory. What do I need to install for `sphinx-build`?

There are still some warnings in the VS2019 build: https://github.com/KhronosGroup/KTX-Software/actions/runs/6679753019/job/18176456781?pr=698.

ShukantPal:
I develop on a M1/M2 MacBook Pro. I'll need more details about how you tried to build it to understand why it failed.

`sphinx-build` is what generates the documentation. Did you enable `KTX_FEATURE_PY`?

ShukantPal:
@MarkCallow Make sure your Python version is arm64. You can run the following to check this:

```python
python3 -c "import platform; is_virtualizer = platform.machine() in {'x86_64', 'i386'}; print('Python is running on macOS\'s x86 virtualizer.' if is_virtualizer else 'Python is not running on macOS\'s x86 virtualizer.')"
```

I also updated the CMake build to be more reliable after a clean clone of the repo.

ShukantPal:
GitHub seems to be down right now. I’ll come back to this over the weekend.

MarkCallow:
> I develop on a M1/M2 MacBook Pro. I'll need more details about how you tried to build it to understand why it failed.
>
> sphinx-build is what generates the documentation. Did you enable KTX_FEATURE_PY?

Via the CMake GUI I did a config and generate with `KTX_FEATURE_PY` and `KTX_FEATURE_DOC` enabled. Then in Xcode I built the `pyktx` target.

I received an e-mail from this thread, whose corresponding comment I can't find, asking me about my Python3 version. I am using the one in /usr/bin which, as far as I can tell, from where it stashes stuff, was installed as part of the Xcode command line tools package. I ran the script you provided.

```bash
 $ python3 -c "import platform; is_virtualizer = platform.machine() in {'x86_64', 'i386'}; print('Python is running on macOS\'s x86 virtualizer.' if is_virtualizer else 'Python is not running on macOS\'s x86 virtualizer.')"
Python is not running on macOS's x86 virtualizer.
```

I pulled your latest after the above commits. A build of `pyktx` now succeeds but I still get the architecture warning. The documentation did not appear in my docs output directory and there was no pyktx in the table of contents so I built the `all-docs` target. I got the following error:

```
/Users/mark/Projects/khronos/github/KTX-Software/build/macos/build/KTX-Software.build/Debug/pyktx-dev.build/Script-44C90E35038C9494D9AC2C14.sh: line 5: IF:0, ,:: No such file or directory
Command PhaseScriptExecution failed with a nonzero exit code
```

The table of contents now contains "pyktx Reference" but the associated file is not found when I click the link:

```
Firefox can’t find the file at /Users/mark/Projects/khronos/github/KTX-Software/build/macos/docs/html/pyktx/index.html.
```

MarkCallow:
Unfortunately the build on M2 macOS is still failing and the warning is still being issued:

```
/Users/mark/Projects/khronos/github/KTX-Software/ld:1:1: ignoring file '/Users/mark/Projects/khronos/github/KTX-Software/build/macos/Debug/libktx.4.3.0.dylib': found architecture 'arm64', required architecture 'x86_64'

make: *** [html] Error 127
sphinx-build: command not found
```

On the warning, is this attempting to compile the python for x86_64?

One the error, do I need to manually install sphinx-build?

ShukantPal:
> On the warning, is this attempting to compile the python for x86_64?

No.

> On the error, do I need to manually install sphinx-build?

You shouldn't have to. It works on my machine and in the CI too. At this point, you should try to install it manually `python3 -m pip install sphinx`

Make sure your pip programs can be accessed on the PATH. For me,
```txt
> which sphinx-build
/opt/homebrew/bin/sphinx-build

> cat $(which sphinx-build)
#!/opt/homebrew/opt/python@3.11/bin/python3.11
# -*- coding: utf-8 -*-
import re
import sys
from sphinx.cmd.build import main
if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])
    sys.exit(main())
```

> This RuntimeWarning still appears in the Windows build

That's an issue with pip trying to detect if CPython is a debug build. It's a harmless warning.

https://github.com/pypa/wheel/blob/b90a4bc408ce621818b3e92bb55955431a7fa22a/src/wheel/bdist_wheel.py#L109

ShukantPal:
I'm able to reproduce the `sphinx-build` problem if I try to build using Xcode IDE. I'm not sure what Xcode is doing here.

ShukantPal:
I pushed a change that should hopefully make it work on Xcode now.

MarkCallow:
I tried your latest changes. I was able to build and sphinx-build was found. I got 3 warnings. This one

```
Running Sphinx v6.2.1

/Users/******/Library/Python/3.9/lib/python/site-packages/urllib3/__init__.py:34: NotOpenSSLWarning: urllib3 v2.0 only supports OpenSSL 1.1.1+, currently the 'ssl' module is compiled with 'LibreSSL 2.8.3'. See: https://github.com/urllib3/urllib3/issues/3020
```

which isn't labelled with "warning" so Xcode doesn't flag it, and two instances of
```
/Users/******/Projects/khronos/github/KTX-Software/ld:1:1: ignoring file '/Users/******/Projects/khronos/github/KTX-Software/build/macos/Debug/libktx.4.3.0.dylib': found architecture 'arm64', required architecture 'x86_64'
```

which you've said is harmless. However since it does get flagged by Xcode I'd really like to find a way to stop it.

I found some other issues:

1. After building the `pyktx` target the docs appear in `${CMAKE_BUILD_DIRECTORY}/interface/python_binding/docs` and in `${CMAKE_SOURCE_DIRECTORY}/interface/python_binding/docs` but not in `${CMAKE_BUILD_DIRECTORY}/docs` which is the build target for all other docs. You have to build the `all.doc` target for them to appear in `${CMAKE_BUILD_DIRECTORY}/docs`. I'd like to avoid the duplicate docs in `${CMAKE_BUILD_DIRECTORY}/interface/python_binding/docs` and `${CMAKE_SOURCE_DIRECTORY}/interface/python_binding/docs`. My preference is to have the docs built only to `${CMAKE_BUILD_DIRECTORY}/docs`.
2. Most of build ends up in `${CMAKE_SOURCE_DIRECTORY}/interface/python_binding/{build,dist}`. I'd prefer if the build artifacts went to `${CMAKE_BUILD_DIRECTORY}/interface/python_binding.
3. Xcode's "Clean Build Folder" did seem to clean any of the build artifacts from either `${CMAKE_SOURCE_DIRECTORY}/interface/python_binding/{build,dist}` or `${CMAKE_BUILD_DIRECTORY}/interface/python_binding`.
4. After this I manually removed `${CMAKE_BUILD_DIRECTORY}/interface/python_binding` because I wanted to see if the docs here were created when building `pyktx` or `all.doc`. The build failed with
    ```
    /Users/mark/Projects/khronos/github/KTX-Software/build/macos/build/KTX-Software.build/Debug/pyktx-deps.build/Script-80DFDEDA6A2296B2958EB164.sh: line 4: cd: /Users/mark/Projects/khronos/github/KTX-Software/build/macos/interface/python_binding: No such file or directory
    ```
    I had to manually recreate the directory. Not sure how it originally got created nor why it failed this time.

5. Sometimes the build stalled for a long time here
    ``` 
    Successfully built pyktx-4.3.0.tar.gz
    * Creating venv isolated environment...
    * Installing packages in isolated environment... (setuptools >= 40.8.0, wheel)
    * Getting build dependencies for wheel...
    * Installing packages in isolated environment... (cffi>=1.15.1, wheel)
    * Building wheel...

    ```

ShukantPal:
@MarkCallow I'm not getting the warning from `ld` when building pyktx, nor does the error appear in the Travis CI logs. There are other warnings coming `ld` that already exist in the `main` branch, so I'm sure this one is harmless. I don't want to spend too much time on warnings. Here are my Xcode build logs, and I'm not getting warnings:
[build.txt](https://github.com/KhronosGroup/KTX-Software/files/13455028/build.txt)

I've moved the docs & sdist/bdist builds into the `${KTX_BUILD_DIR}`. There is still a `build` directory generated for running the tests & docs. This is needed because the native module has to be compiled for the Python code to run. The `build` folder doesn't have any build artifacts.

ShukantPal:
> /Users/******/Library/Python/3.9/lib/python/site-packages/urllib3/__init__.py:34: NotOpenSSLWarning: urllib3 v2.0 only supports OpenSSL 1.1.1+, currently the 'ssl' module is compiled with 'LibreSSL 2.8.3'. See: https://github.com/urllib3/urllib3/issues/3020

This is a warning related to your Python installation. It is not something for KTX's build to fix.

MarkCallow:
Thanks for the rapid fixes.

There are still 2 of these warnings in the CI build log:

```
C:\Users\runneradmin\AppData\Local\Temp\build-env-38lihm5l\Lib\site-packages\wheel\bdist_wheel.py:109: RuntimeWarning: Config variable 'Py_DEBUG' is unset, Python ABI tag may be incorrect
```

The warnings I am seeing in my build are because I am building on arm64 (M2 Mac). Non of the CI builds build on arm64.
```
...libktx.4.3.0.dylib':  found architecture 'arm64', required architecture 'x86_64'
```

Obviously being on arm64 libktx was built for that. Something in the Python build is "requir(ing)" x86_64. Whether it is harmless or not I don't know. I will try building with the latest code and attempt to run the tests.

Thanks for the info regarding the SSL-related warning.

interface/python_binding/CMakeLists.txt
```diff
@@ -0,0 +1,163 @@
+# Copyright 2023 Shukant Pal
+# SPDX-License-Identifier: Apache-2.0
+
+find_package (Python3 COMPONENTS Interpreter)
+file(GLOB pyktx_py_src ${CMAKE_SOURCE_DIR}/interface/python_binding/pyktx/*.py)
+list(TRANSFORM pyktx_py_src REPLACE "${CMAKE_SOURCE_DIR}/interface/python_binding/pyktx/" "${KTX_BUILD_DIR}/interface/python_binding/docs/pyktx." OUTPUT_VARIABLE pyktx_py_rst_filenames)
+list(TRANSFORM pyktx_py_rst_filenames REPLACE ".py$" ".rst" OUTPUT_VARIABLE pyktx_py_rst)
+
+message(STATUS "pyktx_py_rst is ${pyktx_py_rst}")
```
MarkCallow, line 9: Do we need this message? It looks like a debug message to me.

MarkCallow:
I built the latest successfully. It does appear that building `all.doc` is necessary to get the documents in their final destination.

I tried to run the tests. As I could not find info in `BUILDING.md` or in `interface/python_binding/README.md` I just tried to run one of the test files from within the tests directory. You can see the result.

```
 python3 test_ktx_texture.py
Traceback (most recent call last):
  File "/Users/mark/Projects/khronos/github/KTX-Software/interface/python_binding/tests/test_ktx_texture.py", line 4, in <module>
    from pyktx import *
ModuleNotFoundError: No module named 'pyktx'
```

Please update `interface/python_binding/README.md` with information about building the binding and running the tests. If it is very long then consider making a separate `BUILDING.md`. Add a link to whichever it is from the top level `BUILDING.md`.

MarkCallow:
> The warnings I am seeing in my build are because I am building on arm64 (M2 Mac). None of the CI builds build on arm64.

Actually the Linux arm64 build on Travis (at line 116 in `.travis.yml`) runs on arm64 hardware but pyktx is not selected for build there. Please add it.

ShukantPal:
@MarkCallow I might know why you're seeing the warning from `ld`. Since you're using the Apple provided version of CPython, the CPython binary has both x86-64 and arm64 architectures. That's why `ld` attemps to create a universal binary; but that's not needed.

Screenshot 2023-11-24 at 10 28 31 PM

interface/python_binding/CMakeLists.txt
ShukantPal: Deleted, thanks.

ShukantPal:
> I tried to run the tests. As I could not find info in BUILDING.md or in interface/python_binding/README.md I just tried to run one of the test files from within the tests directory. You can see the result.

I updated the `README.md` to make it clear how to run the tests. If you already have libktx installed at its system location (e.g. `/usr/local/include`), simply switching `ON` the `KTX_RUN_TESTS` environment variable and running `buildscript.py` should do the trick.

Otherwise you would also have to set the `LIBKTX_INCLUDE_DIR` and `LIBKTX_LIB_DIR` environment variables (which is what the CMake script does).

ShukantPal:
> There are still 2 of these warnings in the Windows' CI build log:
>
> C:\Users\runneradmin\AppData\Local\Temp\build-env-38lihm5l\Lib\site-packages\wheel\bdist_wheel.py:109: > RuntimeWarning: Config variable 'Py_DEBUG' is unset, Python ABI tag may be incorrect


That too is a warning related to CPython's internals, not pyktx's code. Here is a relevant GitHub issue: https://github.com/pypa/wheel/issues/461

ShukantPal:
I've enabled the Python builds for Linux and also deleted the unneeded debug statements.

MarkCallow:
I followed your instructions for running the tests. Here is the log

```
 $ KTX_RUN_TESTS=true python3 buildscript.py
generating ./pyktx/native.c
(already up-to-date)
the current directory is '/Users/mark/Projects/khronos/github/KTX-Software/interface/python_binding'
ld: warning: ignoring file '/usr/local/lib/libktx.4.3.0.dylib': found architecture 'arm64', required architecture 'x86_64'
```

It does not appear to have run the tests. This is the warning you explained above. Your explanation makes sense but we need to figure out how to run the code. Or did the tests run and they are silent when they work?

interface/python_binding/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Building
+
+To build and test pyktx,
+
+```bash
+# Set LIBKTX_INSTALL_DIR if you've installed libktx at the default system location.
+# Otherwise set LIBKTX_INCLUDE_DIR, LIBKTX_LIB_DIR to wherever you've built libktx.
+cd ${PROJECT_DIR}/interface/python_binding
+KTX_RUN_TESTS=ON python3 buildscript.py
```
MarkCallow, line 18: What about just building? Presumably `python3 buildscript.py`. Are LIBKTX env vars necessary for building or just testing? Please clarify.

ShukantPal: 1. Yes excluding the run tests variable won't run tests.

2. KTX env vars are necessary for building, running , and testing UNLESS you've already installed libktx in its typical system location (eg /usr/local or Program Files on windows)

ShukantPal:
@MarkCallow You need to set the variable to ON not "true". See the README for exact command

interface/python_binding/README.md
ShukantPal: So basically if you the `cmake install` command on your machine, you can simply run

```bash
KTX_RUN_TESTS=ON python3 ./buildscript.py
```

otherwise you'd run

```bash
            LIBKTX_INCLUDE_DIR=${CMAKE_SOURCE_DIR}/include
            LIBKTX_LIB_DIR=${KTX_BUILD_DIR}/Debug
            KTX_RUN_TESTS=ON
            DYLD_LIBRARY_PATH=${KTX_BUILD_DIR}/Debug:$DYLD_LIBRARY_PATH
            LD_LIBRARY_PATH=${KTX_BUILD_DIR}/Debug:$LD_LIBRARY_PATH
        python3 buildscript.py
```

MarkCallow: Presumably you can use Release or Debug. Correct?

ShukantPal: That's right.

ShukantPal:
Seems like the Linux build is trying to build pyktx before the C++ build. Will look into it this weekend

MarkCallow:
> You need to set the variable to ON not "true". See the README for exact command

My bad. It works with "ON".

Still get the warning. Please add a note to the instructions about this warning with macOS Python to cut down on bug reports. Also mention pyktx in the main BUILDING.md with a link to the pyktx README.md.

MarkCallow:
> Seems like the Linux build is trying to build pyktx before the C++ build. Will look into it this weekend

You seem to be using `-L/home/travis/build/KhronosGroup/KTX-Software/build` to find the library. After a recent change in `main`, that you've merged, the library is now in `home/travis/build/KhronosGroup/KTX-Software/build/<config>` where config is `Release` or `Debug`. All binaries are now put in that directory instead of being spread around the various target build directories.

MarkCallow:
The first Linux build (Debug x86_64) has just completed successfully. It is now running the tests. More good news: the "FAILURE" message I reported is not in the log. That utility command must have needed the library too.

MarkCallow:
Assuming the rest of the CI jobs complete successfully the remaining item is the note I requested in the README about builds with macOS's (actually Xcode command line tools') Python3 giving the warning about no x86_64 library. The CI builds aren't using that Python, right?

The deployment is done only in one of the Windows CI jobs, right? We don't want to be deploying multiple times.

MarkCallow:
> the note I requested in the README

Something along the lines of

"When building on macOS using a Python3 installation with a universal CPython binary, such as that installed with the Xcode command-line tools (/usr/bin/python3), `ld` will issue a warning

```
ld: warning: ignoring file '/usr/local/lib/libktx.4.3.0.dylib': found architecture 'arm64', required architecture 'x86_64'
```

'arm64' and 'x86_64' may be reversed depending on the build machine architecture. This happens because libktx is not a universal binary so only supports the current platform architecture. The message can be ignored."

MarkCallow:
> the note I requested in the README

And the link from the main `BUILDING.md` to this README.md.

With those 2 items this PR will be ready to merge.

ShukantPal:
Thanks Mark, I will do so as soon as I get free time.

ShukantPal:
> The deployment is done only in one of the Windows CI jobs, right? We don't want to be deploying multiple times.

We "publish" artifacts from both GitHub Actions (Windows) and Travis CI (Linux+macOS). However, only the Windows CI uploads the source distribution & binary distributions. Travis CI uploads only the binary distributions ("wheels").

The `publish.yml` workflow then actually does the deployment. It's a separate job that downloads all the `pyktx` artifacts from a GitHub release and publishes them to PyPI (the package manager registry for Python).

If possible, I would recommend testing this workflow after the PR gets merged. (Remember, to publish pyktx you will need to add a `PYPI_API_TOKEN` secret in GitHub Actions corresponding to the PyPI account you sign up for)

interface/python_binding/README.md
```diff
@@ -24,4 +24,12 @@ cd ${PROJECT_DIR}/interface/python_binding
 KTX_RUN_TESTS=ON python3 buildscript.py
 ```
 
-If you are on a POSIX system (macOS or Linux), make sure libktx is on your `DYLD_LIBRARY_PATH` and `LD_LIBRARY_PATH`.
\ No newline at end of file
+If you are on a POSIX system (macOS or Linux), make sure libktx is on your `DYLD_LIBRARY_PATH` and `LD_LIBRARY_PATH`.
+
+> When building on macOS using against a universal CPython binary, such as that installed with the Xcode command-line tools (/usr/bin/python3), ld will issue a warning
```
MarkCallow, line 29: s/against //

ShukantPal:
```suggestion
> When building on macOS against a universal CPython binary, such as that installed with the Xcode command-line tools (/usr/bin/python3), ld will issue a warning
```

I think "against" makes more sense, since the ABI has to match the Python interpreter being used.

MarkCallow (approved):

MarkCallow:
Thank you very much @ShukantPal for your sterling work on this and for your perseverance over the last 6 montsh through all the requested additions and changes.

MarkCallow:
@ShukantPal I have built a draft release that includes Python. The following are included:

```
pyktx-4.3.0-cp310-cp310-linux_aarch64.whl 90.7 KB
pyktx-4.3.0-cp310-cp310-linux_x86_64.whl 91 KB
pyktx-4.3.0-cp311-cp311-macosx_12_0_x86_64.whl 46 KB
```
There is no Windows build. Should there be? It is not there because the deploy task in `.github/workflows/windows.yml` does not list any *.whl files. Please let me know as soon as possible. I want to finish the release.

I will try the publish step soon, possibly tomorrow.

I had build failures in configurations that do not include Python because the deployment section had the following:
```
ruby ci_scripts/github_release.rb -s ${GITHUB_TOKEN} -r ${TRAVIS_REPO_SLUG} -c $REL_DESC_FILE -t ${TRAVIS_TAG} --overwrite true --draft true --prerelease true $BUILD_DIR/KTX-Software-*-* $BUILD_DIR/interface/python_binding/dist/*.whl
```

When no wildcard is found bash passes the literal argument (`.../*.whl`) to the program. I add nullglob to stop this:
```
shopt -s nullglob; ruby ci_scripts/github_release.rb -s ${GITHUB_TOKEN} -r ${TRAVIS_REPO_SLUG} -c $REL_DESC_FILE -t ${TRAVIS_TAG} --overwrite true --draft true --prerelease true $BUILD_DIR/KTX-Software-*-* $BUILD_DIR/interface/python_binding/dist/*.whl
```

However it is still failing in the Emscripten build. I'll have to investigate further.

ShukantPal:
There should be wheels for Windows. I wonder if the output directory is correct -- I remember testing it on my Windows machine.

ShukantPal:
There should also be a tar.gz or .zip file

MarkCallow:
I am talking about the _Upload To Release_ part of `windows.yml`. It does not reference *.whl files at all. _Upload pyktx artifacts_ is uploading somewhere else.

```
    - name: Upload pyktx artifacts
      if: matrix.options.package == 'YES' && matrix.options.py
      uses: kittaakos/upload-artifact-as-is@v0
      with:
        path: ${{env.BUILD_DIR}}/interface/python_binding/dist/

    - name: Upload To Release
      uses: softprops/action-gh-release@v1
      if: matrix.options.package == 'YES' && matrix.toolset == 'CLangCL' && github.event_name == 'push' && github.ref_type == 'tag'
      with:
        draft: true
        prerelease: true
        files: ${{env.BUILD_DIR}}/KTX-Software-*.exe*
      env:
        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}

```

ShukantPal:
That dist/ folder is supposed to contain the wheels in the upload pyktx artifacts action

MarkCallow:
We need to replace kittaakos/upload-artifact-as-is because it uses node12 which is being removed from Actions. That action has not been touched form 3 years and has been archived so it is unlikely to be updated. Please find an alternate.

> That dist/ folder is supposed to contain the wheels in the upload pyktx artifacts action

As for why there is nothing in the dist directory I don't yet know. However this misses the point of my comment which is that there is nothing in Upload To Release to even attempt to upload pyktx.

MarkCallow:
There was stuff in the dist directory. The problem was the condition on the artifacts upload. I changed

```
      if: matrix.options.package == 'YES' && matrix.options.py
```
to
```
     if: matrix.options.package == 'YES' && matrix.options.py == 'ON'
```

and the warning that no packages were found went away as now the action rightly only happens when pyktx is enabled.

I've also added deployment of the *.whl files to the Upload to Release action. I'm running a test build now.

I just found another issue. publish.yml is run immediately a tag is pushed so long before the other builds have completed and deployed the assets to the release. Even if all builds were done in GitHub Actions I do not know how to make it wait until all other actions have completed. Do you know? As Linux and macOS builds are done on another provider there is no chance for it to tell if the release is complete. I think publish will always have to be done manually so we should remove the "on: push" from publish.yml.

I'm in the process of getting an API token so I can test publishing. Do they have some equivalent of draft so I can check what got published before I make it public?

ShukantPal:
I would suggest running `publish.yml` manually then, and removing the `on: push` part.

You can make the `publish.yml` workflow accept an input to publish to https://test.pypi.org instead. That is a test version of PyPI, and then you can inspect what is published.

MarkCallow:
Deployment of the *.whl files from Windows is now working. I have in the draft release

```
 pyktx-4.3.0-cp311-cp311-win_amd64.whl 
```

There are no .whl files for macOS or Windows arm64 because, as I understand it, pyktx can't be cross-compiled. Does it matter that these are missing and won't be deployed to PyPI? (What a confusing name when written in a sans-serif font.)

Once I've got an API token I'll work on publish.yml. How do you make the workflow accept an input and how to you provide that input?

MarkCallow:
`pyktx-4.3.0-cp311-cp311-win_amd64.whl `

What does the `cp311` mean and why is it repeated in the file name?

ShukantPal:
> What does the cp311 mean and why is it repeated in the file name?

CPython 3.11

I don't know why it's repeated, but it is auto generated by the build tooling.

ShukantPal:
> Does it matter that these are missing and won't be deployed to PyPI?

No, as long as the source distribution (tar or zip from the build) is provided -- the pip package manager will build the wheel at installation time.

ShukantPal:
> How do you make the workflow accept an input and how to you provide that input?

I can provide an example later (I'm out eating dinner right now). When you run the workflow in GitHub, there will be a drop-down or text field depending on how you declare the input.

MarkCallow:
> I can provide an example later 

Thank you. I look forward to seeing it.

> No, as long as the source distribution (tar or zip from the build) is provided -- the pip package manager will build the wheel at installation time.

No tar or zip is being deployed to the GitHub Release only .whl files. Therefore publish won't upload anything to PyPI. I guess I'll have to modify the deployments to include at least one source .tar. Does it matter which platform build it is from?

ShukantPal:
> Does it matter which platform build it is from?

No

MarkCallow:
Don't forget the example workflow.

MarkCallow:
Hi @ShukantPal, I Hope you have a great holiday season and I belated wish you Happy Birthday. Thanks for the example workflow.  I have some related questions:

1. How do I tell `gh-action-pypi-publish` to publish to test.pypi.org instead of pypi.org.
2. If published to test.pypi.org, once tested how do I transition it to pypi.org?

Khronos is working on setting up an organization account on pypi.org. Currently we are awaiting a response from pypi.org.

And I have more questions:

1. How do I build the distribution. The deployment sections you added to the .yml files deploy `interface/python/binding/dist` but when I build pyktx on my Mac either by building the target in Xcode or manually invoking `buildscript.py` no dist directory is created. I looked in `README.md` but it did not help.
2. The reason I was trying to create/find `dist` is so I could look at the tar file and get its name so as to add something to one of the .yml files to deploy it. What is its name?

MarkCallow:
I figured out the tar file name and now have one source tar deployed into the draft release.

I have just now realized that the version number used in the pyktx files is missing the modifier, `-beta1` in the current case. All the other packages have it. Please fix the build so it is included. You can submit a PR and I can cherry-pick the fix into the branch where I'm working on the release.

MarkCallow:
@ShukantPal, This is now on the critical path to getting the beta release out so please answer my questions quickly. I just tried to build the package target but still no dist directory.

MarkCallow:
I was looking in the source directory not the build directory for the `dist` directory. Silly me. I still need to know what script builds the packages so I can fix the version number. I'd prefer if you could provide the fix but as you are on holiday I'll understand if you cannot.

MarkCallow:
I now see where the version is being set. I need your advice. My concern is if I upload the beta packages labelled as 4.3.0 to PyPI then when I later overwrite them with true 4.3.0 release packages how would pip users be able to upgrade to the new packages. Won't pip and PyPI consider them the same? Therefore I think we need to include the tweak, `-beta1`, in the package name. If we do so, would someone wishing to install the wrapper have to use `4.3.0-beta1` in the pip command or will `4.3` work.

ShukantPal:
> How do I tell gh-action-pypi-publish to publish to test.pypi.org instead of pypi.org.

You can use [expression syntax](https://docs.github.com/en/actions/learn-github-actions/expressions) for this: `${{ inputs.test ? ‘<TEST_REGISTRY_URL>’ : ‘<REGISTRY_URL’ }}`

> If published to test.pypi.org, once tested how do I transition it to pypi.org?

You need to run the publish workflow again, but this time without toggling the test input to true.

> How do I build the distribution. The deployment sections you added to the .yml files deploy interface/python/binding/dist but when I build pyktx on my Mac either by building the target in Xcode or manually invoking buildscript.py no dist directory is created. I looked in README.md but it did not help.

Make sure you are running CMake with `KTX_FEATURE_PY=ON`, and then build the `pyktx` target.

> The reason I was trying to create/find dist is so I could look at the tar file and get its name so as to add something to one of the .yml files to deploy it. What is its name?

I am not sure, but I would suggest using a glob - `*.tar.gz`. The `.zip` if you are building on Windows.

>  Won't pip and PyPI consider them the same? Therefore I think we need to include the tweak, -beta1, in the package name. If we do so, would someone wishing to install the wrapper have to use 4.3.0-beta1 in the pip command or will 4.3 work.

Please read about [PyPI versioning schemes](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#choosing-a-versioning-scheme)

You can append `b1`,`b2`,… after the version number to specify a beta release.

MarkCallow:
Thanks for your answers.

> You can use [expression syntax](https://docs.github.com/en/actions/learn-github-actions/expressions) for this: ${{ inputs.test ? ‘<TEST_REGISTRY_URL>’ : ‘<REGISTRY_URL’ }}

This did not answer my question. Where would I do this check of `inputs.test`? No URL is currently being passed to gh-action-pypi-publish. Does it have an optional input for the URL?

> You need to run the publish workflow again, but this time without toggling the test input to true.

How do I remove the packages from test once published to the real registry?

I'll read about the version schemes.

MarkCallow:
I changed the byproducts and LIBKTX_VERSION setting in the custom commands to use LIBKTX_VERSION_FULL. The build now produces files with version `4.3.0b1` which is `4.3.0-beta1` normalized per PyPI rules. The tests ran successfully. I've just started a CI build with this change.

MarkCallow:
>  I changed the byproducts ... to use LIBKTX_VERSION_FULL.

Since the actual byproducts use the normalized name as mentioned, I expect they will be built every time a build is run as the CMake custom command will be looking for the unnormalized name. Is there a way to access the normalization function in the Python build system to use it to normalize the byproduct names given to CMake?

ShukantPal:
> Is there a way to access the normalization function in the Python build system to use it to normalize the byproduct names given to CMake?

You may want to do the normalization within CMake using string operations like `replace`.

https://cmake.org/cmake/help/latest/command/string.html#replace

Then, pass that into the Python build directly and use it to find the output.

ShukantPal:
> This did not answer my question. Where would I do this check of inputs.test? No URL is currently being passed to gh-action-pypi-publish. Does it have an optional input for the URL?

You can use the following snippet:
```
with:
    repository-url: https://test.pypi.org/legacy/
```
here https://github.com/KhronosGroup/KTX-Software/blob/d3ef5ed8b4d443804fc07b348cc503b30f7da938/.github/workflows/publish.yml#L37

MarkCallow:
@ShukantPal you have a `pyktx` project on test.pypi.org. How can I take it over and publish to it?

ShukantPal:
@MarkCallow I have deleted my published pyktx project on Test PyPi. You should be able to publish it with your account now.

MarkCallow:
@ShukantPal,  I'm still waiting for various permissions and account set up for publishing to PyPI. I've gone ahead and made v4.3.0-beta1 release without it. Will fix later. The pyktx assets are deployed in Releases for download. The pyktx docs are deployed at https://github.khronos.org/KTX-Software/pyktx/index.html.

But there is a problem. Anything in the various directories named with `_` prefixes are not found. Apparently this is something to do with GitHub's use of Jekyll:  https://github.blog/2009-12-29-bypassing-jekyll-on-github-pages/. Is it possible to name these directories differently when generating the docs? If not, I'll have to figure out how to create a `.nojekyll` file in the built docs root. At present none of the scripts or css is found as they're all in _static.

MarkCallow:
I'm trying to get publishing working. I made changes to the action file to ready it for using trusted publishing and renamed it `publish-pyktx.yml`. Unfortunately for bureaucratic reasons we can't get things set up for trusted publishing yet so I've changed it to use an API token. I am having great difficulty getting the action to recognize the test-pypi input. I keep getting a startup error:

```
 Invalid workflow file: .github/workflows/publish-pyktx.yml#L60
The workflow is not valid. .github/workflows/publish-pyktx.yml (Line: 60, Col: 9): Unrecognized named-value: 'test-pypi'. Located at position 1 within expression: test-pypi == false .github/workflows/publish-pyktx.yml (Line: 95, Col: 9): Unrecognized named-value: 'test-pypi'. Located at position 1 within expression: test-pypi

```
What is wrong? Look at .github/workflows/publish-pyktx.yml` in branch `test_py_publish`.

The error message has not changed since my first attempt where I used `if: test-pypi`. It is now '`if: ${{ github.event.inputs.test-pypi }}` yet the error message has not changed. In between I had `if: ${{ inputs.test-pypi }}` and got the identical error message.

I will not be able to work on this for nearly a month. I am very disappointed that I could not get this working before going on my break. There is no rush but please investigate while this message is still in your mind.

MarkCallow:
> What is wrong? Look at .github/workflows/publish-pyktx.ymlin branchtest_py_publish`.

I figured it out. I was not selecting the correct branch when I was dispatching the action. Stupid.

How it gets all the way to attempting to deploy but I get 

```
zipfile.BadZipFile: File is not a zip file
Checking dist/pyktx-4.3.0b1-cp310-cp310-linux_aarch64.whl: 
```

Please figure out why the .whl file is not apparently a zip file. See [here](https://github.com/KhronosGroup/KTX-Software/actions/runs/7498674379/job/20414169920)  for the build log with the python traceback.

ShukantPal:
https://github.com/pypa/gh-action-pypi-publish/issues/148

MarkCallow:
I resolved the 'not a zip file' issue. The curl command for downloading the assets did not enable redirects so the downloaded files all had zero bytes. I now have the following issues which appear to be with the packages we are creating. Please fix.

```
Checking dist/pyktx-4.3.0b1-cp310-cp310-linux_aarch64.whl: FAILED
ERROR    `long_description` has syntax errors in markup and would not be        
         rendered on PyPI.                                                      
         No content rendered from RST source.                                   
WARNING  `long_description_content_type` missing. defaulting to `text/x-rst`.   
Checking dist/pyktx-4.3.0b1-cp310-cp310-linux_x86_64.whl: FAILED
ERROR    `long_description` has syntax errors in markup and would not be        
         rendered on PyPI.                                                      
         No content rendered from RST source.                                   
WARNING  `long_description_content_type` missing. defaulting to `text/x-rst`.   
Checking dist/pyktx-4.3.0b1-cp311-cp311-macosx_12_0_x86_64.whl: FAILED
ERROR    `long_description` has syntax errors in markup and would not be        
         rendered on PyPI.                                                      
         No content rendered from RST source.                                   
WARNING  `long_description_content_type` missing. defaulting to `text/x-rst`.   
Checking dist/pyktx-4.3.0b1-cp311-cp311-win_amd64.whl: FAILED
ERROR    `long_description` has syntax errors in markup and would not be        
         rendered on PyPI.                                                      
         No content rendered from RST source.                                   
WARNING  `long_description_content_type` missing. defaulting to `text/x-rst`.   
Checking dist/pyktx-4.3.0b1.tar.gz: PASSED with warnings
WARNING  `long_description_content_type` missing. defaulting to `text/x-rst`.   
WARNING  `long_description` missing.                                            
```

MarkCallow:
@ShukantPal please submit your fixes as a PR against the `test_py_publish` branch.

I have created scripts to retrieve the release metadata and download assets I think that functionality may be more generally useful. The drawback is that publish_pyktx now has to check out the repo. The scripts and updated .yml are now in the branch. I deleted most of the older logs. Look at this [current log](https://github.com/KhronosGroup/KTX-Software/actions/runs/7523673990/job/20477376892) to see the errors shown in my previous comment.

MarkCallow:
@ShukantPal this is now on the critical path to releasing 4.3.0. Please provide fixes for the descriptions as soon as possible.

PR #701, opened 2023-05-12
Fix normalization when the result overflows

wasimabbas-arm:
Using uint for max values now because uint Max32 can't be represented in IEEE754 single precision float. In calculations it will be promoted to float. Changed normalization function for single channel images to return max of the data type assuming (x, 0, 0, 0) vectors. And changed the test values accordingly


wasimabbas-arm:
@MarkCallow Have a look at this fix and let me know what you think.

MarkCallow:
Thanks @wasimabbas-arm.

PR #702, opened 2023-05-15
Reimplement image input handling for toktx.

MarkCallow:
* Adds imageio library.
* Reworks toktx to use the imageio library and break up the
  gigantic main loop.
* Adds {fmt] library used by imageio.

The output plugins and the EXR input plugin are empty shells.

All toktx tests pass. Some golden images are modified due to
toktx now doing a transfer function conversion from BT709
to sRGB when reading npbm files. Some tests are modified because
toktx no longer chooses sRGB or Linear when the gamma is
somewhere between them.


MarkCallow:
This PR shows as having a failed build because the VS2017 build did not work. That has been removed from CI but GitHub failed to notice the removal and continues to show the failed build.

PR #704, opened 2023-05-18
Only deploy on version tags.

MarkCallow:


MarkCallow:
No idea why GH is still trying an Appveyor build. There is no more .appveyor.yml. Merging despite these "failures".

Issue #720, opened 2023-06-19
Compressor leaks native memory when using Java interface on Windows/arm64 or Windows/x64


robnugent:
Take the following test program and run it on Windows/arm64 or Windows/x64. Within a small number of minutes the JVM will crash. The HotSpot dump file shows e.g. the following as the stack trace:

```
Stack: [0x000000ee1c400000,0x000000ee1c500000],  sp=0x000000ee1c4fe910,  free space=1018k
Native frames: (J=compiled Java code, j=interpreted, Vv=VM code, C=native code)
C  [KERNELBASE.dll+0x6536c]
C  [VCRUNTIME140.dll+0x6480]
C  [ktx.dll+0x161e13]

Java frames: (J=compiled Java code, j=interpreted, Vv=VM code)
J 964  org.khronos.ktx.KtxTexture2.compressAstcEx(Lorg/khronos/ktx/KtxAstcParams;)I libktx@4.2.0 (0 bytes) @ 0x000002e187c0d9f7 [0x000002e187c0d9a0+0x0000000000000057]
J 1072 c2 uk.co.abraded.toktx.ToKTXBug.convertToASTC(II)I uk.co.abraded.toktx (132 bytes) @ 0x000002e187c2764c [0x000002e187c27280+0x00000000000003cc]
j  uk.co.abraded.toktx.ToKTXBug.run()V+61 uk.co.abraded.toktx
j  java.lang.Thread.run()V+11 java.base@17.0.1
v  ~StubRoutines::call_stub

siginfo: EXCEPTION_UNCAUGHT_CXX_EXCEPTION (0xe06d7363), ExceptionInformation=0x0000000019930520 0x000000ee1c4fea70 0x00007ffdae248648 0x00007ffdae040000 
```

I'll try to append the whole dump file

Sometimes, there is a message to stderr saying something like 'malloc failed'.

Note the following:

1) Windows Task Manager show 'committed' memory spiralling during this test until the point of failure
2) It fails even if you adjust the test to just use one thread - it just takes longer.

```
package uk.co.abraded.toktx;

import java.text.DecimalFormat;
import java.util.ArrayList;
import java.util.List;
import java.util.Random;
import org.khronos.ktx.KtxAstcParams;
import org.khronos.ktx.KtxCreateStorage;
import org.khronos.ktx.KtxErrorCode;
import org.khronos.ktx.KtxPackAstcBlockDimension;
[...]

PR #721, opened 2023-06-20
Fix memory leak of input_image in ktxTexture2_CompressAstcEx

wasimabbas-arm:
Fixes the memory leak identified in #720.


MarkCallow:
Thanks Wasim.

PR #724, opened 2023-07-03
WIP: First commit of encode astc support

wasimabbas-arm:
Adding astc support to encode.

### DIFF
--- a/tools/ktx/command_encode.cpp
+++ b/tools/ktx/command_encode.cpp
@@ -111,6 +111,9 @@ int CommandEncode::main(int argc, _TCHAR* argv[]) {
 
 void CommandEncode::OptionsEncode::init(cxxopts::Options& opts) {
     opts.add_options()
+		("format", "KTX format enum. The enum names are matching the VkFormats without the VK_FORMAT_ prefix."
+		          " If the format is an ASTC formats the ASTC encoder specific options become valid,"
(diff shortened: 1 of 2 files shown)


MarkCallow:
Sorry for leaving you hanging @wasimabbas-arm. After our discussions and seeing this WIP proposal I feel it is not a good fit with the CLI design for `encode`. I think it is probably better to create a new tool to apply ASTC compression to an existing uncompressed .ktx2 file. We are discussing within Khronos. I'll let you know.

MarkCallow:
@wasimabbas-arm after becoming more familiar with the new tool set and the options across various tools I am still of the opinion that this is not a good fit for the encode tool. Please create a new tool for encoding to block-compressed GPU formats. The initial version will support only ASTC. Specify the format with a `--format` option like `ktx create`. It should only accept supported block-compressed VkFormat enum values. The tool can place limits on the acceptable format of the input .ktx2 file.

I am struggling for a name for this tool. The best I've come up so far is `encode-gpu`. I don't like the hyphen though and `gpuencode` is too difficult to parse for quick understanding. Perhaps we should have a separate tool for each GPU format supported. If it is only ASTC and BC that is not so bad (`encode-astc` and `encode-bc`). If there are many others then I prefer a single tool. Probably we should rename `encode` to `encode-basis` to match.

Let me know what you think.

I'm going to close this PR.

wasimabbas-arm:
@MarkCallow I agree. As far as naming goes, I think it should be `encode-astc`, `encode-basis` and `encode-anythingelse` that comes in the future.

PR #731, opened 2023-07-06
Update KTX-Software-CTS reference to specific commit

aqnuep:
This PR only updates the submodule reference for the KTX-Software-CTS repository, as it should always refer to a specific commit to avoid new tests causing failures on old code.

@MarkCallow, could you please merge this in? Apparently this was overlooked when merging the KTX-Tools branch into main.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -5,4 +5,4 @@
 [submodule "tests/cts"]
 	path = tests/cts
 	url = https://github.com/KhronosGroup/KTX-Software-CTS.git
-	branch = main
+	branch = b881ca6ff438b663b98952d3361e5e3fce14938c


VaderY (approved):

VaderY (changes requested):
I think this is fine as is because this is not the submodule ref and it refers to a different "branch" (the default remote branch, rarely used).

MarkCallow:
> I think this is fine as is because this is not the submodule ref and it refers to a different "branch" (the default remote branch, rarely used).

What is fine as is, this PR? What refers to the default remote branch, the commit hash in this update?

In general should KTX-Software's `main` point to the submodule's `main`? Then when a tag is added we need to ensure the submodule reference at that tag points at a specific commit?

How should we manage this from now on?

MarkCallow (commented):
Please see my comment in the main conversation.

VaderY:
@MarkCallow By fine I mean I dont think there is an error in how we managed submodules so far, so this PR is not needed.

> In general should KTX-Software's `main` point to the submodule's `main`? Then when a tag is added we need to ensure the submodule reference at that tag points at a specific commit?

git submodule is managed on a per commit basis wish SHA. Each KTX commit references a CTS commit for it. So no need to do any extra work or synchronization beyond the normal development workflow

aqnuep:
Sorry, my bad. Indeed, @VaderY is right, that the branch specified in .gitsubmodule won't matter, as it is only used when you explicitly request the latest changes from the tracked branch to be pulled manually, hence this PR can be dropped.

Issue #733, opened 2023-07-08
Problem with convertion files (v4.3.0-alpha3)


Yason83:
While performing file conversion (files attached in the [archive](https://github.com/KhronosGroup/KTX-Software/files/11992583/images.zip)), the command line
`ktx create --format R8G8B8A8_SRGB --encode basis-lz --clevel 5 --qlevel 255 baseColor_22.png baseColor_22.ktx`

Problems
1. Fails to convert the files. 
2. The error message is not displayed.

MarkCallow:
The example command says `baseColor_22.png` but there is no such file in the zip file. I tried with `baseColor_18.png`. It worked. Here is the result:

<img width="1280" alt="Screenshot 2023-07-09 at 21 24 51" src="https://github.com/KhronosGroup/KTX-Software/assets/2244683/b1048bb4-388f-45bc-8fa3-da3fab9da4ba">

You need to provide a lot more information about the problem. What do you mean "Fails to convert the files?"

Yason83:
I tested on Windows11 64 (v4.3.0-alpha3)
After running the script, the result is not saved
`ktx create --format R8G8B8_SRGB --encode basis-lz --clevel 5 --qlevel 255 baseColor_18.png baseColor_18.ktx2`

![48](https://github.com/KhronosGroup/KTX-Software/assets/32398829/bc2683e4-c794-48fa-9505-87993c970052)

MarkCallow:
It appears to be a Windows-specific problem. I was previously trying on macOS/arm64. When I try on Windows/arm64 I get no output file and no errors as reported in this issue. I also get no warning messages from `create`. I didn't get any on macOS either. I'm specifying `R8G8B8A8_SRGB` format with `baseColor_18.png` as input. 

In the debugger, I'm getting an access violation exception at line 815 in image.hpp. Two questions: why the violation and why does the app silently exit. @VaderY please investigate.

VaderY:
Successfully reproduced it. Will look into it ASAP.

VaderY:
The issue was typo that caused miss addressing. I will add additional tests to cover these and post a PR later today with the fix.

PR #735, opened 2023-07-10
Fix incorrect index calculations in image conversions in ktx create

VaderY:
Fix incorrect index calculations in image conversions in ktx create (#733, #736)

### DIFF
--- a/tools/imageio/exr.imageio/exrinput.cc
+++ b/tools/imageio/exr.imageio/exrinput.cc
@@ -259,12 +259,12 @@ void ExrInput::readImage(void* outputBuffer, size_t bufferByteCount,
     // Copy the data
     const auto copyData = [&](unsigned char* ptr, uint32_t dataSize) {
         const auto sourcePtr = [&](uint32_t channel, uint32_t x, uint32_t y) {
-            return reinterpret_cast<const unsigned char*>(image.images[channel] + (y * height + x) * dataSize);
+            return reinterpret_cast<const unsigned char*>(image.images[channel] + (y * width + x) * dataSize);
         };
 
         for (uint32_t y = 0; y < height; ++y) {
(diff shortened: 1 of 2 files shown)


MarkCallow (approved):

MarkCallow:
@aqnuep please review.

MarkCallow:
Why did the user not receive any notice of the crash?

And, less importantly, why did it produce an output file on macOS?

aqnuep:
We discussed this change with @VaderY. It was a trivial oversight where row stride was incorrectly derived from the height instead of the width.

I'm not sure about the different behaviors across platforms, but I'd expect this typo would mainly only cause data corruption for non-square textures due to the incorrect row stride.

aqnuep (approved):
LGTM.

aqnuep:
FYI, @MarkCallow, @VaderY also added a bunch of odd sized textures to the CTS to have better coverage on this front.

We apologize for the typo, but it's been a huge set of functionalities to cover with implementation and tests, and obviously this slipped through the cracks.

VaderY:
> Why did the user not receive any notice of the crash?

Technically the exit code was an indicator (it was 0xC0000005 = Access Violation). But as it was a bug and not an error that we should ever handle, there was no reasonable way to produce an error message about it.
There are ways with C++ to intercept such crashes (breakpad, crashbed, signals, etc) and attempt to collect more information, write out coredumps and notify the user. But it is not a trivial undertaking to implement or adopt these in a cross platform manner.
 
> And, less importantly, why did it produce an output file on macOS?

Given some bad luck these kind of undefined behaviors are sometimes hidden on certain platform. When I checked in windows builds every compiler yielded a proper crash. Its possible that the application owned the memory pages that was written to and there was nothing important.

PR #743, opened 2023-07-22
Remove unused variable 's' and 'e' in ktxUpgrader::main

FuXiii:
When use `MinGW` to compile, it will output error:

```Cmd
E:\TestDelete\WebAssemblyTest\KTX-Software\tools\ktx2ktx2\ktx2ktx2.cpp:190:47: error: unused variable '
    std::vector<_tstring>::const_iterator it, s, e;
                                              ^
E:\TestDelete\WebAssemblyTest\KTX-Software\tools\ktx2ktx2\ktx2ktx2.cpp:190:50: error: unused variable '
    std::vector<_tstring>::const_iterator it, s, e;
                                                 ^
```

then remove unused variable `s` and `e`. It will compile succes.

### DIFF
--- a/tools/ktx2ktx2/ktx2ktx2.cpp
+++ b/tools/ktx2ktx2/ktx2ktx2.cpp
@@ -187,7 +187,7 @@ ktxUpgrader::main(int argc, _TCHAR* argv[])
     processCommandLine(argc, argv);
     validateOptions();
 
-    std::vector<_tstring>::const_iterator it, s, e;
+    std::vector<_tstring>::const_iterator it;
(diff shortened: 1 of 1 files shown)


FuXiii:
If I set `CMAKE_CXX_FLAGS` with `-Wno-unused-variable` it will compile success. 😂
This shouldn't be a problem, at most it can be considered compiler option settings.

MarkCallow:
Thanks for this. The warning is new in the GCC version in your MinGW. It also appears with the latest Xcode/Clang.
I intend to fix this warning as you did and also create a CMake config option for turning on warnings as errors which we will set in CI. I'll provide a different PR in a few days.

Issue #567, opened 2022-05-01
How to combine with VulkanMemoryAllocator


rHermes:
Hey!

I just wanted to ask how this might be combined with something like: https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator

As far as I can see, there is no option in `ktxTexture_VkUploadEx()` to specify how the image is allocated. I would like to use the VMA functions for this. How is this inteded to be achieved?

Best regards,
Teodor Spæren

MarkCallow:
When you create the VulkanDeviceInfo struct (with`ktxVulkanDeviceInfo_{Create,Construct}`), that is used to pass device info to `ktxVkUploadEx`, you can pass a pointer to a `VkAllocationCallbacks` struct. This will be used by any allocation done by `ktxVkUploadEx`.

 I'm not familiar with VulkanMemoryAllocator, which long postdates `ktxVkUploadEx`, but as `VkAllocationCallbacks` is the Vulkan API interface for providing external allocators, I expect the library must have such a structure that you can obtain and pass to `ktxVulkanDeviceInfo_{Create,Construct}`.

rHermes:
Thanks for the quick reply!

I did indeed look at the `VkAllocationCallbacks` struct, but as far as I understand, this is only for host only memory, things like data structs and so forth, and not for allocating `VkDeviceMemory` which is what VMA deals with. What I'm currently looking at is to just reimplement the entire logic of the current ktxTexture_VkUploadEx(), but replace the allocation calls with the VMA equivalent, but I was wondering if there was a way to avoid doing this.

MarkCallow:
Indeed `VkAllocationCallbacks` appears to be for host memory.

I don't think there is any alternative than reimplementing `ktxTexture_VkUploadEx()`. Currently it calls `vkAllocateMemory` to obtain device memory. Perhaps you can add device memory allocation pointers to `ktxVulkanDeviceInfo` which default to `vkAllocateMemory` and `vkFreeMemory` and use those pointers for device memory allocation in `ktxTexture_VkUploadEx()`.

I would accept a PR to add VMA support to the `libktx` provided `ktxTexture_VkUploadEx()` can still be used as today without VMA and there is no significant performance loss.

rHermes:
Hey! The mentioned PR doesn't really resolve this situation. VulkanMemoryAllocator has additional structures that is needed. We will need to provide a more generic solution for this, or even implement a specific version for VMA.

[...]

PR #748, opened 2023-08-08
Support for external allocators:

toomuchvoltage:
Support for external allocators:
* The newly introduced API surface area matches that of VMA's advanced usage and another hand-rolled memory allocator in a content-heavy application.
* All suballocator callbacks -- allocate, bind image, bind buffer, map, unmap and free -- are expected to guard VkDeviceMemory operations within a mutex.
* ~~Each texture now also keeps track of its VkDeviceMemory offset.~~ Potential sparse bindings support removed this.
* The 64 bit allocationId is to be used as a book-keeping measure by external suballocator callbacks to keep track of and free up suballocations. The external allocator can use a hashtable (ala std::unordered_map in C++) to keep track of the page(s) alloted to this suballocation. ('Pages' here refers to potential sparse bindings).


toomuchvoltage:
I cannot tell you how much better everything works now in my engine.

Memory fragmentation is _waaaay_ down and it is harder to run out of memory with texture stream-in/out events.

If you need me to clean it up and de-duplicate code: with pleasure. But I cannot stress how much more efficient this is as opposed to constant VkDeviceMemory creation/deletions. Not to mention, this should also save you from the 4K limit on Windows drivers for live VkDeviceMemory objects.

MarkCallow:
Thank you for this. Does it resolve the remaining parts of #567?

Please de-duplicate the code. I suggest making the new function work with a null sub-allocator and then calling it from the existing `VkUploadEx`. Please make typedefs for the various function pointer types passed to the new functions.

I need code to test this. At a minimum please add a new sample to `vkloadtests` or modify an existing sample to add a command line option to tell it to use a sub-allocator. Add a line to call the new test to the list at the end of `VulkanLoadTests.cpp`.

The TravisCI builds (macOS and Linux) didn't trigger. Hopefully it is just that the account needed topping up and your next commit will trigger them.

MarkCallow:
Ping @toomuchvoltage.

toomuchvoltage:
> Ping @toomuchvoltage.

Hi @MarkCallow , apologies for the delay, have been very busy.  I will get back to you shortly with some updates.

toomuchvoltage:
Hi @MarkCallow , so the most recent force-push basically addresses all of #567 . While considering your suggestion I still realized that a new API surface area may be necessary: the reason for this was that the number of function pointers increased to six. Alloc and free are not enough: bind image, bind buffer, map and unmap also need to be guarded with the same mutex as alloc and free as otherwise multi-threaded Vulkan rendering will be very unhappy. This is a consideration that VMA's advanced usage _also_ considers. I can still go back and modify `UploadEx`'s signature, but it would balloon up by a lot and if one is satisfied not relying on a suballocator, they would have to pass 6 new NULL pointers as parameters.

Speaking of VMA, it exposes these more granular calls for advanced usage, all of which perform mutually exclusive VkDeviceMemory operations:

- vmaAllocateMemory() or vmaAllocateMemoryPages()
- vmaBindBufferMemory()
- vmaBindImageMemory()
- vmaMapMemory()
- vmaUnmapMemory()
- vmaFreeMemory() or vmaFreeMemoryPages()

The six new callbacks provided to an `UploadEx_WithPotentialSuballocator()` call are supposed to provide callbacks that wrap around these. The allocation callback should also return a `uint64_t` that references the page procurement(s) rather than `VmaAllocation`(s). Such a 64 bit number could be generated by a mersenne twister PRNG and be a key into a hashmap (a.l.a std::unordered_map) that references the suballocations as mentioned previously.

An example of how this is done for a hand-rolled suballocator is provided here: https://github.com/toomuchvoltage/HighOmega-public/blob/sauray_vkquake2/HighOmega/src/gl.cpp#L288-L337 . All supplementary memory management code is contained at the top of that module.
Thread-safe allocation is demo'd here: https://github.com/toomuchvoltage/HighOmega-public/blob/sauray_vkquake2/HighOmega/src/gl.cpp#L288-L337
with the corresponding deallocation shown here: https://github.com/toomuchvoltage/HighOmega-public/blob/sauray_vkquake2/HighOmega/src/gl.cpp#L3164

I will gladly proceed to add a test, but just make sure: would it be OK for the test to have a dependency on VMA? There is no better way to genuinely test without making that explicit functional dependency.

toomuchvoltage:
Hi @MarkCallow , I just made another improvement to reduce the API's surface area. All suballocator callbacks are now packed into a single struct. An optional pointer to that struct can be passed reducing `UploadEx` suballocator parameters down to 1.

Updated thread-safe allocation usage: https://github.com/toomuchvoltage/HighOmega-public/blob/sauray_vkquake2/HighOmega/src/gl.cpp#L3494-L3503
Updated thread-safe free usage: https://github.com/toomuchvoltage/HighOmega-public/blob/sauray_vkquake2/HighOmega/src/gl.cpp#L3164-L3171

toomuchvoltage:
Actually, @MarkCallow let me know if the signature changes to `UploadEx` are ok. If so, I will proceed with updating the existing (broken) tests alongside adding new tests.

MarkCallow:
I like having the sub-allocation functions in a structure and the typedefs.

The signature changes to `ktxTexture_VkUploadEx` and `ktxVulkanTexture_Destruct` are not okay as they can break existing applications, as you've seen with the existing tests. The functions with the suballoc struct pointer parameter will have to have new names. These new functions can be called by the functions with the existing signatures.

The new tests having a dependency on VMA is fine provided that only those tests have the dependency and the rest of the KTX-Software project can be built without it. How big is VMA? Is it small enough to include in the KTX-Software repo or should we require people to download it?

toomuchvoltage:
Hi @MarkCallow , just made sure that the old function signatures remain as is. Would appreciate a quick check to ensure that my specifications are also in line with our doxygen rules. I also managed to make yet another improvement as well: I ensured that the interface can take in (hopefully thread-safe) suballocator callbacks that have the potential to do sparse bindings.

Resultingly, the suballocator callbacks no longer return a `VkDeviceMemory` along with an offset back as there may be multiple of these pages per a single procurement. Of course, `UploadEx`'s `memcpy` code can right now only handle a single allocation (i.e. non-sparse) -- and hence the `assert(numPages == 1)` in the code -- but this leaves the room for future enhancements of `UploadEx` without interface breaking changes in the event that sparse bindings' uploads are implemented.

I also made accesses to the suballocator directory in my examples thread-safe as well. Here are my handrolled thread-safe suballocator callbacks that have potential sparse bindings support: https://github.com/toomuchvoltage/HighOmega-public/blob/sauray_vkquake2/HighOmega/src/gl.cpp#L288-L411
Allocation usage is provided here: https://github.com/toomuchvoltage/HighOmega-public/blob/sauray_vkquake2/HighOmega/src/gl.cpp#L3568-L3577
Deallocation usage is provided here: https://github.com/toomuchvoltage/HighOmega-public/blob/sauray_vkquake2/HighOmega/src/gl.cpp#L3238-L3245

toomuchvoltage:
Oh I forgot to mention, VMA is a single header library... but it is written in C++ while providing a C interface as well. So _if_ the test has to be written in C, a library -- possibly a library containing thread-safe callbacks just like mine -- has to be written around it to be included in the test sample.

lib/vkloader.c
```diff
@@ -749,20 +749,29 @@ linearTilingPadCallback(int miplevel, int face,
...
+ * @param [in,out] vkTexture              pointer to a ktxVulkanTexture structure into which
+ *                                        the function writes information about the created
+ *                                        VkImage.
+ * @param [in] tiling                     type of tiling to use in the destination image
+ *                                        on the Vulkan device.
+ * @param [in] usageFlags                 a set of VkImageUsageFlags bits indicating the
+ *                                        intended usage of the destination image.
+ * @param [in] finalLayout                a VkImageLayout value indicating the desired
+ *                                        final layout of the created image.
+ * @param [in] subAllocatorCallbacks      pointer to a potential set of thread-safe
```
MarkCallow, line 770: pointer to a set of thread-safe suballocator callbacks that wrap around suballocator calls: alloc, bindbuffer, bindimage, map, unmap and free. They use a uint64_t stored in the @c allocationId field of the structure pointed at by @a vkTexture to reference allocated page(s).

lib/vkloader.c
```diff
@@ -816,6 +826,14 @@ ktxTexture_VkUploadEx(ktxTexture* This, ktxVulkanDeviceInfo* vdi,
...
     ktx_uint32_t elementSize = ktxTexture_GetElementSize(This);
     ktx_bool_t               canUseFasterPath;
+    // Ensure that all callbacks are provided (or at least not NULL) before we start using them
+    ktx_bool_t               useSuballocator = (subAllocatorCallbacks && 
+                                                subAllocatorCallbacks->allocMemFuncPtr &&
+                                                subAllocatorCallbacks->bindBufferFuncPtr && 
+                                                subAllocatorCallbacks->bindImageFuncPtr && 
+                                                subAllocatorCallbacks->memoryMapFuncPtr && 
+                                                subAllocatorCallbacks->memoryUnmapFuncPtr && 
+                                                subAllocatorCallbacks->freeMemFuncPtr);
```
MarkCallow, line 837: I am concerned about silently manually allocate VkDeviceMemory when an effort has been made to provide suballocator callbacks. How about returning KTX_INVALID_OPERATION if subAllocatorCallbacks is non-zero but not all callbacks have been provided.

lib/vkloader.c
```diff
@@ -1044,19 +1062,36 @@ ktxTexture_VkUploadEx(ktxTexture* This, ktxVulkanDeviceInfo* vdi,
...
+                        memReqs.size, 0,
+                        (void**)&pMappedStagingBuffer));
+        }
+        else {
+            uint64_t numPages = 0ull;
+            stagingAllocId = subAllocatorCallbacks->allocMemFuncPtr(&memAllocInfo, &memReqs, &numPages);
+            if (stagingAllocId == 0ull) {
+                return KTX_OUT_OF_MEMORY;
+            }
+            assert(numPages == 1ull); // We do NOT support sparse bindings yet.
```
MarkCallow, line 1087: Is numPages reliant on the provided allocmem function? If so, could it sometimes return 1 and other times > 1? If so then I think this should be an error not an assert.

lib/vkloader.c
```diff
@@ -1763,18 +1858,53 @@ generateMipmaps(ktxVulkanTexture* vkTexture, ktxVulkanDeviceInfo* vdi,
...
+    ktx_bool_t useSuballocator = (subAllocatorCallbacks &&
+                                  subAllocatorCallbacks->allocMemFuncPtr &&
+                                  subAllocatorCallbacks->bindBufferFuncPtr &&
+                                  subAllocatorCallbacks->bindImageFuncPtr &&
+                                  subAllocatorCallbacks->memoryMapFuncPtr &&
+                                  subAllocatorCallbacks->memoryUnmapFuncPtr &&
+                                  subAllocatorCallbacks->freeMemFuncPtr);
+
+    vkTexture->vkDestroyImage(device, vkTexture->image, pAllocator);
+    if (useSuballocator)
```
MarkCallow, line 1885: Same comment I raised at line 836.

lib/vkloader.c
```diff
@@ -1763,18 +1858,53 @@ generateMipmaps(ktxVulkanTexture* vkTexture, ktxVulkanDeviceInfo* vdi,
...
+        vkTexture->vkFreeMemory(device, vkTexture->deviceMemory, pAllocator);
+}
+
+/**
+ * @memberof ktxVulkanTexture
+ * @~English
+ * @brief Destructor for the object returned when loading a texture image.
+ *
+ * Calls ktxVulkanTexture_Destruct_WithPotentialSuballocator() without a
+ * set of (hopefully thread-safe) suballocator callbacks.
```
MarkCallow, line 1897: s/ (hopefully thread safe)//

MarkCallow (changes requested):
Looks great overall. I spotted some documentation things I'd like fixed and have a couple of questions about error handling.

You need to fix the code warnings that are causing CI builds to fail. (CI builds are done with warnings as errors set.)

include/ktxvulkan.h
```diff
@@ -124,15 +125,34 @@ typedef struct ktxVulkanTexture
     uint32_t depth; /*!< The depth of the image. */
     uint32_t levelCount; /*!< The number of MIP levels in the image. */
     uint32_t layerCount; /*!< The number of array layers in the image. */
+    uint64_t allocationId; /*!< An allocationId referencing suballocation(s). */
```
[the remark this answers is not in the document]
toomuchvoltage: Perfect, will do!

include/ktxvulkan.h
```diff
@@ -124,15 +125,34 @@ typedef struct ktxVulkanTexture
...
+    ktxVulkanTexture_subAllocatorAllocMemFuncPtr allocMemFuncPtr;
+    ktxVulkanTexture_subAllocatorBindBufferFuncPtr bindBufferFuncPtr;
+    ktxVulkanTexture_subAllocatorBindImageFuncPtr bindImageFuncPtr;
+    ktxVulkanTexture_subAllocatorMemoryMapFuncPtr memoryMapFuncPtr;
+    ktxVulkanTexture_subAllocatorMemoryUnmapFuncPtr memoryUnmapFuncPtr;
+    ktxVulkanTexture_subAllocatorFreeMemFuncPtr freeMemFuncPtr;
+} ktxVulkanTexture_subAllocatorCallbacks;
+
+KTX_API void KTX_APIENTRY
+ktxVulkanTexture_Destruct_WithPotentialSuballocator(ktxVulkanTexture* This, VkDevice device,
```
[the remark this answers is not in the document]
toomuchvoltage: Sure thing! 👍

include/ktxvulkan.h
```diff
@@ -208,6 +228,13 @@ ktxVulkanDeviceInfo_Destruct(ktxVulkanDeviceInfo* This);
 KTX_API void KTX_APIENTRY
 ktxVulkanDeviceInfo_Destroy(ktxVulkanDeviceInfo* This);
 KTX_API KTX_error_code KTX_APIENTRY
+ktxTexture_VkUploadEx_WithPotentialSuballocator(ktxTexture* This, ktxVulkanDeviceInfo* vdi,
```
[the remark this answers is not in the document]
toomuchvoltage: Yup, will take care of this!

lib/vkloader.c
```diff
@@ -749,20 +749,29 @@ linearTilingPadCallback(int miplevel, int face,
  * limited number of formats and features. Generally @c VK_IMAGE_TILING_OPTIMAL
  * is preferred. The latter requires a staging buffer so will use more memory
  * during loading.
+ * 
+ * If a pointer to a thread-safe set of suballocator callbacks is provided, they
+ * will be used instead of manual allocation of a VkDeviceMemory. A 64 bit uint
```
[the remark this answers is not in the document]
toomuchvoltage: Ok, will do!

lib/vkloader.c
```diff
@@ -749,20 +749,29 @@ linearTilingPadCallback(int miplevel, int face,
...
+ * is returned on memory procurement that references the suballocated page(s).
```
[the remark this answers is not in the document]
toomuchvoltage: 👍

lib/vkloader.c
toomuchvoltage: Ok, got it! I guess the @A and @c have meaning here.

lib/vkloader.c
toomuchvoltage: Yea, that works: more explicit.

lib/vkloader.c
toomuchvoltage: The issue here is that if the allocator has sparse binding support numPages can be > 1. Currently the way the code is structured, there needs to be additional work before the code can handle numPages > 1 (i.e. it would entail memcpy calls more than once and some heavy testing to ensure we're not messing things up). Do we have an error code like "feature not supported"?

lib/vkloader.c
toomuchvoltage: Sounds good! 👍

lib/vkloader.c
toomuchvoltage: 👍

toomuchvoltage: Actually regarding this... what should I do here since `Destruct` has no return value. Should the inner call return a `ktx_error_code_e` while the outer one returns void?

lib/vkloader.c
toomuchvoltage: Just used `KTX_UNSUPPORTED_FEATURE`. Re-open if not satisfied.

toomuchvoltage:
Dear @MarkCallow , I think I pretty much addressed everything. Regarding checking the completeness of suballocator callbacks in `Destruct`: I basically return a KTX error code if they're incomplete but this does not bubble up to the regular `Destroy` as that returns nothing. Such an interface is reasonable as generally free/destroy calls return void in Vulkan (i.e. `vkFreeDescriptorSets` or `vkFreeCommandBuffers`). Should we make an exception for silent failures in this case?

If everything looks good at this stage, let me know and I will provide the VMA tests shortly.

Much appreciated!

lib/vkloader.c
```diff
@@ -749,46 +749,62 @@ linearTilingPadCallback(int miplevel, int face,
  * limited number of formats and features. Generally @c VK_IMAGE_TILING_OPTIMAL
  * is preferred. The latter requires a staging buffer so will use more memory
  * during loading.
+ * 
+ * If a pointer to a set of suballocator callbacks is provided, they
+ * will be used instead of manual allocation of VkDeviceMemory. A 64 bit uint
+ * that references the suballocated page(s) is returned on memory procurement 
+ * and saved in the @c allocationId field of the structure pointed to by @A vkTexture.
```
MarkCallow, line 756: s/@A/@a/. An error in my request. Sorry.

lib/vkloader.c
MarkCallow: @a means display the word in italics as it is one of the function arguments. That @a became @A is the result of some autocorrect I didn't notice before. @c means display the word in typewriter font as it is code. See https://www.doxygen.nl/manual/commands.html#cmda.

lib/vkloader.c
```diff
@@ -749,46 +749,62 @@ linearTilingPadCallback(int miplevel, int face,
...
+ *                                        on the Vulkan device.
+ * @param [in] usageFlags                 a set of VkImageUsageFlags bits indicating the
+ *                                        intended usage of the destination image.
+ * @param [in] finalLayout                a VkImageLayout value indicating the desired
+ *                                        final layout of the created image.
+ * @param [in] subAllocatorCallbacks      pointer to a set of suballocator callbacks 
+ *                                        that wrap around suballocator calls: alloc, 
+ *                                        bindbuffer, bindimage, map, unmap and free.
+ *                                        They use a uint64_t stored in the @c allocationId
+ *                                        field of the structure pointed at by @A vkTexture
```
MarkCallow, line 775: s/@A/@a/. We're victims of autocorrect here too.

lib/vkloader.c
MarkCallow: On reflection I think KTX_INVALID_VALUE is better as it is a problem with the value of the subAllocatorCallbacks argument rather than the internal state of the texture object not allowing the operation.

lib/vkloader.c
```diff
@@ -1763,18 +1873,62 @@ generateMipmaps(ktxVulkanTexture* vkTexture, ktxVulkanDeviceInfo* vdi,
...
+ * @sa ktxVulkanTexture_Destruct_WithSuballocator() for details and use that for complete
+ *     control.
  */
 void
 ktxVulkanTexture_Destruct(ktxVulkanTexture* vkTexture, VkDevice device,
                           const VkAllocationCallbacks* pAllocator)
 {
-    vkTexture->vkDestroyImage(device, vkTexture->image, pAllocator);
-    vkTexture->vkFreeMemory(device, vkTexture->deviceMemory, pAllocator);
+    ktxVulkanTexture_Destruct_WithSuballocator(vkTexture, device, pAllocator, NULL);
```
MarkCallow, line 1930: s/ktxVulkan/(void)ktxVulkan/ to indicate we know we are ignoring the return value.

lib/vkloader.c
MarkCallow: I think this is fine but, as in line with another comment I just made w.r.t. VkUploadEx_WithSubAllocator use KTX_INVALID_VALUE.

lib/vkloader.c
```diff
@@ -1763,18 +1873,62 @@ generateMipmaps(ktxVulkanTexture* vkTexture, ktxVulkanDeviceInfo* vdi,
...
  *
- * @param vkTexture  pointer to the ktxVulkanTexture to be destructed.
- * @param device     handle to the Vulkan logical device to which the texture was
- *                   loaded.
- * @param pAllocator pointer to the allocator used during loading.
+ * @param vkTexture                pointer to the ktxVulkanTexture to be destructed.
+ * @param device                   handle to the Vulkan logical device to which the texture was
+ *                                 loaded.
+ * @param pAllocator               pointer to the allocator used during loading.
+ * @param subAllocatorCallbacks    potential function pointer for guarded suballocator callbacks.
```
MarkCallow, line 1883: Change description to

"pointer to a structure of guarded suballocator function callbacks."

MarkCallow:
The failure in some of the Windows builds is due to a clang update, with a new warning, in the GitHub CI Windows runners. I plan to fix the code later today. After the fix you will need to rebase on main.

lib/vkloader.c
toomuchvoltage: Done! 👍

lib/vkloader.c
toomuchvoltage: 👍

lib/vkloader.c
toomuchvoltage: Fixed!

lib/vkloader.c
toomuchvoltage: Fixed!

lib/vkloader.c
toomuchvoltage: Just changed both the specs and the return value.

lib/vkloader.c
toomuchvoltage: Yup, just changed the return value + specs.

toomuchvoltage:
> The failure in some of the Windows builds is due to a clang update, with a new warning, in the GitHub CI Windows runners. I plan to fix the code later today. After the fix you will need to rebase on main.

@MarkCallow Sounds good. Just pushed changes addressing the most recent feedbacks. If all looks good, I'll get the tests ready. Also would appreciate a ping here once your changes on main are complete! 🙏

MarkCallow (approved):

MarkCallow:
> Just pushed changes addressing the most recent feedbacks. If all looks good, I'll get the tests ready. Also would appreciate a ping here once your changes on main are complete!

All looks good. I'll ping you when the build issue is fixed. I'm suffering from a very very slow computer I have to use for testing fixes to the issue.

MarkCallow:
The fix for clang 16 is now in `main`.

toomuchvoltage:
> The fix for clang 16 is now in `main`.

Thanks @MarkCallow . I will post some updates soon. I'm tied up a bit on my end at the moment.

toomuchvoltage:
Hi @MarkCallow , I just added a new test under `vkloadtests`. Could you please take a look at it to see if everything looks good? If so, I'll proceed with the rebase. Here's a screenshot of it running:

image

toomuchvoltage:
A couple of notes:
* The contents of the folders in `Debug` or `Release` had to be dumped in both folders for `vkloadtests` to run:
image
* Had to explicitly use `KTX_FEATURE_LOADTEST_APPS=Vulkan` instead of ``KTX_FEATURE_LOADTEST_APPS=ON`. Otherwise the `vkloadtests` VC project is not constructed.

toomuchvoltage:
Dear @MarkCallow , I think at this point the only remaining errors in CI are related to warnings from VMA. Any way around this?

MarkCallow:
> * Had to explicitly use `KTX_FEATURE_LOADTEST_APPS=Vulkan` instead of `KTX_FEATURE_LOADTEST_APPS=ON`. Otherwise, the `vkloadtests` VC project is not constructed.

I changed this a couple of months ago. As there is no Vulkan SDK for Windows/ARM it is not possible to build the vkloadtests target there so I need to be able to build just the gl3loadtests target.

MarkCallow:
As for the warnings I think the best way to mark it as a system include in cmake. Look at how `other_include` is added to the vkloadtests target and do something similar or move the vma headers there. We have an `other_projects` for the source of external projects that are incorporated via CMake. The {fmt} library is included that way. If vma has a CMakeLists, as {fmt} does, put it in `other_projects` otherwise put it in `other_include`

toomuchvoltage:
> As for the warnings I think the best fix is to mark it as a system include in cmake. Look at how `other_include` is added to the vkloadtests target and do something similar or move the vma headers there. We have an `other_projects` for the source of external projects that are incorporated via CMake. The {fmt} library is included that way. If vma has a CMakeLists, as {fmt} does, put it in `other_projects` otherwise put it in `other_include`

Ahh, thanks for the tip. I've been really struggling with this for like a solid day. I'll just wait to see what the most recent change does... and after that I'm gonna move it to `other_include`.

MarkCallow:
> * The contents of the folders in `Debug` or `Release` had to be dumped in both folders for `vkloadtests` to run:

If this is what happens with the original vkloadtests then don't worry about it.

toomuchvoltage:
Hi @MarkCallow , I think everything's pretty much set to go. There's one test that is failing with missing licensing information I think: `Unfortunately, your project is not compliant with version 3.0 of the REUSE Specification :-(`. Any ideas on that?

toomuchvoltage:
Hi @MarkCallow , all done. It's telling me that merging must happen by an authorized user, so I guess you'll have to merge this out?

MarkCallow (approved):

MarkCallow:
Can you move the VMA wrappers to VulkanLoadTestSample.{cpp,h} and also add a callbacks structure to the class? The latter is so there is no need to fill it in again when calling destruct. I phrase as a question because I do not know if these functions are general or specific to this case being a simple texture.

tests/loadtests/vkloadtests/Texture.cpp
```diff
@@ -118,10 +217,28 @@ Texture::Texture(VulkanContext& vkctx,
...
+        callbacks.memoryMapFuncPtr = VMA_CALLBACKS::MapMemoryCWrapper;
+        callbacks.memoryUnmapFuncPtr = VMA_CALLBACKS::UnmapMemoryCWrapper;
+        callbacks.freeMemFuncPtr = VMA_CALLBACKS::FreeMemCWrapper;
+
+        ktxresult = ktxTexture_VkUploadEx_WithSuballocator(kTexture, &vdi, &texture,
+                                                           static_cast<VkImageTiling>(tiling),
+                                                           VK_IMAGE_USAGE_SAMPLED_BIT,
+                                                           VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL, &callbacks);
+    }
+    else
```
MarkCallow, line 237: Add comment:
```
// Keep separate call to also test ktxTexture_VkUploadEx.
```

toomuchvoltage: Added!

tests/loadtests/vkloadtests/VulkanLoadTestSample.h
```diff
@@ -51,6 +65,16 @@ class VulkanLoadTestSample : public LoadTestSample {
                                     const std::string sBasePath);
 
   protected:
+
+    ktxVulkanTexture_subAllocatorCallbacks callbacks = {
```
MarkCallow, line 69: `callbacks` is a too generic a name. I'm struggling to come up with anything better.  My first thought was `vmaCallbacks` but in theory other suballocators could also be used, right?

toomuchvoltage: Yes, that is correct. Just called it `subAllocatorCallbacks`.

MarkCallow (commented):
Name change is fine.

Thinking to also move the useSubAllocator bool to VulkanLoadTestSample but ideally then VulkanLoadTestSample would handle parsing the command line for the matching option. IIRC there is no hierarchical processing of the command line. So we'll leave this idea for later.

Please look into the build errors on Windows. They look like something to do with the way you are using VMA.

toomuchvoltage:
> Name change is fine.
> 
> Thinking to also move the useSubAllocator bool to VulkanLoadTestSample but ideally then VulkanLoadTestSample would handle parsing the command line for the matching option. IIRC there is no hierarchical processing of the command line. So we'll leave this idea for later.
> 
> Please look into the build errors on Windows. They look like something to do with the way you are using VMA.

Hi @MarkCallow , I need your help on this one. I've been incapable of replicating it on my side. My usage is perfectly in line with what VMA wants: just 1 include in a CPP file. It seems like it's a bunch of warnings about VMA itself? I thought dropping the file in `other_includes` prevents that warning... but not from `VulkanLoadTestSample`? I could use a second look and some hints/pointers. Deeply appreciated! 🙏

MarkCallow:
Possibly msvc has no `-systemi` equivalent that CMake can use. Not sure. But just in case try

```
#if defined(_MSC_VER)
  #pragma warning(push)
  #pragma warning(disable: 4100)
  #pragma warning(disable: 4234)
#endif
```

and

```
#if defined(_MSC_VER)
  #pragma warning(pop)
#endif
```

around the include of VMA.

toomuchvoltage:
Hi @MarkCallow , can you give this another spin? Hopefully it's fixed.

toomuchvoltage:
Dear @MarkCallow , I also attempted to address some warnings highlighted here in the last commit. If you could review and let me know whether the changes actually address these, it would be greatly appreciated! 🙏 

image

lib/vkloader.c
```diff
@@ -1366,7 +1476,7 @@ ktxTexture_VkUpload(ktxTexture* texture, ktxVulkanDeviceInfo* vdi,
  *
  * This simplly calls ktxTexture_VkUploadEx.
  *
- * @copydetails ktxTexture::ktxTexture_VkUploadEx
+ * @copybrief ktxTexture::ktxTexture_VkUploadEx
```
MarkCallow, line 1479: Please revert back to @copydetails.

lib/vkloader.c
```diff
@@ -1406,7 +1516,7 @@ ktxTexture1_VkUpload(ktxTexture1* texture, ktxVulkanDeviceInfo* vdi,
...
```
MarkCallow, line 1519: Revert this too.

MarkCallow (changes requested):
Please revert the 2 changes indicated.

MarkCallow:
You need to also suppress warning 4189 which I missed in the blizzard of 4100 warnings in the CI log and I transposed digits in the other warning. It should be 4324 not 4234. Sorry about that.

As far as I can see, you have fixed fix the undocumented `ktxVulkanTexture_subAllocatorCallbacks` warning by providing documentation. The other Doxygen warnings (missing reference targets) are new with the latest version of Doxygen. I'm looking into them now. I won't hold up this PR for fixes. Ignore them.

lib/vkloader.c
toomuchvoltage: Reverted in most recent force-push.

lib/vkloader.c
toomuchvoltage: Reverted in most recent force-push.

toomuchvoltage:
> You need to also suppress warning 4189 which I missed in the blizzard of 4100 warnings in the CI log and I transposed digits in one of the other warnings. It should be 4324 not 4234. Sorry about that.
> 
> As far as I can see, you have fixed fix the undocumented `ktxVulkanTexture_subAllocatorCallbacks` warning by providing documentation. The other Doxygen warnings (missing reference targets) are new with the latest version of Doxygen. I'm looking into them now. I won't hold up this PR for fixes. Ignore them.

All done! I guess time for another go... 🙏

MarkCallow (approved):
Looks good. I'll try the build again.

MarkCallow:
Thank you very much for your hard work on this.

MarkCallow:
Did you run the --use-vma test under the Vulkan debug layers? I am getting warnings and an error. I'm running under MoltenVK on an M2 MacBook. It says that it is attempting to map memory that was allocated without the VK_MEMORY_PROPERTY_HOST_VISIBLE bit set. Then an error telling me memory mapping failed because GPU-only memory can't be mapped then crash. Please investigate.

toomuchvoltage:
Hi @MarkCallow , I've identified and fixed the issue. It had to do with my memory property flag detection being tied to a specific vendor (nVidia). I've fixed the issue in my most recent remote branch. You can perhaps rip out the last commit and merge once more if that works. If not, I can open a new PR.

MarkCallow:
Please open a new PR. Sorry about the bum steer on the `vkloadtests` option to use to enable validation.

toomuchvoltage:
> Please open a new PR. Sorry about the bum steer on the `vkloadtests` option to use to enable validation.

No, my bad. That was a really good find. I switched cards to my Radeon VII and the test crashed.

PR #778, opened 2023-10-02
Fix for NSIS PATH size limitation

aqnuep:


aqnuep:
@MarkCallow, I've verified, using the NSIS version with large string support does fix the installation-time PATH limit issue. Here's the PR to fix it.

The mingw build fails, but there seems to be some general issue with Github CI, it's not related to the changes. Please take a look.

MarkCallow:
Terrific. Thanks. Please try adding the following line after line 41 of `.github/workflows/mingw.yml`

```
        static: 0
```

It will prevent the action from trying to remove the now non-existent file. Hopefully whatever other effects it has will not break the build.

aqnuep:
> Terrific. Thanks. Please try adding the following line after line 41 of `.github/workflows/mingw.yml`
> 
> ```
>         static: 0
> ```
> 
> It will prevent the action from trying to remove the now non-existent file. Hopefully whatever other effects it has will not break the build.

It seems it just causes other problems down the line.

Is mingw installation actually needed? Other Khronos ecosystem repositories seem to be able to use mingw without any prior setup necessary.

I guess I'll give it a try.

MarkCallow:
> It seems that this will solve all of our problems with mingw.

Thanks. I wasn't aware that the Windows runners had mingw already installed. A contributor set up the MinGW build.

PR #779, opened 2023-10-02
KTXwriterScParams support

### LINKED ISSUE #758: No `ktx` tool writes KTXwriterScParams metadata.
I've only just noticed this. Something else I missed during project review.

@aqnuep I don't expect you do fix this but please give me a clue as to how to capture the relevant options (names and values) to a std::string in cxxopts to become the value of `KTXwriterScParams`. The options to be captured are those in `compress_utils.h`, `encode_utils.h` and the ASTC options in `ktx create`.

The information is incredibly useful when looking at an existing .ktx2 file to see exactly how it was compressed.

aqnuep:
Fixes #758.

Depends on the CTS PR https://github.com/KhronosGroup/KTX-Software-CTS/pull/5.

@MarkCallow FYI (please make sure you merge the CTS PR first with rebasing policy).

### DIFF
--- a/tools/ktx/command_create.cpp
+++ b/tools/ktx/command_create.cpp
@@ -525,10 +525,11 @@ struct OptionsCreate {
 };
 
 struct OptionsASTC : public ktxAstcParams {
-    bool astc = false;
+    std::string astcOptions{};
+    bool encodeASTC = false;
     ClampedOption<ktx_uint32_t> qualityLevel{ktxAstcParams::qualityLevel, 0, KTX_PACK_ASTC_QUALITY_LEVEL_MAX};
 
-    OptionsASTC() {
+    OptionsASTC() : ktxAstcParams() {
         threadCount = std::thread::hardware_concurrency();
         if (threadCount == 0)
             threadCount = 1;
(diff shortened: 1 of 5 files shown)


MarkCallow:
Please rebase to get the MinGW fix.

Looks good but the repetition of so many string values is unfortunate. Is there a way to reduce the repetition? If not, please define a const string object for each value and refer to that at each place where needed.

aqnuep:
> Please rebase to get the MinGW fix.
> 
> Looks good but the repetition of so many string values is unfortunate. Is there a way to reduce the repetition? If not, please define a const string object for each value and refer to that at each place where needed.

Yeah, I tried to minimize that, in fact eliminated one of them. We could use some consts, sure.

MarkCallow:
> Yeah, I tried to minimize that, in fact eliminated one of them. We could use some consts, sure.

Please do it.

Note that me e-mail service is currently broken. I can neither send nor receive messages from my regular addresses. Therefore my responses may be somewhat delayed.

aqnuep:
Removed "astc-mode" option, as requested. Instead, default to LDR mode for UNORM/SRGB.

aqnuep:
Updated CTS git submodule hash. Ready for merge.

Issue #792, opened 2023-10-29
ktxTexture_GetImageSize doesn't handle 3D textures


AlexRouSg:
I am trying to create a 3D `ktxTexture2` from image data in memory however it fails at `ktxTexture_SetImageFromMemory` with `KTX_INVALID_OPERATION.`

Debugging the code it appears as if `ktxTexture_GetImageSize` doesn't handle 3D texture sizes at all which then triggers this failure condition in `ktxTexture2_setImageFromStream`
```
    if (srcSize != imageByteLength)
        return KTX_INVALID_OPERATION;
```

This is on the main branch just pulled at the time of this issue creation.

Unless I am meant to set the individual 2D slices of the 3D texture then the documentation is not clear that is what I have to do.

MarkCallow:
The documentation for `ktxTexture[12]_GetImageSize says

```
 * For arrays, this is the size of a layer, for cubemaps, the size of a face
 * and for 3D textures, the size of a depth slice.
 ```

The documentation for the `faceSlice` parameter of `ktxTexture2_setImageFromStream` says

```
 * @param[in] faceSlice cube map face or depth slice of the image to set.
```

which to me clearly indicates you have to upload a slice at a time. Happy to make clarifications, if you'd like to propose something concrete.

I'd also be happy to better support creation of 3d textures. Are you working a file format that supports 3d image data? Please propose an API that would helpful to you. Probably will need new SetImage functions that take all slices at once and would not be usable with any other type of texture.

Are you generating mip levels for your 3d textures? The generator currently used by the KTX tools only supports 2d mip level generation. An extra dimension of filtering is needed for 3d.

AlexRouSg:
> which to me clearly indicates you have to upload a slice at a time. Happy to make clarifications, if you'd like to propose something concrete.

That's kinda surprising to me that you have to do it that way, cause I've edited the `ktxTexture_GetImageSize` function to include depth and it works and creates a valid ktx2 file when given an uncompressed R8G8B8A8 image. But I don't know what is the correct size calculation to handle every format.

> Are you generating mip levels for your 3d textures?

Not currently but maybe, still investigating what to do atm.

MarkCallow:
> That's kinda surprising to me that you have to do it that way, 

It works as documented...

> cause I've edited the ktxTexture_GetImageSize function to include depth and it works

[...]

PR #794, opened 2023-11-02
libktx: update ktxTexture2_setImageFromStream to allow setting the entire level's data in one call

AlexRouSg:
This change updates `ktxTexture2_setImageFromStream` to allow setting `faceSlice` to `-1`,
This allows setting the entire level's data in one call rather having to do a loop and setting each slice individually.

Fixes: #792

### DIFF
--- a/lib/writer2.c
+++ b/lib/writer2.c
@@ -180,7 +180,8 @@ appendLibId(ktxHashList* head, ktxHashListEntry* writerEntry)
  * @param[in] This      pointer to the target ktxTexture object.
  * @param[in] level     mip level of the image to set.
  * @param[in] layer     array layer of the image to set.
- * @param[in] faceSlice cube map face or depth slice of the image to set.
+ * @param[in] faceSlice cube map face or depth slice of the image to set or
+ *                      KTX_FACESLICE_WHOLE_LEVEL to set the entire level.
  * @param[in] src       ktxStream pointer to the source.
  * @param[in] srcSize   size of the source image in bytes.
  *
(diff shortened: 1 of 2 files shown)


lib/writer2.c
```diff
@@ -180,7 +180,7 @@ appendLibId(ktxHashList* head, ktxHashListEntry* writerEntry)
  * @param[in] This      pointer to the target ktxTexture object.
  * @param[in] level     mip level of the image to set.
  * @param[in] layer     array layer of the image to set.
- * @param[in] faceSlice cube map face or depth slice of the image to set.
+ * @param[in] faceSlice cube map face or depth slice of the image to set or (-1) to set the entire level.
```
MarkCallow, line 183: Make this and other similar lines <= 80 characters. Going on to a second line is okay.

lib/writer2.c
```diff
@@ -208,13 +208,19 @@ ktxTexture2_setImageFromStream(ktxTexture2* This, ktx_uint32_t level,
...
         return KTX_INVALID_OPERATION;
 
-    result = ktxTexture_GetImageOffset(ktxTexture(This),
-                                       level, layer, faceSlice,
-                                       &imageByteOffset);
-    if (result != KTX_SUCCESS)
-       return result;
-
-    imageByteLength = ktxTexture_GetImageSize(ktxTexture(This), level);
+    if (faceSlice == (ktx_uint32_t)(-1)) {
```
MarkCallow, line 211: Since this does not change the type of the faceSlice parameter, which I agree with, let's document this as specify UINT_MAX (i.e. the value of -1 cast to uint32_t) instead of -1 to set the entire level. Consider making a define to be used, e.g. KTX_ENTIRE_LEVEL.

MarkCallow (changes requested):
Thanks for this. See my small comments.

lib/writer2.c
AlexRouSg: done

lib/writer2.c
AlexRouSg: done, I did it that way to avoid including limits.h

MarkCallow (approved):

PR #830, opened 2023-12-29
Fix RPM default directory permission

### LINKED ISSUE #827: libglvnd-devel conflict when installing KTX-Software-4.2.1 x86_64 RPM on Fedora with a pre-installed SDL2
On Fedora 39, installing 4.2.1 reports this error:

```
rpm -ivh  KTX-Software-4.2.1-Linux-x86_64.rpm 
Verifying...                          ################################# [100%]
Preparing...                          ################################# [100%]
	file /usr/share/man from install of ktx-software-4.2.1-1.x86_64 conflicts with file from package filesystem-3.18-6.fc39.x86_64
	file /usr/include/KHR from install of ktx-software-4.2.1-1.x86_64 conflicts with file from package libglvnd-devel-1:1.7.0-1.fc39.x86_64

```

It looks like the conflicts is due by a previous install of `SDL2` that depends on `libglvnd-devel`

there is a  file located there:
```
ll /usr/include/KHR/
total 12
-rw-r--r--. 1 root root 11131 Sep 12 15:02 khrplatform.h
```

The point is it is conflicting with the `/usr/include/KHR/` directory, not with the file itself.

The same issue is for the `/usr/share/man` directory: It already exists and it report as a conflict.

----

I didn't find any `CPACK_RPM_USER_BINARY_SPECFILE` to point to a rpm `.spec` file to look into it and find the possible error, so i guess is something auto-geneerated with CPack default parameters that makes the install target directory mandatory to be not present, like if it is not a system directory to be shared among other packages.

Raffaello:
This fixes #827,  the conflicting permissions in the directories, those lead to impossibility to install the package.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1246,6 +1246,16 @@ elseif(LINUX)
     set(CPACK_GENERATOR DEB RPM TBZ2)
     set(CPACK_PACKAGE_CHECKSUM SHA1)
     set(CPACK_PACKAGE_DESCRIPTION_FILE "${CMAKE_CURRENT_SOURCE_DIR}/cmake/ReadMe.txt")
+
+    # This fixes the directory permission derived from Ubuntu ${BUILD_DIR} of 0775 to 0755.
+    # otherwise when installing the RPM, it conflicts with the base filesystem.rpm that is
+    # setting the directory permissions and owns them to 0755.
+    # See GitHub Issue #287
+    # NOTE: Alternatively It would be better build the RPM through a docker or RedHat OS in the CI
+    set(CPACK_RPM_DEFAULT_DIR_PERMISSIONS
+    OWNER_READ OWNER_WRITE OWNER_EXECUTE
+    GROUP_READ GROUP_EXECUTE
+    WORLD_READ WORLD_EXECUTE)
 elseif(WIN32)
     set(CPACK_GENERATOR "NSIS")
     # Add logo to top left of installer pages. Despite the installer-
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1253,9 +1253,9 @@ elseif(LINUX)
     # See GitHub Issue #827.
     # NOTE: Alternatively It would be better build the RPM through a docker or RedHat OS in the CI
     set(CPACK_RPM_DEFAULT_DIR_PERMISSIONS
-    OWNER_READ OWNER_WRITE OWNER_EXECUTE
-    GROUP_READ GROUP_EXECUTE
-    WORLD_READ WORLD_EXECUTE)
+        OWNER_READ OWNER_WRITE OWNER_EXECUTE
+        GROUP_READ GROUP_EXECUTE
+        WORLD_READ WORLD_EXECUTE)
 elseif(WIN32)
     set(CPACK_GENERATOR "NSIS")
     # Add logo to top left of installer pages. Despite the installer-
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -646,7 +646,7 @@ macro(common_libktx_settings target enable_write library_type)
         find_package(Threads REQUIRED)
         target_link_libraries(
             ${target}
-        PRIVATE
+        PUBLIC
             dl
             Threads::Threads
         )
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -646,7 +646,7 @@ macro(common_libktx_settings target enable_write library_type)
         find_package(Threads REQUIRED)
         target_link_libraries(
             ${target}
-        PUBLIC
+        PRIVATE
             dl
             Threads::Threads
         )
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1250,7 +1250,7 @@ elseif(LINUX)
     # This fixes the directory permission derived from Ubuntu ${BUILD_DIR} of 0775 to 0755.
     # otherwise when installing the RPM, it conflicts with the base filesystem.rpm that is
     # setting the directory permissions and owns them to 0755.
-    # See GitHub Issue #287
+    # See GitHub Issue #827.
     # NOTE: Alternatively It would be better build the RPM through a docker or RedHat OS in the CI
     set(CPACK_RPM_DEFAULT_DIR_PERMISSIONS
     OWNER_READ OWNER_WRITE OWNER_EXECUTE
--- a/other_projects/cxxopts/cmake/cxxopts.cmake
+++ b/other_projects/cxxopts/cmake/cxxopts.cmake
@@ -159,5 +159,14 @@ function(cxxopts_install_logic)
             DESTINATION "${CMAKE_INSTALL_LIBDIR_ARCHIND}/pkgconfig"
     )
 
+    # this should fix directory permission derived from Travis-CI ubuntu ${BUILD_DIR} from 0775 to 0755.
+    # otherwise when installing the RPM, it will conflicting with the base filesystem.rpm that is
+    # setting the directory permission and owns them to 0755
+    # NOTE: Alternatively would be better build the RPM through a docker or RedHat OS in the CI
+    set(CPACK_RPM_DEFAULT_DIR_PERMISSIONS
+    OWNER_READ OWNER_WRITE OWNER_EXECUTE
+    GROUP_READ GROUP_EXECUTE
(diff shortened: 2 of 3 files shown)


Raffaello:
related issue #827

other_projects/cxxopts/cmake/cxxopts.cmake
```diff
@@ -159,5 +159,14 @@ function(cxxopts_install_logic)
             DESTINATION "${CMAKE_INSTALL_LIBDIR_ARCHIND}/pkgconfig"
     )
 
+    # this should fix directory permission derived from Travis-CI ubuntu ${BUILD_DIR} from 0775 to 0755.
```
Raffaello, line 162: if it fixes, change to `fix` instead of `should` and reference the github issue too (#827)
(when travis ci is completed, the rpm pacakge should be updated in the release page as a pre-release version [4.3.x] or something)

Raffaello:
I was hoping to run Travis-CI in debug mode and ssh into the container to just check the resulting rpm(s) permissions.

with this command:

```bash
rpm -q --queryformat="[%{FILEMODES:perms} %{FILEUSERNAME}:%{FILEGROUPNAME} %{FILENAMES}\n]" KTX-Software-...
```

but i don't have permissions on that.

----

I have create a podman container running `ubuntu:jammy` and compiling within the container a basic debug build using the ci_script `BUILD_DIR=build PACKAGE=YES ./ci_scripts/build_linux.sh`

the resulting rpm package: `build/KTX-Software-4.3.0-beta1-Linux-x86_64.rpm`

listing the file permissions looks ok:

````
drwxr-xr-x root:root /usr/include/KHR             # <--- this is 0755 instead of 0775
-rw-r--r-- root:root /usr/include/KHR/khr_df.h
```

MarkCallow:
Thank you for this.

I suspect the directory permission difference is due to the earlier version of cmake used by Travis CI (3.26.3) rather than something else in the Ubuntu environment.

Because of the repo secrets debug mode is disabled. Also packages are only deployed to GitHub Releases for version-tagged commits. Before the next release I will build a draft release and check the permissions with the command you gave.

MarkCallow:
Unfortunately it is not working on my local Ubuntu 22.04.

```
 $ rpm -q --queryformat="[%{FILEMODES:perms} %{FILEUSERNAME}:%{FILEGROUPNAME} %{FILENAMES}\n]" KTX-Software-*.rpm
-rwxr-xr-x root:root /usr/bin/ktx
-rwxr-xr-x root:root /usr/bin/ktx2check
-rwxr-xr-x root:root /usr/bin/ktx2ktx2
-rwxr-xr-x root:root /usr/bin/ktxinfo
-rwxr-xr-x root:root /usr/bin/ktxsc
-rwxr-xr-x root:root /usr/bin/toktx
drwxrwxr-x root:root /usr/include/KHR
-rw-r--r-- root:root /usr/include/KHR/khr_df.h
-rw-r--r-- root:root /usr/include/ktx.h
-rw-r--r-- root:root /usr/include/ktxvulkan.h
drwxr-xr-x root:root /usr/lib/.build-id
drwxr-xr-x root:root /usr/lib/.build-id/4f
lrwxrwxrwx root:root /usr/lib/.build-id/4f/3bc0d541f899a3d5e2870b34841a759c213a45
drwxr-xr-x root:root /usr/lib/.build-id/6d
lrwxrwxrwx root:root /usr/lib/.build-id/6d/4ffb3d56efbf335b8e5f273558bcfb6d980be9
drwxr-xr-x root:root /usr/lib/.build-id/8c
lrwxrwxrwx root:root /usr/lib/.build-id/8c/f8b8f0212398200dbaeedbafde06e2905af7a4
drwxr-xr-x root:root /usr/lib/.build-id/a6
lrwxrwxrwx root:root /usr/lib/.build-id/a6/6441bb2ab2d55830a7bcc1ad800606a2f1c1dd
drwxr-xr-x root:root /usr/lib/.build-id/ce
lrwxrwxrwx root:root /usr/lib/.build-id/ce/04c917bb65d7cd542d75726d0b520029b2621a
drwxr-xr-x root:root /usr/lib/.build-id/d8
lrwxrwxrwx root:root /usr/lib/.build-id/d8/61939374586c6c456291e4dbf1de51bbaa2362
drwxrwxr-x root:root /usr/lib/cmake
drwxrwxr-x root:root /usr/lib/cmake/ktx
-rw-r--r-- root:root /usr/lib/cmake/ktx/KtxConfig.cmake
-rw-r--r-- root:root /usr/lib/cmake/ktx/KtxConfigVersion.cmake
-rw-r--r-- root:root /usr/lib/cmake/ktx/KtxTargets-release.cmake
-rw-r--r-- root:root /usr/lib/cmake/ktx/KtxTargets.cmake
-rw-r--r-- root:root /usr/lib/libktx-4.3.0.jar
lrwxrwxrwx root:root /usr/lib/libktx-jni.so
lrwxrwxrwx root:root /usr/lib/libktx-jni.so.4
-rw-r--r-- root:root /usr/lib/libktx-jni.so.4.3.0
lrwxrwxrwx root:root /usr/lib/libktx.so
lrwxrwxrwx root:root /usr/lib/libktx.so.4
-rw-r--r-- root:root /usr/lib/libktx.so.4.3.0
```

As you can see `/usr/include/KHR` still has 775 permissions as does `/usr/lib/cmake` and `/usr/lib/cmake/ktx`.

Raffaello:
> Thank you for this.
> 
> I suspect the directory permission difference is due to the earlier version of cmake used by Travis CI (3.26.3) rather than something else in the Ubuntu environment.
> 
> Because of the repo secrets debug mode is disabled. Also packages are only deployed to GitHub Releases for version-tagged commits. Before the next release I will build a draft release and check the permissions with the command you gave.

Actually it is very likely that is a bug in `CPackRPM` too: I found one with manual files when building on Fedora: 
https://gitlab.kitware.com/cmake/cmake/-/issues/25545

Raffaello:
> rpm -q --queryformat="[%{FILEMODES:perms} %{FILEUSERNAME}:%{FILEGROUPNAME} %{FILENAMES}\n]" KTX-Software-*.rpm

@MarkCallow I updated as you asked. 
But after your comment not working on ubunty, i tried doing the same building without that change and in my case is still getting the right permissions. So i guess it might not work as expected on a bare Ubuntu system.

Please let me know when you have time. 
Thanks

CMakeLists.txt
```diff
@@ -1246,6 +1246,16 @@ elseif(LINUX)
     set(CPACK_GENERATOR DEB RPM TBZ2)
     set(CPACK_PACKAGE_CHECKSUM SHA1)
     set(CPACK_PACKAGE_DESCRIPTION_FILE "${CMAKE_CURRENT_SOURCE_DIR}/cmake/ReadMe.txt")
+
+    # This fixes the directory permission derived from Ubuntu ${BUILD_DIR} of 0775 to 0755.
+    # otherwise when installing the RPM, it conflicts with the base filesystem.rpm that is
+    # setting the directory permissions and owns them to 0755.
+    # See GitHub Issue #287
```
MarkCallow, line 1253: 827.

MarkCallow:
I just tried with the latest fix and can confirm it works. All directories now have 755 permissions.

> But after your comment not working on ubuntu, i tried doing the same building without that change and in my case is still getting the right permissions. So i guess it might not work as expected on a bare Ubuntu system.

Check your `/etc/login.defs`. Maybe it sets `USERGROUPS_ENAB` or `UMASK` differently to the default. Or perhaps your user is a member of a real group rather than a private user group. Do an `ls -l` on some file you created. If your user name and group name are the same you have a private user group. If not, then the default umask will be `022` instead of `002` so directories will be created with mode `755`.

The fix you have provided is good protection against strange umasks though arguably CPackRPM should be making sure well known shared directories have the expected permissions.

CMakeLists.txt
```diff
@@ -1246,6 +1246,16 @@ elseif(LINUX)
...
+    # NOTE: Alternatively It would be better build the RPM through a docker or RedHat OS in the CI
+    set(CPACK_RPM_DEFAULT_DIR_PERMISSIONS
+    OWNER_READ OWNER_WRITE OWNER_EXECUTE
```
MarkCallow, line 1256: Please follow the style used in other multi-line sets in this file:
```
set(foo
    bar
    non
)
```

Raffaello:
this also Fixes #831

PR #844, opened 2024-01-27
Add license, license_files, long_description_content_type, long_description in pyktx setup.py

ShukantPal:
@MarkCallow https://github.com/KhronosGroup/KTX-Software/pull/698#issuecomment-1891222766

### DIFF
--- a/interface/python_binding/setup.py
+++ b/interface/python_binding/setup.py
@@ -21,8 +21,12 @@ setup(
     ],
     include_package_data=True,
     install_requires=["cffi>=1.15.1"],
+    license="Apache 2.0",
+    license_files=('LICENSE'),
+    long_description_content_type="text/markdown",
+    long_description="This Python package provides a Pythonic interface to libktx. It uses CFFI to generate the C bindings.",
     packages=['pyktx'],
     package_dir={'pyktx': 'pyktx'},
     setup_requires=["cffi>=1.15.1"],
-    url='https://github.com/KhronosGroup/KTX-Software'
+    url='https://github.com/KhronosGroup/KTX-Software',
 )


MarkCallow:
This PR fixed the previous problems. Now deploy is almost working but now I get this error:

```
ERROR    HTTPError: 400 Bad Request from https://test.pypi.org/legacy/          
         Binary wheel 'pyktx-4.3.0-cp310-cp310-linux_aarch64.whl' has an        
         unsupported platform tag 'linux_aarch64'.                              
```

What is the proper tag? Please fix the architecture name or advise if I should remove this .whl from the files we attempt to upload.

MarkCallow:
Looking at [this page](https://packaging.python.org/en/latest/specifications/platform-compatibility-tags/#platform-tags-for-common-linux-distributions) it seems we should use `manylinux` not `linux`. See also [this](https://github.com/pypa/manylinux). Please advise.

MarkCallow:
As I told you before we also still need to find an alternative to `kittaakos/upload-artifact-as-is` as it uses version 12 of node.js which has been removed from GitHub Actions so the action is forced to run on node 16. There seems to be no problem with it doing so but node 16 is deprecated and will be removed at which point it will be forced to use node 20. Then who knows. Plus the warning is ugly. The GitHub project has been archived and hasn't been touched in 4 years.

This is much less urgent than the platform tag as it is not a showstopper for the 4.3.0 release.

ShukantPal:
> [Building manylinux-compatible wheels is not trivial
](https://github.com/pypa/manylinux?tab=readme-ov-file#docker-images)


The binary wheels are not essentials for publishing the package; since pip will auto build pyktx against the system `libktx` on install time, it's best to remove linux wheels from the publishing process.

ShukantPal:
https://github.com/KhronosGroup/KTX-Software/pull/845

PR #847, opened 2024-02-02
Exclude linux *.whl files from publishing to PyPI

ShukantPal:
This is needed since we don't have a workflow to build manylinux wheels at moment, and [it's non-trivial to implement that](https://github.com/pypa/manylinux?tab=readme-ov-file#docker-images)

#845


MarkCallow:
Thanks for the tip to exclude the Linux wheels. I've implemented that via a modified RE rather than an extra step so I'll close this.

PR #855, opened 2024-02-07
Prevent warning if VK_NO_PROTOTYPES is already defined

jherico:


MarkCallow:
@jherico how come you encountered this?  `vk_funcs.h` is internal to the library.

jherico:
The [Vulkan-Sample](https://github.com/KhronosGroup/Vulkan-Samples/) uses KTX as a git submodule and incorporates the build directly into it's top level project, so any upstream warnings show up as part of the build of that project.

MarkCallow:
> The [Vulkan-Sample](https://github.com/KhronosGroup/Vulkan-Samples/) uses KTX as a git submodule and incorporates the build directly into it's top level project, so any upstream warnings show up as part of the build of that project.

But where did the other definition of VK_NO_PROTOTYPES come from that was visible when libktx was being compiled? When you say "build directly" do you mean Vukan-Sample is including the libktx sources directly and setting up its own compile commands?

jherico:
Yes, exactly.  https://github.com/KhronosGroup/Vulkan-Samples/blob/main/third_party/CMakeLists.txt#L50

MarkCallow:
Thanks for the info. and this fix.

PR #856, opened 2024-02-09
Fix recreateBytesPlane0 and add test

MarkCallow:
for all valid VkFormats.

Merge needs to wait until issue #857 is fixed so it can be tested against correct DFDs.


lib/dfdutils/queries.c
```diff
@@ -19,6 +19,49 @@
...
+    uint32_t bytesPlane2: 8;
+    uint32_t bytesPlane3: 8;
+    uint32_t bytesPlane4: 8;
+    uint32_t bytesPlane5: 8;
+    uint32_t bytesPlane6: 8;
+    uint32_t bytesPlane7: 8;
+    struct sampleType samples[6];
+};
+#endif
+
```
aqnuep, line 64: I'm concerned about introducing yet another definition for BDFD types. We already have too many, and this is prone to be the source of errors. Can't the unit test simply use the one in e.g. imageio?

Regarding big-endian support, we did find various places in libktx where we had concerns whether it actually works on big-endian machines, so I think big-endian support is a wider open question, but the definitions here wouldn't work on big-endian.

MarkCallow: As the comment says, its just to help with debugging. It isn't used. This file is actually in the dfdutils so it is difficult to share any header except a public header. Possibly I could add this struct definition to `dfd.h`. What do you think?

aqnuep: Well, if this is only expected to be used by tests, I'd rather put it there. Putting it into dfdutils makes it look like an official definition to be used by dfdutils library users.

MarkCallow:
@aqnuep there are now a small number of failures in the extract tool tests having to do with extracting `X8_D24_UNORM_PACK32` format to .raw files. I suspect it because the previous version of `recreate...` returned a value of 3 for this format. The new one returns the correct 4. Most likely I just need to regenerate the golden files. What is a good way to examine the content of a .raw file? I can see the current golden file has a size of 192 while the new output is 256 bytes which suggests it is indeed due to the 4 instead of 3.

Apologies for the bug in the previous version of `recreate...`.

I think I'm going to rename the function to `reconstructBytesPlane0`. I prefer your "reconstruct" and the current name is too long. I think I'll also make it return the value. For dfdutils users I'll have to provide a wrapper for backward API compatibiility.

aqnuep:
*> What is a good way to examine the content of a .raw file?*

There's no easier way than to inspect the file in a hex viewer. You should
be able to compare the payloads.
But you can also try to feed that raw file back into ktx create, if that
makes it easier.

-Daniel

lib/dfdutils/queries.c
MarkCallow: I'll delete it. It's at the top of `texture2.c` if ever needed again.

MarkCallow:
> I can see the current golden file has a size of 192 while the new output is 256 bytes which suggests it is indeed due to the 4 instead of 3.

The incorrect bytesPlane0 value messed up the calculation in `ktxTexture2_GetImageSize` resulting in not writing enough bytes to the raw file. I've updated the golden images in https://github.com/KhronosGroup/KTX-Software-CTS/pull/17. Hopefully I've managed to point this branch at the branch in the CTS. .gitmodules is still pointing at branch main.

aqnuep:
> Hopefully I've managed to point this branch at the branch in the CTS. .gitmodules is still pointing at branch main.

The branch in .gitmodules is the the "tracked" branch. That confused me to in the past, don't worry about it. Only the commit hash the git module points to is relevant.

PR #864, opened 2024-03-01
Flesh out support for VK_FORMAT_R16G16_S10_5_NV.

### LINKED ISSUE #859: Support R16G16_S10_5_NV format
Over in https://github.com/KhronosGroup/KTX-Specification/issues/206 we've decided to support this format.

It is in `vkFormatString` and `isValidFormat` so `ktx info` and `ktx validate` should be fine as these are the functions they use for printing the name and checking validity. `vk2dfd` however produces nothing for this format. `makevk2dfd.pl` needs to be fixed and some tests added. Once that is fixed the code in `unittests.cc` to avoid running the bytesPlane0 test for this format must be removed.

There is no intention to support this in `ktx create` except via raw files.

MarkCallow:
Code is believed complete. Tests have yet to be added. Hence draft status.

This includes a new round-trip test of vk2dfd and dfd2vk for all Vulkan formats except multi-plane and modifications to both functions, or rather their generators, to fix support for some formats, including VK_FORMAT_R16G16_S10_5_NV and the YUV 422 single plane formats. Those listed required changes to`interpretDFD`.

Fixes #859.


MarkCallow:
@fluppeteer, please review. I was unable to add your GitHub id to the reviewers list.

lib/dfdutils/createdfd.c
```diff
@@ -158,6 +162,10 @@ static void writeSample(uint32_t *DFD, int sampleNo, int channel,
         upper.f = 1.0f;
         lower.f = 0.0f;
         break;
+    case s_S10_5:
+        assert(bits == 16 && "Format with this suffix must be 16 bits per channel.");
+        upper.i = 32;
+        lower.i = ~upper.i + 1; // -32
```
aqnuep, line 168: This doesn't seem correct at all.

Per the Vulkan spec, the S10_5_NV format has a range of [-1024,1023.96875] in floating point format, as it's a fixed-point format with 1 sign bit, 10 integer bits and 5 fractional bits.

aqnuep: Thus, assuming this is treated as an integer format, then the range should be the full 0 - 0xFFFFFFFF range, and if it is treated as float, then it probably should be -1024.0 - 1023.96875.

MarkCallow: Float flag is not set.

lib/dfdutils/dfd.h
```diff
@@ -35,7 +35,8 @@ enum VkSuffix {
     s_SINT,    /*!< Signed integer format. */
     s_SFLOAT,  /*!< Signed float format. */
     s_UFLOAT,  /*!< Unsigned float format. */
-    s_SRGB     /*!< sRGB normalized format. */
+    s_SRGB,     /*!< sRGB normalized format. */
```
fluppeteer, line 38: Trivially, removed the space replaced by the comma.
```suggestion
    s_SRGB,    /*!< sRGB normalized format. */
```

lib/dfdutils/dfd.h
```diff
@@ -111,10 +113,11 @@ enum InterpretDFDResult {
     i_SRGB_FORMAT_BIT          = 1u << 2u, /*!< sRGB transfer function. */
     i_NORMALIZED_FORMAT_BIT    = 1u << 3u, /*!< Normalized (UNORM or SNORM). */
     i_SIGNED_FORMAT_BIT        = 1u << 4u, /*!< Format is signed. */
-    i_FLOAT_FORMAT_BIT         = 1u << 5u, /*!< Format is floating point. */
-    i_COMPRESSED_FORMAT_BIT    = 1u << 6u, /*!< Format is block compressed (422). */
-    i_YUVSDA_FORMAT_BIT        = 1u << 7u, /*!< Color model is YUVSDA. */
-    i_UNSUPPORTED_ERROR_BIT    = 1u << 8u, /*!< Format not successfully interpreted. */
+    i_FIXED_FORMAT_BIT         = 1u << 5u,
```
fluppeteer, line 116: Should we have a documentation line on this, even if it's trivial or obvious?
```suggestion
    i_FIXED_FORMAT_BIT         = 1u << 5u, /*!< Format has a fixed-point representation. */
```

lib/dfdutils/dfd2vk.inl
```diff
@@ -71,133 +80,170 @@ if (KHR_DFDVAL(dfd + 1, MODEL) == KHR_DF_MODEL_RGBSDA) {
...
+          return VK_FORMAT_B10G11R11_UFLOAT_PACK32;
+      } else if (R.size == 10 && G.size == 10 && B.size == 0) { 
+          return VK_FORMAT_R10X6G10X6_UNORM_2PACK16;
+      } else if (R.size == 12 && G.size == 12 && B.size == 0) { 
+          return VK_FORMAT_R12X4G12X4_UNORM_2PACK16;
+      }
+    } else if (wordBytes == 8) { /* 4PACK16 */
+      if (r & i_YUVSDA_FORMAT_BIT) {
+        if (R.size == 10 && R.offset == 6 && A.size == 10 && A.offset == 38)
+          return  VK_FORMAT_G10X6B10X6G10X6R10X6_422_UNORM_4PACK16;
```
fluppeteer, line 93: I'm very rusty, but I don't think this is right. 4PACK16 means four values each of which should be treated as 16-bit packed; the 16-bit units are in left-to-right order with increasing address as with other Vulkan and GL formats which list the channels independently, but then the contents of those 16-bit units are listed left-to-right as with packed formats (because of how big-endian numbers are written if you're drawing out the bit assignments).

VK_FORMAT_G10X6B10X6G10X6R10X6_422_UNORM_4PACK16 is 4 16-bit packed values:

G10X6 in the 16 bits starting at address offset 0 (in bytes).
B10X6 in the 16 bits starting at address offset 2 (in bytes).
G10X6 in the 16 bits starting at address offset 4 (in bytes).
R10X6 in the 16 bits starting at address offset 6 (in bytes).

Within those 16-bit quantities (which can have the bytes ordered by the implementation's endianness), the X channel takes up the low 6 bits, and the colour channel takes up the top ten bits.

In other words, for a little-endian implementation and the KDFS convention of counting all the offsets starting at bit 0 of byte 0, I believe VK_FORMAT_G10X6B10X6G10X6R10X6_422_UNORM_4PACK16 matches:

R.size == 10 && R.offset == 54 && A.size == 10 && A.offset == 38.

(TL;DR: swap the conditions with the format below.)

Incidentally, might be nice to add a reminder comment that the second G channel is getting packed into alpha for the purposes of this translation. :-)

lib/dfdutils/dfd2vk.inl
```diff
@@ -71,133 +80,170 @@ if (KHR_DFDVAL(dfd + 1, MODEL) == KHR_DF_MODEL_RGBSDA) {
...
+        if (R.size == 10 && R.offset == 22 && A.size == 10 && A.offset == 54)
+          return VK_FORMAT_B10X6G10X6R10X6G10X6_422_UNORM_4PACK16;
+        if (R.size == 12 && R.offset == 4 && A.size == 12 && A.offset == 36)
```
fluppeteer, line 96: As above, this has R.offset == 52 and A.offset == 36, I think.

lib/dfdutils/dfd2vk.inl
```diff
@@ -71,133 +80,170 @@ if (KHR_DFDVAL(dfd + 1, MODEL) == KHR_DF_MODEL_RGBSDA) {
...
+          return VK_FORMAT_G12X4B12X4G12X4R12X4_422_UNORM_4PACK16;
+        if (R.size == 12 && R.offset == 20 && A.size == 12 && A.offset == 52)
```
fluppeteer, line 98: As above, R.offset == 36 and A.offset == 52.

lib/dfdutils/dfd2vk.inl
```diff
@@ -71,133 +80,170 @@ if (KHR_DFDVAL(dfd + 1, MODEL) == KHR_DF_MODEL_RGBSDA) {
...
-          if (!(r & i_NORMALIZED_FORMAT_BIT) && !(r & i_SIGNED_FORMAT_BIT)) return VK_FORMAT_R8G8B8_UINT;
-          if (!(r & i_NORMALIZED_FORMAT_BIT) && (r & i_SIGNED_FORMAT_BIT)) return VK_FORMAT_R8G8B8_SINT;
-        } else { /* BGR */
-          if ((r & i_SRGB_FORMAT_BIT)) return VK_FORMAT_B8G8R8_SRGB;
-          if ((r & i_NORMALIZED_FORMAT_BIT) && !(r & i_SIGNED_FORMAT_BIT)) return VK_FORMAT_B8G8R8_UNORM;
-          if ((r & i_NORMALIZED_FORMAT_BIT) && (r & i_SIGNED_FORMAT_BIT)) return VK_FORMAT_B8G8R8_SNORM;
-          if (!(r & i_NORMALIZED_FORMAT_BIT) && !(r & i_SIGNED_FORMAT_BIT)) return VK_FORMAT_B8G8R8_UINT;
-          if (!(r & i_NORMALIZED_FORMAT_BIT) && (r & i_SIGNED_FORMAT_BIT)) return VK_FORMAT_B8G8R8_SINT;
+    if (r & i_YUVSDA_FORMAT_BIT) {
+      /* Conventions for mapping YUV to RGB channel names in DFD and Vulkan are different. */
```
fluppeteer, line 109: (As a note, Vulkan is more sensible - R <=> Cr, B <=> Cb, G <=> Y' - but "VYU" looks very odd to the uninitiated, so the KDFS went with convention, and may regret it.)

lib/dfdutils/dfd2vk.inl
```diff
@@ -71,133 +80,170 @@ if (KHR_DFDVAL(dfd + 1, MODEL) == KHR_DF_MODEL_RGBSDA) {
...
+      if (R.size == 1 && R.offset == 0 && A.size == 1 && A.offset == 2)
+        return VK_FORMAT_G8B8G8R8_422_UNORM;
```
fluppeteer, line 111: The leftmost G8 is in the low byte; should that not read G.size == 1 && G.offset == 0 (or R.offset == 3)?

lib/dfdutils/dfd2vk.inl
```diff
@@ -71,133 +80,170 @@ if (KHR_DFDVAL(dfd + 1, MODEL) == KHR_DF_MODEL_RGBSDA) {
...
+      else if (R.size == 1 && R.offset == 1 && A.size == 1 && A.offset == 3)
+        return VK_FORMAT_B8G8R8G8_422_UNORM;
```
fluppeteer, line 113: As above, this is G.offset == 1. (In all these formats G.offset %2 == A.offset %2.)

lib/dfdutils/dfd2vk.inl
```diff
@@ -71,133 +80,170 @@ if (KHR_DFDVAL(dfd + 1, MODEL) == KHR_DF_MODEL_RGBSDA) {
...
+      else if (R.size == 2 && R.offset == 0 && A.size == 2 && A.offset == 4)
+        return VK_FORMAT_G16B16G16R16_422_UNORM;
```
fluppeteer, line 115: Again 16-bit channels in increasing word address order left to right; G.offset == 0, B.offset == 2, "A".offset == 4, R.offset == 6.

lib/dfdutils/dfd2vk.inl
```diff
@@ -71,133 +80,170 @@ if (KHR_DFDVAL(dfd + 1, MODEL) == KHR_DF_MODEL_RGBSDA) {
...
+      else if (R.size == 2 && R.offset == 2 && A.size == 2 && A.offset == 6)
+        return VK_FORMAT_B16G16R16G16_422_UNORM;
```
fluppeteer, line 117: For completeness, B.offset == 0, G.offset == 2, R.offset == 4, "A".offset = 6.

lib/dfdutils/interpretdfd.c
```diff
@@ -288,8 +319,20 @@ enum InterpretDFDResult interpretDFD(const uint32_t *DFD,
                     currentByteOffset = sampleByteOffset;
                     currentBitLength = sampleBitLength;
                     if (sampleChannelPtr->size) {
-                        /* Uh-oh, we've seen this channel before. */
-                        return i_UNSUPPORTED_NONTRIVIAL_ENDIANNESS;
+                        if (model == KHR_DF_MODEL_YUVSDA && sampleChannel == KHR_DF_CHANNEL_YUVSDA_Y) {
+                            if (sampleChannelPtr == R) {
```
fluppeteer, line 323: Note that strictly speaking this should be checking whether the coordinates for the sample are different, since a big-endian 16-bit YUYV channel would be using two samples for each of Y0, U, Y1 and V. I presume that's rare enough that it's not worth the code to handle it, at least for now.

lib/dfdutils/makedfd2vk.pl
```diff
@@ -297,8 +294,17 @@ sub checkSuffices {
     else if (wordBytes == 2) { /* PACK16 */
       if (A.size == 4) {
         if (R.offset == 12) return VK_FORMAT_R4G4B4A4_UNORM_PACK16;
-        else return VK_FORMAT_B4G4R4A4_UNORM_PACK16;
-      } else if (A.size == 0) { /* Three channels */
+        else if (B.offset == 12) return VK_FORMAT_B4G4R4A4_UNORM_PACK16;
```
fluppeteer, line 297: B.offset == 4.

lib/dfdutils/makedfd2vk.pl
```diff
@@ -307,55 +313,105 @@ sub checkSuffices {
         if (B.offset == 10) return VK_FORMAT_A1B5G5R5_UNORM_PACK16_KHR;
         return VK_FORMAT_B5G5R5A1_UNORM_PACK16;
       }
-    } else if (wordBytes == 4) { /* PACK32 */
+    } else if (wordBytes == 4) { /* PACK32 or 2PACK16 */
```
fluppeteer, line 316: 2PACK16 should be wordBytes == 2, but I'm worried about the knock-on effects elsewhere.

lib/dfdutils/makedfd2vk.pl
```diff
@@ -307,55 +313,105 @@ sub checkSuffices {
...
+print "          return VK_FORMAT_R10X6G10X6_UNORM_2PACK16;\n";
+print "      } else if (R.size == 12 && G.size == 12 && B.size == 0) { \n";
+print "          return VK_FORMAT_R12X4G12X4_UNORM_2PACK16;\n";
+print "      }\n";
+
+$fourPack16Decode = << 'END_4PACK16';
+    } else if (wordBytes == 8) { /* 4PACK16 */
+      if (r & i_YUVSDA_FORMAT_BIT) {
+        if (R.size == 10 && R.offset == 6 && A.size == 10 && A.offset == 38)
+          return  VK_FORMAT_G10X6B10X6G10X6R10X6_422_UNORM_4PACK16;
```
fluppeteer, line 339: R.offset == 54, A.offset == 38.

lib/dfdutils/makedfd2vk.pl
```diff
@@ -307,55 +313,105 @@ sub checkSuffices {
...
+        if (R.size == 10 && R.offset == 22 && A.size == 10 && A.offset == 54)
+          return VK_FORMAT_B10X6G10X6R10X6G10X6_422_UNORM_4PACK16;
```
fluppeteer, line 341: R.offset = 38.

lib/dfdutils/makedfd2vk.pl
```diff
@@ -307,55 +313,105 @@ sub checkSuffices {
...
+        if (R.size == 12 && R.offset == 4 && A.size == 12 && A.offset == 36)
+          return VK_FORMAT_G12X4B12X4G12X4R12X4_422_UNORM_4PACK16;
```
fluppeteer, line 343: R.offset == 52.

lib/dfdutils/makedfd2vk.pl
```diff
@@ -307,55 +313,105 @@ sub checkSuffices {
...
+        if (R.size == 12 && R.offset == 20 && A.size == 12 && A.offset == 52)
+          return VK_FORMAT_B12X4G12X4R12X4G12X4_422_UNORM_4PACK16;
```
fluppeteer, line 345: R.offset == 36.

lib/dfdutils/makedfd2vk.pl
```diff
@@ -307,55 +313,105 @@ sub checkSuffices {
...
+  } else { /* Not a packed format */
+END_4PACK16
+
+print $fourPack16Decode;
+
+$yuvDecode = << 'END_YUV';
+    if (r & i_YUVSDA_FORMAT_BIT) {
+      /* Conventions for mapping YUV to RGB channel names in DFD and Vulkan are different. */
+      if (R.size == 1 && R.offset == 0 && A.size == 1 && A.offset == 2)
+        return VK_FORMAT_G8B8G8R8_422_UNORM;
```
fluppeteer, line 362: R.offset == 3. (Non-packed, left-to-right increasing address order in Vulkan.)

lib/dfdutils/makedfd2vk.pl
```diff
@@ -307,55 +313,105 @@ sub checkSuffices {
...
+      else if (R.size == 1 && R.offset == 1 && A.size == 1 && A.offset == 3)
+        return VK_FORMAT_B8G8R8G8_422_UNORM;
```
fluppeteer, line 364: R.offset == 2.

lib/dfdutils/makedfd2vk.pl
```diff
@@ -307,55 +313,105 @@ sub checkSuffices {
...
+      else if (R.size == 2 && R.offset == 0 && A.size == 2 && A.offset == 4)
+        return VK_FORMAT_G16B16G16R16_422_UNORM;
```
fluppeteer, line 366: R.offset == 3, A.offset == 2.

lib/dfdutils/makedfd2vk.pl
```diff
@@ -307,55 +313,105 @@ sub checkSuffices {
...
+      else if (R.size == 2 && R.offset == 2 && A.size == 2 && A.offset == 6)
+        return VK_FORMAT_B16G16R16G16_422_UNORM;
```
fluppeteer, line 368: R.offset == 4.

fluppeteer:
I *think* this mostly looks good, except that the multi-channel packing convention wires got crossed. I've added notes where I think the numbers should be different. Please poke me if you change them and want a re-review, or want to push back; I'll try to respond promptly this time.

lib/dfdutils/dfd.h
MarkCallow: Probably should have a commend but I will s/has/is/ as it fits better with the other comments.

lib/dfdutils/dfd2vk.inl
MarkCallow: I agree with your analysis of these packed types. Not sure how I got those R offsets so wrong. There must be errors in the vk2dfd as well since the round trip tests are all working. Did you review vk2dfd and not spot anything or did you not review it?

fluppeteer: Looks like you read them all as a single packed (32- or 64-bit) quantity. I did look at vk2dfd, and I just had another look, and don't think it looks wrong. I'll keep looking in case I'm missing something.

lib/dfdutils/interpretdfd.c
MarkCallow: IIRC correctly @fluppeteer I sent you an email asking some questions about samplePosition. All of the YUVSDA format DFDs we are dealing with have different sample positions for each sample not just Y1 and Y2. I am unaware of any Vulkan YUVSDA formats that would have samplePositions that are all the same or all 0 (as in RGBSDA). Hence I did not test samplePosition. The keyword here is "unaware".  I think you have a far better understanding of YUV than I do. If you think it should be tested I will add that.

aqnuep: FYI, for the single-plane 4:2:2 YUV formats (YUYV) we set the sample positions to be cosited even when creating KTX files (Y: 64, 128, 0, 0; U: 64, 128, 0, 0; Y: 192, 128, 0, 0; V: 64, 128, 0, 0), considering that there are currently no explicit controls to configure cositing explicitly.

lib/dfdutils/makedfd2vk.pl
MarkCallow: No. `wordBytes` will be 4. These formats are correctly, in my view, detected as packed because the offset is not a multiple of 8.

```
@param wordBytes[in,out] Pointer to a uint32_t to receive the byte size of
 *                          the channels (unpacked) or total size (packed).
 ```

lib/dfdutils/interpretdfd.c
fluppeteer: Ah - thanks @aqnuep - sorry, crossed over. :-) I'll look at the code shortly.

fluppeteer: (Cosited is fine, but I suspect we should be saying 0,0,0,0 for most of the samples and 128,0 for the second Y.)

lib/dfdutils/makedfd2vk.pl
fluppeteer: Oh, right, not the size of the "channel" word. I said I was rusty. :-) Ignore me.

lib/dfdutils/makedfd2vk.pl
MarkCallow: Fix made locally.

lib/dfdutils/makedfd2vk.pl
MarkCallow: Fix made locally.

lib/dfdutils/makedfd2vk.pl
MarkCallow: Fix made locally.

lib/dfdutils/makedfd2vk.pl
MarkCallow: Fix made locally.

lib/dfdutils/makedfd2vk.pl
MarkCallow: Fix made locally.

lib/dfdutils/makedfd2vk.pl
MarkCallow: Fix made locally.

lib/dfdutils/dfd.h
MarkCallow: Fix made locally.

lib/dfdutils/makedfd2vk.pl
aqnuep: > No. wordBytes will be 4. These formats are correctly, in my view, detected as packed because the offset is not a multiple of 8. Hence wordBytes == bytesPlane0.

Well, not really, I mean they are packed, but per-channel.

This is the general mPACKn format issue we've fixed whereas typeSize is actually 2 for these formats, just like regular 16-bit channel formats.

This wordBytes value is used in legacy code as the typeSize. But the DFD does not have the expression power to determine the wordBytes/typeSize from the DFD only.

One simply cannot differentiate at the DFD level a 4-byte format with 4-byte words with two 2-byte channels packed into them (i.e. PACK32) and a 4-byte format with 2-byte words with 2-byte channels (i.e. 2PACK16), and this makes a difference from the perspective of endianness.

This is an issue that we fixed for the Vk->DFD direction through the separate Vk->typeSize conversion, but it's not resolvable for DFD->Vk unless we add a typeSize (wordBytes) input argument making it a (DFD+typeSize)->Vk conversion. Conversions between VkFormat and DFD are simply not bijective, multiple VkFormats could result in the same DFD (e.g. a hypothetical `R10X6G10X6_PACK32` format would not be differentiable from a `R10X6G10X6_2PACK16` format at the DFD level, even though they have different packing granularities and thus different in-memory representation on a LE vs BE system).

aqnuep: Also, packed vs unpacked is not a correct categorization, hence the description of wordBytes itself is wrong, because mPACKn formats with m > 1 will consist of multiple, separately packed groups of channels (1 channel in each group in this case), so the whole packed vs unpacked format categorization (and the related assumptions in various parts of the code base) is wrong.

lib/dfdutils/dfd2vk.inl
MarkCallow: Whoops!!! The confusion here is the different mapping of YUV to RGB used by Vulkan and KDFS. 

`G10X6B10X6G10X6R10X6` for Vulkan is YUYV which in terms of KDFS  channel id values in `khr_df.h` is RGRB and in the output from `interpretDFD` is RGAB. Hence my original offsets were correct. I shall revert my local changes and add a more detailed comment about the mapping.

@fluppeteer please check if you agree.

lib/dfdutils/makedfd2vk.pl
aqnuep: > I could completely believe the mPACKn formats postdated this code, so the terminology could easily be misleading.

That is part of it, but also IIRC the DFD was never meant to abstract endianness, in fact it intentionally defines different representations depending on endianness (although I'm sure, @fluppeteer, you know that better), which is very different from how VkFormats are described.

fluppeteer: Yup, that's what I was incoherently trying to say in my first paragraph. :-)

lib/dfdutils/dfd2vk.inl
fluppeteer: *Oh*. Sorry, because KHR_DF_CHANNEL_RGBSDA_R = KHR_DF_CHANNEL_YUVSDA_Y (etc.) - and we're assuming that it's being interpreted as YUYV, not as the underlying GBGR (which it doesn't have to be).

That makes sense; I've not re-checked all your numbers, but it seems plausible. I'd advocate using aliased enum values so you can actually refer to the channels in YCbCr/YUV terms for these formats, and noting that we're assuming the YCBCR conversion (although if there's no samplerYCbCrconversion object then that still makes packed 444 ambiguous, because it looks like RGB).

Sorry for causing confusion.

lib/dfdutils/interpretdfd.c
fluppeteer: Edit: Finally had a look at the relevant bit of makevk2dfd.pl, which quite clearly documents what's going on. Sorry I'd missed this going in; I do think the offsets that incorporate a half-pixel coordinate offset are incorrect - see table 96 in the KDFS for a 4:2:0 format which places the Y samples at integers, or the definition of V210 in Table 100. I appreciate that the KDFS doesn't (AFAIR) say that one has to do this.

Since this merge doesn't affect those lines, I assume that's already in KTX. Is it possible to incorporate renormalisation (subtract minx and miny) in KTX and treat future content as using a 0 minimum? I'd be amenable to a note saying that someone reading a DFD might want to treat them as equivalent, but that the canonical version would start at 0. The placement of coordinate offsets should really be a matter for the implementation - DX has used both conventions at different times.

aqnuep: Of course, we could change the used sample positions for subsampled formats, if needed.

aqnuep: Just for me to understand though, why is 0 the midpoint of the pixel for sample position if the sample positions are unsigned integers? How could then one express e.g. sample positions left of the midpoint?

fluppeteer: Interesting. 0 isn't the midpoint so much as it's the origin of the repeating pixel pattern, which is assumed to map to the coordinates that your texture sampling code/hardware uses; in Vulkan terms they're integer i,j coordinates. That's a per-axis thing - I could imagine a 2x2 coordinate format (the result of checkerboard rendering?) with samples at 1,0 and 0,1. A 4:4:4 format which had chroma offset by half a pixel relative to luma (because... reasons?) would probably place one sample at 0 and the next at 0.5 (128). If the implementation (as with Vulkan's texturing) maps fractional coordinates such that the integer sample grid isn't a floor operation, that's its own choice.

But you do have a point; I could imagine a format where, in the interests of orthogonality, you might want to express a sample at a negative offset relative to what all the other formats you're dealing with would call "0", and that relocating a different sample to "0" would cause problems. I've kind of assumed that fractional coordinates happen at the higher coordinate end - you can have a sample at 0 and 0.5, but not -0.5 and 0. Arguably the latter is what you get if you flip a 4:2:2 image with an odd width. Cosited-even samples already have to flip to cosited-odd under similar circumstances, and transpositions already have to flip the direction of downsampling, so it's not entirely unreasonable.

I think in the specific case of a we-don't-know-what-it-is 4:2:2 format the origin should still align to the luma sample locations. You have a problem with any offset that it may not be possible to express the chroma sample location within the texel block coordinate range, if it falls on a boundary. The problem feels common enough that it's a bit unfortunate to have to rely on an extension block to disambiguate it, though. You could work around it by having the texel block size in coordinates different from the expressible subsample range, but that complicates the decoder, and once you're handling format specially the whole DFD goes out of the window.

I wonder whether this is worth adopting four of the "flags" bits to indicate inverting the coordinate system for each coordinate (unless I've forgotten using them for something; I'd thought about using a flag to indicate a bitfield channel representation, which would reduce the number of samples in some cases, but it's a relatively minor win for a doubling in the amount of decode logic). That might actually simplify the "flipping a 4:2:2" case. I need to think more, but if you have any thoughts on the subject, do let me know.

lib/dfdutils/dfd2vk.inl
MarkCallow: > I'd advocate using aliased enum values

Unfortunately it is not an enumeration here, it is the parameters passed to `interpretDFD`. I'll experiment with how to make it more clear. #defines and unions come to mind.

lib/dfdutils/interpretdfd.c
MarkCallow: I'll admit to being lost in this conversation. As the current sample positions pre-date this PR we'll leave them in place, not delay this PR and continue the discussion. It sounds like sample positions are generally implementation dependent. Since the DFD here is describing the data in the KTX file, not for any particular implementation, maybe the 0 and 1 that was suggested somewhere in here is the best option. Answering the question of how image processing operations would use the sample positions may help to make the solution clear. It is not a question I can answer.

lib/dfdutils/makedfd2vk.pl
MarkCallow: I've used #defines. The code is much more clear.

lib/dfdutils/dfd2vk.inl
MarkCallow: I used #defines. Much clearer.

lib/dfdutils/interpretdfd.c
MarkCallow: We need to capture this conversation to continue it after this PR is closed. Should I create an issue?

fluppeteer: Thanks Mark - please do. It relates to 4:2:2 and 4:2:0 formats in general, but certainly not specifically this one, and we should handle it separately.

fluppeteer:
The #define approach looks much easier to read (not a phrase I'll often say) - thanks for clarifying, and sorry again to have caused the churn!

Assuming all the test are happy, I think this is good to go, and we can talk about coordinate offsets separately.

PR #868, opened 2024-03-11
Merge ktx compare to main

aqnuep:
This PR contains all the code changes for the ktx compare tool, alongside other changes related to fixes needed to be done in order for ktx compare to work as expected (including the fixes related to the handling of the D24S8 format).

@MarkCallow, @lexaknyazev, FYI.

### DIFF
--- a/tools/ktx/command.h
+++ b/tools/ktx/command.h
@@ -77,6 +77,7 @@ enum class ReturnCode {
     DFD_FAILURE = RUNTIME_ERROR,
     NOT_SUPPORTED = 5,
     NOT_IMPLEMENTED = 6,
+    DIFFERENCE_FOUND = 7
 };
 using rc = ReturnCode;
 
@@ -364,6 +365,10 @@ class InputStream {
 public:
     InputStream(const std::string& filepath, Reporter& report);
 
+    const std::string& str() {
+        return filepath;
+    }
+
     /*explicit(false)*/ operator std::istream&() {
         return *activeStream;
     }
@@ -386,6 +391,11 @@ class OutputStream {
 public:
     OutputStream(const std::string& filepath, Reporter& report);
     ~OutputStream();
+
+    const std::string& str() {
+        return filepath;
+    }
+
     void writeKTX2(ktxTexture* texture, Reporter& report);
     void write(const char* data, std::size_t size, Reporter& report);
 };
--- a/tools/ktx/command_compare.cpp
+++ b/tools/ktx/command_compare.cpp
@@ -468,9 +468,11 @@ struct DiffMismatch {
 class PrintDiff {
     PrintIndent& printIndent;
     OutputFormat outputFormat;
-    std::vector<std::string> context;
-    bool firstContext;
-    bool different;
+    std::optional<std::string> jsonSection = {};
+    bool jsonSectionEmpty = false;
+    bool firstContext = true;
+    std::vector<std::string> context = {};
+    bool different = false;
 
     void printContext() {
         if (!context.empty()) {
@@ -482,12 +484,40 @@ class PrintDiff {
         }
     }
 
+    void beginJsonOutput() {
+        assert(jsonSection.has_value());
+
+        const auto space = outputFormat != OutputFormat::json_mini ?  " " : "";
+        const auto nl = outputFormat != OutputFormat::json_mini ?  "\n" : "";
+
+        printIndent(0, ",{}", nl);
+
+        if (std::exchange(jsonSectionEmpty, false))
+            printIndent(1, "\"{}\":{}{{{}", *jsonSection, space, nl);
+    }
+
 public:
     PrintDiff(PrintIndent& output, OutputFormat format)
-        : printIndent(output), outputFormat(format), context(), firstContext(true), different(false) {}
+        : printIndent(output), outputFormat(format) {}
 
     bool isDifferent() const {
         return different;
+    }
+
+    void beginJsonSection(std::string&& section) {
+        jsonSection = section;
+        jsonSectionEmpty = true;
+    }
+
+    void endJsonSection() {
+        if (jsonSection.has_value()) {
+            const auto nl = outputFormat != OutputFormat::json_mini ?  "\n" : "";
+            if (!jsonSectionEmpty) {
+                printIndent(0, "{}", nl);
+                printIndent(1, "}}");
+            }
+            jsonSection.reset();
+        }
     }
 
     void setContext(std::string&& ctx) {
@@ -519,17 +549,17 @@ public:
             if (diff.hasValue(1))
                 printIndent(0, "+{}: {}\n", diff.textHeader, diff.value(1, outputFormat));
         } else {
-            printIndent(0, ",{}", nl);
-            printIndent(1, "\"{}\":{}[{}", diff.jsonPath, space, nl);
+            beginJsonOutput();
+            printIndent(2, "\"{}\":{}[{}", diff.jsonPath, space, nl);
             if (diff.hasValue(0))
-                printIndent(2, "{},{}", diff.value(0, outputFormat), nl);
+                printIndent(3, "{},{}", diff.value(0, outputFormat), nl);
             else
-                printIndent(2, "null,{}", nl);
+                printIndent(3, "null,{}", nl);
             if (diff.hasValue(1))
-                printIndent(2, "{}{}", diff.value(1, outputFormat), nl);
+                printIndent(3, "{}{}", diff.value(1, outputFormat), nl);
             else
-                printIndent(2, "null{}", nl);
-            printIndent(1, "]");
+                printIndent(3, "null{}", nl);
+            printIndent(2, "]");
         }
     }
 
@@ -552,19 +582,19 @@ public:
                 diff.printText(1, printIndent, "+");
             }
         } else {
-            printIndent(0, ",{}", nl);
-            printIndent(1, "\"{}\":{}[{}", diff.jsonPath, space, nl);
+            beginJsonOutput();
+            printIndent(2, "\"{}\":{}[{}", diff.jsonPath, space, nl);
             if (diff.hasValue(0)) {
-                diff.printJson(0, printIndent, 2, space, nl);
+                diff.printJson(0, printIndent, 3, space, nl);
                 printIndent(0, ",{}", nl);
             } else
-                printIndent(2, "null,{}", nl);
+                printIndent(3, "null,{}", nl);
             if (diff.hasValue(1)) {
-                diff.printJson(1, printIndent, 2, space, nl);
+                diff.printJson(1, printIndent, 3, space, nl);
                 printIndent(0, "{}", nl);
             } else
-                printIndent(2, "null{}", nl);
-            printIndent(1, "]");
+                printIndent(3, "null{}", nl);
+            printIndent(2, "]");
         }
     }
 
@@ -583,13 +613,12 @@ public:
         different = true;
 
         const auto space = outputFormat != OutputFormat::json_mini ?  " " : "";
-        const auto nl = outputFormat != OutputFormat::json_mini ?  "\n" : "";
 
         if (outputFormat == OutputFormat::text) {
             printIndent(0, "+{}\n", diff.textMsg);
         } else {
-            printIndent(0, ",{}", nl);
-            printIndent(1, "\"{}\":{}[]", diff.jsonPath, space);
+            beginJsonOutput();
+            printIndent(2, "\"{}\":{}[]", diff.jsonPath, space);
         }
     }
 };
@@ -759,7 +788,7 @@ class CommandCompare : public Command {
 
         std::array<std::string, 2> inputFilepaths;
         ContentMode contentMode = ContentMode::raw;
-        uint64_t perPixelOutputLimit = 0;
+        std::size_t perPixelOutputLimit = 0;
         bool allowInvalidInput = false;
         bool ignoreFormatHeader = false;
         bool ignoreSupercomp = false;
@@ -868,13 +897,13 @@ class CommandCompare : public Command {
                     report.fatal_usage("--per-pixel-output is specified but --content was not set to \"image\".");
                 const auto perPixelOutputStr = to_lower_copy(args[kPerPixelOutput].as<std::string>());
                 if (perPixelOutputStr == "all") {
-                    perPixelOutputLimit = UINT64_MAX;
+                    perPixelOutputLimit = SIZE_MAX;
                 } else if (perPixelOutputStr == "none") {
                     perPixelOutputLimit = 0;
                 } else {
                     std::size_t parsedCharacters = 0;
                     try {
-                        perPixelOutputLimit = std::stoull(perPixelOutputStr, &parsedCharacters);
+                        perPixelOutputLimit = static_cast<std::size_t>(std::stoull(perPixelOutputStr, &parsedCharacters));
                     } catch (std::exception&) {}
                     if (parsedCharacters != perPixelOutputStr.length())
                         report.fatal_usage("Invalid --per-pixel-output arugment: \"{}\".", perPixelOutputStr);
@@ -971,6 +1000,9 @@ private:
     void compareKVEntry(PrintDiff& diff, const std::string_view& key,
         ktxHashListEntry* entry1, ktxHashListEntry* entry2);
     void compareSGD(PrintDiff& diff, InputStreams& streams);
+    void compareImages(PrintDiff& diff, InputStreams& streams);
+    void compareImagesRaw(PrintDiff& diff, InputStreams& streams);
+    void compareImagesPerPixel(PrintDiff& diff, InputStreams& streams);
 
     void read(InputStream& stream, std::size_t offset, void* readDst, std::size_t readSize, std::string_view what) {
         stream->seekg(offset);
@@ -1044,10 +1076,10 @@ void CommandCompare::executeCompare() {
             validationMessages.emplace_back(std::move(messagesOS).str());
         }
 
-        bool hasValidationError = false;
+        bool hasValidationMessages = false;
         for (std::size_t i = 0; i < inputStreams.size(); ++i) {
             if (!validationMessages[i].empty()) {
-                if (std::exchange(hasValidationError, true))
+                if (std::exchange(hasValidationMessages, true))
                     fmt::print("\n");
 
                 fmt::print("Validation {} for '{}'\n", validationResults[i] == 0 ? "successful" : "failed",
@@ -1058,12 +1090,16 @@ void CommandCompare::executeCompare() {
         }
 
         for (std::size_t i = 0; i < inputStreams.size(); ++i) {
-            if (validationResults[i] != 0)
+            if (validationResults[i] != 0) {
                 if (validationResults[i] != +rc::INVALID_FILE || !options.allowInvalidInput)
                     throw FatalError(ReturnCode{validationResults[i]});
+
+                // Image comparison is only supported for valid input files
+                options.contentMode = ContentMode::ignore;
+            }
         }
 
-        if (hasValidationError)
+        if (hasValidationMessages)
             fmt::print("\n");
 
         PrintIndent out{std::cout};
@@ -1073,6 +1109,7 @@ void CommandCompare::executeCompare() {
         compareDFD(diff, inputStreams);
         compareKVD(diff, inputStreams);
         compareSGD(diff, inputStreams);
+        compareImages(diff, inputStreams);
 
         if (diff.isDifferent())
             throw FatalError(rc::DIFFERENCE_FOUND);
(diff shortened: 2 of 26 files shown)


aqnuep:
@MarkCallow could you prioritize reviewing this and the related CTS PR this week?

It is practically the combined PR of all the previously reviewed PRs plus rebasing/bug fixing, so it should be fairly straightforward.

tools/ktx/command_compare.cpp
```diff
@@ -0,0 +1,2465 @@
...
+    "\x1A" as an unescaped "\x1A" sequence inside a JSON string breaks nearly every JSON tool.
+    Note that this does not change the value of the string only its representation.
+
+    @note @b ktx @b compare prints using UTF-8 encoding. If your console is not
+    set for UTF-8 you will see incorrect characters in output of the file
+    identifier on each side of the "KTX nn".
+
+    The following options are available:
+    @snippet{doc} ktx/command.h command options_format
+    <dl>
```
MarkCallow, line 836: For some reason that I can't see, most of the text is hugely indented as you can see in the following. It makes reading the documentation difficult on a reasonable width terminal. Also some options have a different indent than others. I think the latter may be due to some coming from snippets.

Oh! Wait. This output is from `ktx compare --help`. I'm not sure this doxygen code is the source for this output.

```
  -f, --format text|json|mini-json
                                Specifies the report output format. Possible options are:
                                  text: Human readable text based format
                                  json: Formatted JSON
                                  mini-json: Minified JSON
                                 (default: text)
      --content raw|image|ignore
                                Controls how image content is compared. Possible values are:
                                  raw: Encoded image data is compared verbatim, as it appears in 
                                the file
                                  image: Effective image data is compared per texel block
                                  ignore: Ignore image contents
                                Note: When the mode is set to @b image, the effective image data of 
                                individual mip levels, layers, and faces is compared texel block by 
                                texel block after applying one or more of the following 
                                transformations, if needed:
                                - If the texture is supercompressed (e.g. using Zstandard or ZLIB), 
                                the images are deflated before comparison.
                                - If the texture is encoded using BasisLZ, the images are 
                                transcoded to R8G8B8A_UNORM before comparison.
                                For block compressed textures and textures encoded using UASTC the 
                                texel blocks are the individual compressed blocks, while for other 
                                formats the texel blocks are the individual pixels of the image.
                                When comparing files that use different supercompression schemes, 
                                or otherwise different encoding, additional options may need to be 
                                used to avoid unexpected differences to be reported related to the 
                                meta information of the files. For example:
                                - The supercompression scheme can be ignored with --ignore-supercomp
                                - Compressed byte length and other index section differences can be 
                                ignored with --ignore-index all or --ignore-index level
                                - DFD section differences can be ignored with --ignore-dfd all or 
                                --ignore-dfd all-except-color-space
                                - SGD section differences can be ignored with --ignore-sgd all or 
                                --ignore-sgd payload
                                 (default: raw)
      --per-pixel-output all|<number>|none
                                Controls whether per pixel / texel block difference output is 
                                generated when --content is set to image:
                                  all - Every single difference is output (may r
```

tools/ktx/command_compare.cpp
```diff
@@ -0,0 +1,2465 @@
...
+            @b ignore - Ignore image contents. <br />
+            The default mode is @b raw, meaning that the encoded image data must match exactly.
+            Note: When the mode is set to @b image, the effective image data of individual mip
+            levels, layers, and faces is compared texel block by texel block after applying
+            one or more of the following transformations, if needed: <br />
+            \- If the texture is supercompressed (e.g. using Zstandard or ZLIB), the images
+               are deflated before comparison. <br />
+            \- If the texture is encoded using BasisLZ, the images are transcoded to
+               R8G8B8A_UNORM before comparison. <br />
+            For block compressed textures and textures encoded using UASTC the texel blocks are
```
MarkCallow, line 850: I think UASTC images should be transcoded too. Then people can compare UASTC and BasisLZ compression after the fact.

tools/ktx/command_compare.cpp
```diff
@@ -0,0 +1,2465 @@
...
+                in the basic DFD block. <br />
+            @b unknown - Ignore any unrecognized DFD blocks. <br />
+            @b extended - Ignore all DFD blocks except the basic DFD block. <br />
+            @b none - Do not ignore any DFD blocks. <br />
+            The default mode is @b none, meaning that all DFD entries will be compared.
+        </dd>
+        <dt>\--ignore-bdfd-bytesplane</dt>
+        <dd>Ignore BDFD bytesPlane values. Useful when comparing supercompressed files with
+            non-supercompressed ones, as bytesPlane is set to zero for supercompressed files.</dd>
+        <dt>\--ignore-metadata all | &lt;keys&gt; | none</dt>
```
MarkCallow, line 901: all | key<,...> | none

tools/ktx/command_compare.cpp
```diff
@@ -0,0 +1,2465 @@
...
+        <dd>Controls the comparison of metadata (KVD) entries. Possible options are: <br />
+            @b all - Ignore all metadata entries. <br />
+            @b &lt;keys&gt; - Ignore the specified comma separated list of metadata keys. <br />
```
MarkCallow, line 904: key&lt;,...&gt; - Ignore ...

MarkCallow (changes requested):
`ktx help compare` produces

```
ktx help fatal: Invalid command specified: "compare". See 'ktx help --help'.
```

I think we need header lines (at least in text output) like
```
--- input-file1
+++ input-file2
```
to easily see what the minuses and pluses refer to.

How much work will it be to add color-coding like git diff?

I would like an option to print the name of the file being processed here when the files differ and in `validate` when a file is invalid. It is useful when running a script over multiple files. See `grep -H` for example. How would such be handled in the JSON output and how much work will it be. I realize this is outside the current scope but I want to propose it to the TSG.

aqnuep:
> `ktx help compare` produces
> 
> ```
> ktx help fatal: Invalid command specified: "compare". See 'ktx help --help'.
> ```

Thanks for the feedback, I'll check what's going on.

> I think we need header lines (at least in text output) like
> 
> ```
> --- input-file1
> +++ input-file2
> ```
> 
> to easily see what the minuses and pluses refer to.
> 
> How much work will it be to add color-coding like git diff?
> 
> I would like an option to print the name of the file being processed here when the files differ and in `validate` when a file is invalid. It is useful when running a script over multiple files. See `grep -H` for example. How would such be handled in the JSON output and how much work will it be. I realize this is outside the current scope but I want to propose it to the TSG.

The text output of ktx compare is not supposed to be an actual diff-like output, as it wouldn't work in general for all the types of differences the tool can output. Not to mention that it would require significantly more effort to include context, etc.

Outputting the filenames for every difference would be possible, but it would make things extremely verbose as well. Also, per above, the output is "context-free", and does not output matching parts, hence every single diff would have to be wrapped with the filename headers, which would be highly undesirable.

Color coding would be possible, but you'd have to consider output support for ansi-color (terminal support and not including colors when output is captured in a file stream).

Everything is doable, but the output format has been out there for review for months, and at this point in time we do not have sufficient time to implement these additional requests, even if we'd agree to do them.

tools/ktx/command.h
```diff
@@ -364,6 +365,10 @@ class InputStream {
 public:
     InputStream(const std::string& filepath, Reporter& report);
 
+    const std::string& str() {
```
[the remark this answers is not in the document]
aqnuep: `str()` is a standard name for to-string conversion in C++ STL, but I can rename this to Java-like `toString`, or something, if needed.

tools/ktx/command_extract.cpp
```diff
@@ -412,16 +412,16 @@ void CommandExtract::executeExtract() {
         const auto imageHeight = std::max(1u, texture->baseHeight >> levelIndex);
         const auto imageDepth = std::max(1u, texture->baseDepth >> levelIndex);
 
-        for (uint32_t faceIndex = 0; faceIndex < texture->numFaces; ++faceIndex) {
-            if (options.fragmentURI.facial.is_undefined() ?
-                    faceIndex != 0 :
-                    !options.fragmentURI.facial.contains(faceIndex))
+        for (uint32_t layerIndex = 0; layerIndex < texture->numLayers; ++layerIndex) {
```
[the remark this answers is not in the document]
aqnuep: Not really, as the output names had suffices. I just unified the order of traversal across the tools for consistency.

tools/ktx/command_compare.cpp
aqnuep: The doxygen source and CLI help needed duplication (we still removed quite a bit of duplication in the original effort, but there's no way to merge these two).

The indentation of `--help` output is driven by cxxopts, so we do not have direct control over it.

tools/ktx/command_compare.cpp
aqnuep: It was an explicit request from @lexaknyazev to treat them as block-compressed formats.

BasisLZ needs transcoding (something that we suggested adding support for), because it does not have a block layout, as the supercompression is part of the universal data format itself, hence it is an exception.

tools/ktx/command_compare.cpp
aqnuep: The format `key<,...>` does not make sense, `<x>` means in CLI option format that it's a named value rather than a literal, `[x]` is for optional arguments, at most it should be something like `<key>[,<key>...]`, but that would be unnecessarily verbose IMO, as the description clearly indicates it is supposed to be a comma separated list.

tools/ktx/command.h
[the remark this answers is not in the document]
MarkCallow: Let's stick with the standard name.

tools/ktx/command_extract.cpp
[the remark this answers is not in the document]
MarkCallow: Good.

tools/ktx/schema/info.json
```diff
@@ -249,6 +249,20 @@
                         "VK_FORMAT_ASTC_10x10_SFLOAT_BLOCK_EXT",
                         "VK_FORMAT_ASTC_12x10_SFLOAT_BLOCK_EXT",
                         "VK_FORMAT_ASTC_12x12_SFLOAT_BLOCK_EXT",
```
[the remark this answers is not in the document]
MarkCallow: With what? Older versions of the tools? Old output files people have?

tools/ktx/command_compare.cpp
MarkCallow: Is it possible to change the spacing without modifying the fmt code, e.g, a compile option, maybe in a more recent version of fmt?

I am aware of the duplication. I was not aware at first where the above text was coming from.

tools/ktx/command_compare.cpp
MarkCallow: Then I am misunderstanding "effective image data". What is the difference between `raw` and `image` then for block-compressed textures or UASTC?

MarkCallow:
> The text output of ktx compare is not supposed to be an actual diff-like output, as it wouldn't work in general for all the types of differences the tool can output. Not to mention that it would require significantly more effort to include context, etc.

Maybe I confused you by using 3 pluses and minuses. I know it is not actual diff-like output. I am not looking for context. I am looking for an indication of which file corresponds to the - and which to the +. Basically just printing 2 lines at the top of the output.

> Outputting the filenames for every difference would be possible, but it would make things extremely verbose as well. Also, per above, the output is "context-free", and does not output matching parts, hence every single diff would have to be wrapped with the filename headers, which would be highly undesirable.

On reflection I think having a list of pairs of files to compare by way of a shell script loop is unlikely so this is not needed. To be clear, I was not looking for a filename at every difference, just once at the start of the output.

> Color coding would be possible, but you'd have to consider output support for ansi-color (terminal support and not including colors when output is captured in a file stream).

Indeed. I am not asking you to add color-coding. I am asking for an estimate of the work so I can decide if it is something we can pursue.

> Everything is doable, but the output format has been out there for review for months, and at this point in time we do not have sufficient time to implement these additional requests, even if we'd agree to do them.

Sometimes things only become apparent once you start using a tool.

tools/ktx/ktx_main.cpp
```diff
@@ -163,6 +163,7 @@ KTX_COMMAND_BUILTIN(ktxEncode)
 KTX_COMMAND_BUILTIN(ktxTranscode)
 KTX_COMMAND_BUILTIN(ktxInfo)
 KTX_COMMAND_BUILTIN(ktxValidate)
+KTX_COMMAND_BUILTIN(ktxCompare)
```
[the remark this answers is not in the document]
MarkCallow: This and `ktx help compare` are fixed but `compare` still does not appear in the output of `ktx help`.

tools/ktx/command_compare.cpp
```diff
@@ -0,0 +1,2465 @@
...
+               with --ignore-index all or --ignore-index level. <br />
+            \- DFD section differences can be ignored with --ignore-dfd all or --ignore-dfd
+               all-except-color-space. <br />
+            \- BDFD bytesPlane value differences can be ignored with --ignore-bdfd-bytesplane. <br />
+            \- SGD section differences can be ignored with --ignore-sgd all or --ignore-sgd
+               payload. <br />
+        </dd>
+        <dt>\--per-pixel-output all | &lt;number&gt; | none</dt>
+        <dd>Controls whether per pixel / texel block difference output is generated when
+            --content is set to @b image:
```
MarkCallow, line 868: `<br />` is missing from the end of this line.

MarkCallow:
The man page needs an explanation of the texel bock diff output. The following is from a comparison of 2 R8G8B8A8 cubemaps. My questions, which should be answered in the documentation, follow:

```
+Mismatch in level 0 layer 0 face 0
  Coordinates: 2, 0, 0
    Image byte offset: 0x8
    File byte offset: 0x1100
-    Packed: 0x08, 0x09, 0x14, 0x03
+    Packed: 0x08, 0x09, 0x0a, 0x0b
-    Channels: 8, 9, 20, 3
+    Channels: 8, 9, 10, 11
```

1. Coordinates presumably is image coordinates. Are they based on KTXorientation or do they assume top down orientation?
2. What is the meaning of "Packed" here?
3. What are the Channels? According to the DFDs both files have channel ids of 0, 1, 2 and 15 (RGBA).

tools/ktx/schema/info.json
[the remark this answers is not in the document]
aqnuep: Older tool versions. In general I've been avoiding any backwards-incompatible change to the JSON schemas, but, of course, if that's not a concern, we can remove these.

tools/ktx/command_compare.cpp
aqnuep: This is not coming from the fmt library, but from the cxxopts library we use for CLI option handling.

I'll check if there's any possibility to configure this.

tools/ktx/command_compare.cpp
aqnuep: Raw is not doing per texel block comparison, but a simple memcmp of the entire image payload. Image comparison does per texel block comparison and for UASTC the texel blocks are individual UASTC compressed blocks.

aqnuep:
> The man page needs an explanation of the texel bock diff output. The following is from a comparison of 2 R8G8B8A8 cubemaps. My questions, which should be answered in the documentation, follow:
> 
> ```
> +Mismatch in level 0 layer 0 face 0
>   Coordinates: 2, 0, 0
>     Image byte offset: 0x8
>     File byte offset: 0x1100
> -    Packed: 0x08, 0x09, 0x14, 0x03
> +    Packed: 0x08, 0x09, 0x0a, 0x0b
> -    Channels: 8, 9, 20, 3
> +    Channels: 8, 9, 10, 11
> ```
> 
> 1. Coordinates presumably is image coordinates. Are they based on KTXorientation or do they assume top down orientation?
> 2. What is the meaning of "Packed" here?
> 3. What are the Channels? According to the DFDs both files have channel ids of 0, 1, 2 and 15 (RGBA).

1. Coordinates are in texel coordinates. They are not affected by KTXorientation, they are the "raw" coordinates.
2. Packed is the array of individual packed data elements (each element has a size of `typeSize`)
3. Channels is the array of channel values in integer (for `INT` and `NORM` formats) or float (for `FLOAT` formats) format, and they are output in the same order as the channels appear in the DFD, as discussed and agreed over email (i.e. not always in RGBA order) in order to work as expected for more exotic formats like YUYV which have multiple components of the same type

aqnuep:
I've updated the man page to include information above about the output format.

tools/ktx/ktx_main.cpp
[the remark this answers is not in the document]
aqnuep: Hmm, let me double-check. I thought I verified it.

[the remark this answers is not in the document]
aqnuep: Ah, yes, I missed the man text for the main ktx command. It should be fixed now.

tools/ktx/schema/info.json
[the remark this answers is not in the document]
MarkCallow: Thanks. No need to remove them at this point though people generally seem to update their installations.

tools/ktx/command_compare.cpp
MarkCallow: I meant `cxxopt`. Please check. Its a more general fix than just this PR so I'm not going to hold up merging waiting for an answer.

tools/ktx/command_compare.cpp
MarkCallow: I thought ""effective image data" was the decoded data where possible so images of different but compatible formats could be compared. That is something we can possibly add later.

So the user visible difference is that you'll get told which layer, levels and faces differ instead of just that he images differ. Correct?

tools/ktx/command_compare.cpp
MarkCallow: Somehow a reply I typed earlier today has gone missing. I confused the meanings of `<>` and `[]`. I like it to be clear from a quick scan of the option without having to read the descriptions. How about `<key>[,...]` to reduce the verbosity. I sure I've seen this kind of abbreviation used elsewhere for repeatable arguments.

tools/ktx/command_compare.cpp
```diff
@@ -0,0 +1,2482 @@
...
+            The default mode is @b none, meaning that SGD sections will be always compared.
+            Note: --ignore-sgd payload can be used to compare BasisLZ SGD headers without
+            expecting an exact match for the individual SGD payload sections.
+        </dd>
+    </dl>
+    @snippet{doc} ktx/command.h command options_generic
+
+    Reports about individual texel block differences include the following information: <br />
+    <ul>
+        <li>The level, layer, and face index of the image where the difference was found.</li>
```
MarkCallow, line 923: layer, level and face is the order I normally think of.

tools/ktx/command_compare.cpp
```diff
@@ -0,0 +1,2482 @@
...
+            @b payload - Ignore any unrecognized SGD section and the payload of any known SGD section. <br />
+            @b none - Do not ignore the SGD section. <br />
+            The default mode is @b none, meaning that SGD sections will be always compared.
+            Note: --ignore-sgd payload can be used to compare BasisLZ SGD headers without
+            expecting an exact match for the individual SGD payload sections.
+        </dd>
+    </dl>
+    @snippet{doc} ktx/command.h command options_generic
+
+    Reports about individual texel block differences include the following information: <br />
```
MarkCallow, line 921: Reports about individual texel block differences, which are output if --content is set to @b image, include...

tools/ktx/command_compare.cpp
aqnuep: I've checked cxxopts. The good news is there's a single constant that affects the maximum indentation the descriptions of CLI options will have (maximum, as if all option names are shorter, then smaller indentation is used).

The bad news is that it is only modifiable by actually changing our copy of the cxxopts source code.

Do you want me to do that?

FYI, the default value is 30, meaning that the maximum indentation of descriptions is 32 (+2 because of the indentation of CLI option names themselves).

tools/ktx/command_compare.cpp
```diff
@@ -0,0 +1,2482 @@
...
+    <ul>
+        <li>The level, layer, and face index of the image where the difference was found.</li>
+        <li>The texel coordinates within the image where the difference was found, if --content
```
MarkCallow, line 924: Remove "if --content is set to @b image," here and below as texel block information is only printed when that is the case.

tools/ktx/command_compare.cpp
aqnuep: > I thought ""effective image data" was the decoded data where possible so images of different but compatible formats could be compared. That is something we can possibly add later.

Not in the case of the current image comparison mode. Besides this UASTC specific requirement, a general requirement for this comparison mode was to compare the packed data values, because we wanted to see differences, even if the decoded channel values otherwise match (e.g. two different representations of NaNs for floats, or two different representations of UASTC blocks that otherwise lead to the same transcoded/decoded values).

For a comparison mode that only care about the "color" values (potentially with tolerance) that may also allow comparing different formats (which is not currently supported), it is expected that we'd need to transcode (and decode GPU BC formats, for example), and do the comparison of the decoded channel values instead (potentially even with swizzling applied, if needed). But that's something to decide in the future.

Currently, the two modes are: raw, to compare image data as encoded in the file, and image, to compare packed data elements on a per texel block basis.

> So the user visible difference is that you'll get told which layer, levels and faces differ instead of just that he images differ. Correct?

Yes, for raw you'll only get image-level difference reports on a per layer/level/face basis.

tools/ktx/command_compare.cpp
aqnuep: Sure, fine by me. I still think it's unconventional, but certainly better.

aqnuep: Updated.

tools/ktx/command_compare.cpp
aqnuep: That's not actually how KTX stores these though, as KTX stores things in level-major order.

It's also the convention in graphics APIs, albeit historically GPUs used both level-major and layer-major representations.

tools/ktx/command_compare.cpp
aqnuep: I've used this phrasing, because the level, layer, and face is also output when --content is set to raw.

tools/ktx/command_compare.cpp
aqnuep: I took another stab at this section of the documentation. Please check the latest version.

MarkCallow: The rewrite is much better. Thanks.

MarkCallow:
I've merged https://github.com/KhronosGroup/KTX-Software-CTS/pull/22 though I haven't yet deleted the branch. Please update the reference here, fix the conflicts and mark this ready so I can merge it today.

aqnuep:
I've rebased the PR.

I also had to apply a fix in the dfd2vk code to handle the weird convention the KTX 2.0 spec defines for `D24_UNORM_S8_UINT` that uses OpenGL conventions.

MarkCallow (approved):

aqnuep:
Thank you, @MarkCallow for helping getting this through!

PR #873, opened 2024-03-20
Clarify description of --format argument for create when used with --encode

### LINKED ISSUE #871: Converting transfer functions (using assign-oetf and convert-oetf)
Based on the [documentation](https://github.khronos.org/KTX-Software/ktxtools/ktx_create.html), I would expect the following to work:

```
ktx create --format=R8G8B8_SRGB --assign-oetf=srgb --convert-oetf=linear foo.png foo.ktx2
```

(Converting the sRGB png into a linear color space ktx2 file). It's pretty clear that the documentation says if the oe transfer function assigned is different from the convert transfer function, that it gets converted.

However, I get the error message:

```
ktx create fatal: Invalid value to --convert-oetf "linear" for format "R8G8B8_SRGB". Transfer function must be sRGB for sRGB formats. See 'ktx create --help'.
```

Is my understanding incorrect or is this a bug? What am I missing?

aqnuep:
Fixes #871.

### DIFF
--- a/tools/ktx/command_create.cpp
+++ b/tools/ktx/command_create.cpp
@@ -97,7 +97,7 @@ struct OptionsCreate {
                 (kFormat, "KTX format enum that specifies the image data format."
                     " The enum names are matching the VkFormats without the VK_FORMAT_ prefix."
                     " The VK_FORMAT_ prefix is ignored if present."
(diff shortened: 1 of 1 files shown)


aqnuep:
CTS change in PR https://github.com/KhronosGroup/KTX-Software-CTS/pull/23

MarkCallow (approved):

PR #874, opened 2024-03-22
Add JS bindings for full libktx

### LINKED ISSUE #288: ktxTexture should expose its data via JS bindings
Currently, there's no way to access original or transcoded texture data via JS. This makes the libKTX unusable in node environments.

aqnuep:
This PR has been repurposed from adding the Phasmatic bindings for partial write functionality into adding bindings for nearly the full libktx functionality and a test for all the new functionality.

The PR makes the following minor API changes to libktx:
* Adds `ktxTexture_GetLevelSize` function.
* Makes the parameters to `ktxTexture[12]_Create` `const`.
* Adds `ktxTexture2_{GetPrimaries_e, SetOETF and SetPrimaries}` functions.

It adds the following functions to the JS binding while modernizing and greatly improving the binding code:
* A constructor equivalent to `ktxTexture_Create`.
* `createCopy`
* `findKeyValue`
* `getDataSize`
* `getOETF`
* `getPrimaries`
* `getImage` for retrieving an image from the texture. Fixes #288.
* `setImageFromMemory`
* `compressAstc`
* `compressBasis`
* `deflateZstd`
* `deflateZLIB`
* `addKVPairString` and `addKVPairByte`
* `deleteKVPair`
* `setOETF`
* `setPrimaries`
* `writeToMemory`

It makes available two bindings. First is a read-only binding, `libktx_read.js` which has the previously available functions plus `findKeyValue`, `getDataSize`, `getOETF`, `getPrimaries` and `getImage`. The second, using the original name `libktx.js`, has functions for creating new KTX texture objects and writing them to files.

The JS `ktxTexture` class has been renamed `texture` as typical usage is with an Emscripten module instance name of `ktx` and `new ktx.ktxTexture()` looks weird. Newly added classes do not have `ktx` prefixes.

Compared to the Phasmatic bindings it:
* Removes `compressBasisU` which was bundling calls to `ktxTexture2_CompressBasisEx`, `ktxTexture2_DeflateZstd`, `ktxTexture2_WriteToMemory` and `ktxHashList_addKVPair`. Instead each function now has a binding and can be called by the JS client.
* Removes `createFromBuffer` which was using parameters passed from Javascript to determine the VKFormat and size of the texture then creating it with `ktxTexture2_Create` and finally loading an image from a passed ArrayBufferView into the texture via `ktxTexture_SetImageFromMemory`. Instead a `ktxCreateInfo` can be initialized in the JS code which can call `new ktxTexture` to create the texture then call `setImageFromMemory` to put the image into the texture object.


CMakeLists.txt
```diff
@@ -914,6 +921,44 @@ if(EMSCRIPTEN)
         COMPONENT ktx_js
     )
 
+    add_executable( ktx_js_read interface/js_binding/ktx_wrapper.cpp )
+    target_link_libraries( ktx_js_read ktx_read )
+    target_include_directories(
+        ktx_js_read
+    PRIVATE
+        ${CMAKE_CURRENT_SOURCE_DIR}/other_include
```
MarkCallow, line 929: Why are other_include and lib needed now. They weren't for the previous read-only build.

CMakeLists.txt
```diff
@@ -914,6 +921,44 @@ if(EMSCRIPTEN)
...
+        ${CMAKE_CURRENT_SOURCE_DIR}/lib
+        $<TARGET_PROPERTY:ktx_read,INTERFACE_INCLUDE_DIRECTORIES>
+    )
+    target_link_options(
+        ktx_js_read
+    PUBLIC
+        ${KTX_EMC_LINK_FLAGS}
+        "SHELL:-s EXPORT_NAME=LIBKTX_READ"
```
MarkCallow, line 937: I'd like to avoid duplicating these SHELL: items. Can you put them in a CMake variable to be used in both builds?

interface/js_binding/ktx_wrapper.cpp
```diff
@@ -22,6 +49,136 @@ namespace ktx
         texture(texture&) = delete;
         texture(texture&& other) = default;
 
+        static texture createFromBuffer(const emscripten::val& data, int width, int height, int comps, bool srgb) {
```
MarkCallow, line 52: I'm now recalling some of the discussion about merging this work back to KTX-Software. This function meets the needs of the glTF-Compressor which only needs to make BasisU textures for KHR_texture_basisu but is far from complete as a general compression tool. That is why Phasmatic did not integrate it.

This function needs to have a format argument and many other parameters so it can create textures with various formats and shapes. Or perhaps there should be multiple functions, one for 2D textures like this is and additional functions for arrays, cube maps, 3d. We also need to expose the deflate functions.

We need to have a discussion.

CMakeLists.txt
aqnuep: Because of the new includes of `vk_format_enum.h` and GL headers by `ktx_wrapper` introduced by Phasmatic's changes.

CMakeLists.txt
aqnuep: I was trying to deduplicate the definitions of the read-only vs read-write versions, but didn't turn out to be any cleaner. If you really want to, I can do that nonetheless.

MarkCallow: I found the proper way to do this, setting them as INTERFACE items so they are automatically set for dependencies. See commit 5d13ab5. You'll pick it when merging or rebasing main.

interface/js_binding/ktx_wrapper.cpp
MarkCallow: I want to capture this work in main as otherwise it will bit rot. I will review the changes again and mark what I think needs to be done to build only the ktx-read wrapper by default and to not release it the ktx wrapper.

CMakeLists.txt
```diff
@@ -883,8 +883,15 @@ if(EMSCRIPTEN)
     )
 
     add_executable( ktx_js interface/js_binding/ktx_wrapper.cpp )
```
MarkCallow, line 885: Please add a KTX_FEATURE_JS_WRITE option only visible `if(EMSCRIPTEN)` and defaulting to OFF. The documentation string should say something like "JS wrapper with minimal write functions as used by glTF-Compressor.". Only add ktx_js when set to ON.

CMakeLists.txt
MarkCallow: Will it help to bracket these includes and any other write-related code in `ktx_wrapper` with`#if KTX_FEATURE_WRITE`?

CMakeLists.txt
```diff
@@ -1307,6 +1352,7 @@ elseif(WIN32)
 elseif(EMSCRIPTEN)
     set(CPACK_GENERATOR ZIP)
     set(CPACK_ARCHIVE_KTX_JS_FILE_NAME "${CMAKE_PROJECT_NAME}-${KTX_VERSION_FULL}-Web-libktx")
```
MarkCallow, line 1354: Only include this if `KTX_FEATURE_JS_WRITE` is ON.

interface/js_binding/ktx_wrapper.cpp
MarkCallow, line 52: Bracket this function with `#if KTX_FEATURE_WRITE`.

MarkCallow (commented):
I think the changes in these comments are all that are needed so ktx_js is only built when the corresponding feature is on.

interface/js_binding/ktx_wrapper.cpp
aqnuep: Isn't this an upload path? Why hide it behind `KTX_FEATURE_WRITE`?

aqnuep:
> Will it help to bracket these includes and any other write-related code in ktx_wrapper with#if KTX_FEATURE_WRITE?

For some reason I cannot reply on this comment, so replying here: what do you mean by that?

CMakeLists.txt
aqnuep: Ah, this reply from you also appears here, just noticed.

Maybe that works, but I'd have to check.

interface/js_binding/ktx_wrapper.cpp
MarkCallow: It is for creating a new texture rather than loading one from a KTX file so I'd always considered this part of the write functionality but now I see that the native equivalent is defined in `texture2.c` not `writer2.c`. For equivalence, we should not bracket this function.

CMakeLists.txt
aqnuep: Okay, considering that `createFromBuffer` depends on the GL header and in the other comment thread you agreed to keep that for read-only, I don't think we can change this.

CMakeLists.txt
MarkCallow: Discussion has been helpful. For example I was unaware the cmake_dependent_option was introduced in 3.22. Our CMakeLists.txt uses it a lot. I've just created PR #904 to bump the minimum version to 3.22 after checking that all our environments have at least that.

But yes. I'll take over this PR. On reflection I don't think `KTX_FEATURE_JS_WRITE` is needed.  We just need to comment out the addition of the read-write version to the release archive so we don't release it until such time as has fully fledged write features.

MarkCallow:
@abasilak, @ViNeek please review this PR. Please pay particular attention to the following:

1. Will there be any difficulties using these bindings for the glTF-Compressor?
2. I have changed the name of the factory function for creating the ktx module from `LIBKTX` to `createKtxModule` (and `createKtxReadModule`). I am not planning to provide the old name as an alias because it is a simple change for users to make. What is your opinion.
3. In the test sample I am now setting `window.ktx = <module returned from createKtxModule>` so now I have code like `var texture = new ktx.ktxTexture` which is feels daft.  I'd like to remove the ktx prefix from all the times that currently have it. What do you think? Note that the all the c++ code for the binding is in a `ktx` namespace. The extra prefixes really are superfluous in my opinion. I would, if I can figure out how, provide the old names as aliases. 
4. Point out any errors or places for improvement in the code.

I would like the glTF-Compressor to be updated to use this binding. Is there any room in your existing Khronos contract for that work? Better reply to me privately on this question.

MarkCallow:
@ViNeek thank you for your review and comments. I have made one final (fingers crossed) change. I have changed the enumerator names to be the same as the libktx names minus the `ktx{,_}` prefixes and `_e` suffixes for consistency and to make it easier for people to recall what the JS name will be. Please take a look at 6966f6c.

Issue #875, opened 2024-03-22
The JNI binding for input swizzling is broken


javagl:
The [`setInputSwizzle` method of the Java `KtxBasisParams` class](https://github.com/KhronosGroup/KTX-Software/blob/5414bd0ebfdfdf4ecb8c613a6edb21372303c2b9/interface/java_binding/src/main/java/org/khronos/ktx/KtxBasisParams.java#L117) receives an array of Java `char` values. (The same for `KtxAstcParams` - I'm focussing on one case here).

In the JNI layer, this swizzling information is supposed to be transferred into the [`char inputSwizzle[4];` array of the `ktxBasisParams` structure](https://github.com/KhronosGroup/KTX-Software/blob/5414bd0ebfdfdf4ecb8c613a6edb21372303c2b9/include/ktx.h#L1341)

This is done in [this line](https://github.com/KhronosGroup/KTX-Software/blob/e085435658e70cc825ea705c23807cbf4d14bf3e/interface/java_binding/src/main/cpp/libktx-jni.cpp#L148), by casting the input value to a `jbyteArray`. 

But ... the input it's not a `jbyteArray`, it's a `jcharArray`. 

The point is: The size of `char` in Java is 16 bits. The size of `char` in C/C++ is 8 bits. (Usually. Who really knows?)

This means that the native `inputSwizzle` array receives "garbage" data. In the best case, the first two Java 16 bit `char` values might be interpreted as four C/C++ 8 bit `char` values. But in fact, when compiling in Debug mode, calling the `setInputSwizzle` function will trigger a debug assertion ("String subscript out of range", in `xstring`).

The following is a small test case: It generates an image that contains a red, green, blue, and white "stripe", compresses it, and writes it out.

```java
package de.javagl.jktx;

import java.io.IOException;
import java.nio.file.Path;
import java.nio.file.Paths;

import org.khronos.ktx.KtxBasisParams;
import org.khronos.ktx.KtxCreateStorage;
import org.khronos.ktx.KtxErrorCode;
import org.khronos.ktx.KtxTexture2;
import org.khronos.ktx.KtxTextureCreateInfo;
import org.khronos.ktx.VkFormat;

public class KtxSwizzleTest
{
    static
    {
        System.loadLibrary("./lib/ktx-jni");
    }

    private static int sizeX = 64;
    private static int sizeY = 64;
    
    public static void main(String[] args) throws IOException
    {
[...]

PR #876, opened 2024-03-22
Properly convert Java char to C char in inputSwizzle

javagl:
Fixes https://github.com/KhronosGroup/KTX-Software/issues/875 

As detailed in the issue: The Java `char[]` array for the input swizzle was casted to a `jbyteArray` in the JNI layer, in order to write it to the C/C++ `char[4]` array. This cast was invalid and caused data corruption.

There could be two ways of tackling this:

- One could change the type of the `inputSwizzle` in the Java layer to `byte[]` (which has 8 bits, and would make the cast valid). This would be a breaking change, and may require the inconvenient casting at the call site like
  `p.setInputSwizzle(new byte[]{ (byte)'b', (byte)'r', (byte)'g', (byte)'a' });`
- One could fetch the Java `char[]` array as a `jcharArray`, and cast the elements to (C/C++) `char` individually

This PR takes the second option (but we could do this either way, depending on the preferences from others).

With this change, running the test/example program that is given in the issue produces the following results (showing the PNG files here - the KTX ones are attached below)

Without calling `setInputSwizzle`, the result is this:

output_swizzle_unswizzled

(`output_swizzle_unswizzled.png`)

---

With `setInputSwizzle(new char[]{ 'r', 'g', 'b', 'a' });` (which should not change anything), the result is this:

output_swizzle_rgba_fixed

(`output_swizzle_rgba_fixed.png`)

---

With `setInputSwizzle(new char[]{ 'b', 'r', 'g', 'a' });`, actually swizzling things around, the result is this:

output_swizzle_brga-fixed

(`output_swizzle_brga-fixed.png`)

---

The KTX and PNG files are attached here:

[KTX swizzle results 2024-03-22.zip](https://github.com/KhronosGroup/KTX-Software/files/14726386/KTX.swizzle.results.2024-03-22.zip)

This PR does not involve any unit tests. I wouldn't know how to add one _exactly_ - maybe just comparing the output to a "golden" reference file?


MarkCallow:
Thank you for this. I will review it as soon as I can. I have other PRs to review first.

In the meantime, please add a test for swizzle to `interface/java_binding/src/test/java/org/khronos/ktx/test`.

javagl:
I have added a test that _calls_ `setInputSwizzle` (basically the same as the snippet that was posted in the issue).

But _calling_ the function did alread work before this fix. I don't see a way to check that the output is the desired result. I created a "reference image", and tried to compare the image that is generated in the test to the reference image, but ... _what_ exactly should be compared there? The file contents differs, because the file contains version information like `4.3.0-alpha3~2` and so. Calling `getData` returns ... _something_, but ... I don't know what it is, and apparently, it cannot be compared (doing _the same thing twice_ causes different return values for `getData`...).

There does not seem to be a way to check whether the result image is the "broken" one, 

output_swizzle_brga-broken

or the one where the RGBA -> BRGA swizzle has properly been applied...

output_swizzle_brga-fixed-small

javagl:
I have added a test to compare the result of `getData()` to the data that is obtained from a reference texture, stored as `/tests/testImages/swizzle-brga-reference.ktx`. 

(I tried this initially, and it didn't work as expected, but this was caused by an issue in the `KtxTextLibraryLoader`, mentioned in [this comment](https://github.com/KhronosGroup/KTX-Software/issues/877#issuecomment-2016569030). I'll try to keep track of that and see whether it can easily be resolved)

interface/java_binding/src/test/java/org/khronos/ktx/test/KtxTexture2Test.java
```diff
@@ -227,4 +228,72 @@ public void testCreate() {
...
+		// the former R channel becomes the G channel
+		// the former G channel becomes the B channel
+		// the former A channel remains the A channel
+		KtxBasisParams p = new KtxBasisParams();
+		p.setUastc(false);
+		p.setInputSwizzle(new char[] { 'b', 'r', 'g', 'a' });
+		t.compressBasisEx(p);
+
+		// Obtain the texture data, and compare it to the
+		// expected data of the reference texture
```
MarkCallow, line 268: Since you are creating the texture in the test, I'd prefer not to need a reference file. You can transcode the swizzled `t` to `KTX_TTF_RGBA32` and compare those pixels with the pixels in `rgba`. Given the full on colors you are using I don't think compression & transcode will change them.

interface/java_binding/src/main/cpp/libktx-jni.cpp
```diff
@@ -97,12 +97,14 @@ void copy_ktx_astc_params(JNIEnv *env, jobject params, ktxAstcParams &out)
...
-        0,
-        4,
-        reinterpret_cast<jbyte*>(&out.inputSwizzle)
-    );
+    jobject inputSwizzleObject = env->GetObjectField(params, inputSwizzle);
+    jcharArray inputSwizzleArray = static_cast<jcharArray>(inputSwizzleObject);
+    jchar *inputSwizzleValues = env->GetCharArrayElements( inputSwizzleArray, NULL);
+    for (int i=0; i<4; i++)
+    {
+      out.inputSwizzle[i] = static_cast<char>(inputSwizzleValues[i]);
```
MarkCallow, line 105: A static_cast will convert a 16-bit jchar to an 8-bit char? Seriously!

MarkCallow (changes requested):
One question to be answered and 1 change to make.

MarkCallow:
The test failed on GNU/Linux and macOS but apparently passes on Windows. I'm surprised by that.

interface/java_binding/src/test/java/org/khronos/ktx/test/KtxTexture2Test.java
javagl: I'd also like to avoid that reference file. (One reason: The `tests/testimages` directory already contains >100 files, without any information about what these files are, where they come from, and what they are used for. I'd like to avoid throwing more images on that pile...)

And what you suggested might ... _nearly_ work: From a quick test, converting the resulting data to `KTX_TTF_RGBA32` gave the following input/output pxiel values:
```
Pixel 0 in input  255,   0,   0, 255
Pixel 0 in output   2, 255,   2, 255
...
Pixel 256 in input    0, 255,   0, 255
Pixel 256 in output   0,   0, 253, 255
...
Pixel 512 in input    0,   0, 255, 255
Pixel 512 in output 253,   0,   0, 255
```
The swizzling is applied correctly, but there is a _slight_ deviation in the formerly "pure" colors, of `+/- 2` in some components. I don't know where this might come from - maybe something about SRGB, or it's a compression artifact, or whatnot. If you have an idea what might be causing this (or how to avoid this), I could change it accordingly. 

Regarldless of that, the test _could_ be changed to not use a reference image, by comparing the result of converting the image to `KTX_TTF_RGBA32` to pixel values that are created _programmatically_ (using the same functions that are used for creating the input image).

This could then be 
- A **fixed** comparison to the values that I currently see in the output (`2, 255,   2`)
- Or a comparison with `expected=(0, 255, 0)` **but** with some "epsilon" of maybe `+/-5` to account for the slight changes...

Any preference?

interface/java_binding/src/main/cpp/libktx-jni.cpp
javagl: At this point, this is basically the same as a `out = (char)in` C-Style cast. 

One could probably throw in some Unicode character into the swizzle array, where the _lower-order_ bits correspond to that of the `r` character. 

And we could think about certain forms of error checks here. For example, the KTX tools already do check the swizzle chars against the regex `"^[rgba01]{4}$"`, and we could insert similar checks at different places. This could happen in the JNI layer or (far more easily) in the Java layer. But error handling is one of the broader topics to be sorted out...

javagl:
> The test failed on GNU/Linux and macOS but passes on Windows. I'm surprised by that.

I'm a _little_ bit surprised by that as well. Looking at the failed build output:

```
[ERROR] Tests run: 11, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.248 s <<< FAILURE! - in org.khronos.ktx.test.KtxTexture2Test
[ERROR] testInputSwizzleBasisEx  Time elapsed: 0.014 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: array contents differ at index [0], expected: <49> but was: <77>
	at org.khronos.ktx.test.KtxTexture2Test.testInputSwizzleBasisEx(KtxTexture2Test.java:278)
```

It does compare the array obtained with `getData` there. (And I'm sure that the fact that it expected a `1` but found an `M` is just a coincidence). Now I'm wondering whether the contents of this data might _in any form or way_ depend on the _operating system_ with which the data is created.,,

(Maybe this becomes obsolete when the reference file is no longer used, but I'm still curious...)

interface/java_binding/src/main/cpp/libktx-jni.cpp
MarkCallow: Yes we need to resolve how to do error handling first. That should tell us whether to do it in JNI or Java. I'll mark this as resolved since it doesn't directly affect this PR.

interface/java_binding/src/test/java/org/khronos/ktx/test/KtxTexture2Test.java
MarkCallow: I prefer the fixed comparison as the test will not then be sensitive to changes in the encoder. You could also try `ktxTexture2_CompressAstcEx`. Perhaps that will have less artifacts.

MarkCallow:
> Now I'm wondering whether the contents of this data might in any form or way depend on the operating system with which the data is created.,,

Yes it does. See https://github.com/BinomialLLC/basis_universal/issues/60.

interface/java_binding/src/test/java/org/khronos/ktx/test/KtxTexture2Test.java
javagl: > You could also try ktxTexture2_CompressAstcEx. Perhaps that will have less artifacts.

Actually, in order to have proper test coverage, I can not only 'try' this (as an alternative), but actually have to test this _in addition_ to the current test. Both of them (basis and astc) are going through different code paths (even thought they are doing the same sort of conversion for the swizzling).

javagl:
@MarkCallow The test has been updated to no longer compare to a reference image. Instead, it is converting the texture to `KTX_TTF_RGBA32`, and comparing the data to the expected pixels, which are created "manually" in memory. The relevant part is https://github.com/KhronosGroup/KTX-Software/pull/876/files#diff-8ed75fc18c4588264650ce9f7dd4e693adb82c64a22caee3eb1de19cba383b05R278

As you have seen, there are follow-ups for this PR:

- This PR is only a "bugfix", strictly speaking
- Based on that, I had created https://github.com/KhronosGroup/KTX-Software/pull/879 (so when _this_ PR is merged, the one that adds the supercompression functions could be wrapped up and merged)
- Based on both of them, the PR https://github.com/KhronosGroup/KTX-Software/pull/886 aims at addressing all the cross-cutting things that I noticed (and that you brought up) while creating the first ones: Documentation, error handling, more missing functions, more tests.... This PR will probably still take a while: There are a few questions about how _exactly_ things will be addressed, and some of the points are nearly "open ended" (like documentation and tests), but I'll still try to bring it into a somewhat consistent state ASAP.

javagl:
A quick test suggests that the proposed approach does not work as expected. When implemented based on the pseudocode, the output will be
```
At 0 unswizzled 253,   0,   0, 255
At 0 swizzled     2, 255,   2, 255
At 0 expected     0, 253,   0, 255
...
At 256 unswizzled   2, 255,   2, 255
At 256 swizzled     0,   0, 253, 255
At 256 expected     2,   2, 255, 255
...
```
So apparently

- "compressing _with_ swizzling" and
- "compressing and _then_ swizzling"

produce different results, exactly in ths `+/- 2` range for some components.

Given that...

1. before this PR, swizzling did not work _at all_
2. before this PR, there was no test coverage for "pixel colors" _at all_

... I'm leaning towards not trying to make "this test perfect" **now**. Improving the tests and coverage (_including_ the subtle questions about slight color deviations) may be part of https://github.com/KhronosGroup/KTX-Software/pull/886 . 

If your concern is that this test might break unpredictably, I'll just remove the "pixel color comparison". Then we'd have a **working** swizzle functionality, even though it's not covered by a test, similar to a million other things that are 'working 🤞' but not covered by a test. 

And by the way: All this suggests for me that there is **zero** test coverage for the effects of swizzling on the native side either. One could certainly argue about the responsibility of the Java bindings for "pixel-perfect comparisons" here...

MarkCallow:
There are unittests for the function that performs the swizzle before submitting the data to the encoder and there are tests of the `--input-swizzle` option of `ktx create`. I think the issue you are seeing is due to how block compression works. Try this pseudo-code instead

```
byte input[] = createInput();
byte goldSwizzle[];
for (each pixel) {
    goldSwizzle[pixel] = swizzle(input[pixel]);
}

// Compress to basis
KtxTexture2 swizzled = compressWithSwizzle(input);
KtxTexture2  goldSwizzled = compressWithoutSwizzle(goldSwizzle);

// Transcode to RGBA32
byte swizzledRgba[] = swizzled.transcodeToRgba32();
byte goldRgba[] = goldSwizzled.transcodeToRgba32();

// Check each pixel of the swizzled output if it is
// exactly the same as the swizzled pixel in the
// output that was created without swizzling
for (each pixel) {
    swizzledRgba[pixel] = goldRgba[pixel]);
}
```

javagl:
Thanks @MarkCallow , that seems to work (at least locally - let's see whether the build passes...)

If the number of this kind of tests (which compare actual pixel values) increases, some of the required functionality could be moved into the `TestUtils` class, but that can be done in follow-up PRs.

javagl:
@MarkCallow I think this should be ready to merge, if you agree.

When this is merged, I'd wrap up https://github.com/KhronosGroup/KTX-Software/pull/879 (which should only be a small one then). 

For the "big" one, https://github.com/KhronosGroup/KTX-Software/pull/886 , I'll do some more testing/polishing , and ... we'll have to figure out a way how this could sensibly be "reviewed"....

MarkCallow (approved):

PR #879, opened 2024-03-23
Expose supercompression functions via JNI

javagl:
<sup>Updated:</sup>

This builds upon https://github.com/KhronosGroup/KTX-Software/pull/876 , which is already merged. 

It exposes the `deflateZstd` and `deflateZLIB` functions of the `ktxTexture2` class to the Java `KtxTexture2` class.

This is only a tiny part of the changes that are contained in https://github.com/KhronosGroup/KTX-Software/pull/886 , but could be merged independently.


interface/java_binding/src/main/cpp/libktx-jni.cpp
```diff
@@ -97,12 +97,14 @@ void copy_ktx_astc_params(JNIEnv *env, jobject params, ktxAstcParams &out)
     out.normalMap = env->GetBooleanField(params, normalMap);
     out.perceptual = env->GetBooleanField(params, perceptual);
 
-    env->GetByteArrayRegion(
-        static_cast<jbyteArray>(env->GetObjectField(params, inputSwizzle)),
-        0,
-        4,
-        reinterpret_cast<jbyte*>(&out.inputSwizzle)
-    );
+    jobject inputSwizzleObject = env->GetObjectField(params, inputSwizzle);
```
MarkCallow, line 100: Having the fix from #876 here is going to cause conflicts come merge time. Does this PR need the swizzle fix? If not, please remove it.

interface/java_binding/src/main/java/org/khronos/ktx/KtxTexture2.java
```diff
@@ -45,4 +45,30 @@ public static native KtxTexture2 createFromNamedFile(String filename,
     public static KtxTexture2 createFromNamedFile(String filename) {
         return createFromNamedFile(filename, KtxTextureCreateFlagBits.LOAD_IMAGE_DATA_BIT);
     }
+
+    /**
+     * Deflate the data in a {@link KtxTexture2} object using Zstandard.
+     *
+     * The texture's levelIndex, dataSize, DFD  and supercompressionScheme will
```
MarkCallow, line 52: The data pointer is also updated. Looks like an oversight in the documentation probably emanating from the native documentation. Please update the doc here and in native, if indeed it does not mention this.

interface/java_binding/src/test/java/org/khronos/ktx/test/KtxTexture2Test.java
```diff
@@ -227,4 +227,126 @@ public void testCreate() {
...
+		int scBefore = t.getSupercompressionScheme();
+		assertEquals(KtxSupercmpScheme.NONE, scBefore);
+
+		// Apply Zstd compression
+		t.deflateZstd(10);
+
+		// The supercompression scheme should now be ZSTD
+		int scAfter = t.getSupercompressionScheme();
+		assertEquals(KtxSupercmpScheme.ZSTD, scAfter);
+
```
MarkCallow, line 299: I suggest also inflating the texture again and comparing its data with `rgba`. Does `new` in java 0 initialize the memory or will it be random data.

If the UASTC compression causes this compare to fail, you can skip the `compressBasisEx` step. It is perfectly valid to deflate uncompressed textures.

---- EDIT ----

The native inflate functions are not exposed in the public API. Inflation is always done when a file is opened. (I did not think about something like Java testing when I made that choice. Native tests can always link to the function.) To make my suggestion work you will have to call t.writeToMemory then CreateFromMemory but didn't you tell me those functions aren't exposed in JNI. They need to be.

MarkCallow (changes requested):

interface/java_binding/src/main/cpp/libktx-jni.cpp
javagl: I'll go through some details later today, but only a quick note here: 

The statement
> Builds upon https://github.com/KhronosGroup/KTX-Software/pull/876 
implies: It is really _based on the state_ (commit hash) from the swizzle PR. So when the swizzle PR is merged, there will be no conflict here. And even when the swizzle PR is closed (unmerged), and only _this_ is merged, then the result would still include the swizzle fix. 

I opened this PR - which goes _beyond_ the swizzle fix - as a 'Draft', because there may be a few more details to talk about before merging it. 

(And eventually, even _this_ PR might become obsolete, depending on how the https://github.com/KhronosGroup/KTX-Software/issues/880 will be tackled, but I'll try to create a rough "work plan" based on your other comments)

interface/java_binding/src/main/java/org/khronos/ktx/KtxTexture2.java
javagl: This was taken from https://github.com/KhronosGroup/KTX-Software/blob/5414bd0ebfdfdf4ecb8c613a6edb21372303c2b9/lib/writer2.c#L762 .

I have to rely on the documentation to be complete and correct when "porting stuff". 

But I have taken an note to update the documentation here (in the native layer, which will probably affect _all_ documentations in the "bindings"...)

interface/java_binding/src/test/java/org/khronos/ktx/test/KtxTexture2Test.java
javagl: I know that it would be _better_ to compare the data, to have a more thorough test. And this is **ONLY** Zstd (lossless), so I assume that it will not suffer from the ~"slight inaccuracies" that may appear in other comparisons, like the one described in https://github.com/KhronosGroup/KTX-Software/pull/876#discussion_r1537923784 . 

I could spend dozens of hours with ~"improving things". This includes the tests, and even the _existing_ tests (which, in the JNI binding layer, are ... a bit shallow, to be honest). But I took a note of that as well, and will try to tackle that as part of the overall "improvement"...

interface/java_binding/src/main/java/org/khronos/ktx/KtxTexture2.java
MarkCallow: > I have to rely on the documentation to be complete and correct when "porting stuff".

Absolutely. Sorry for the omission.

interface/java_binding/src/test/java/org/khronos/ktx/test/KtxTexture2Test.java
MarkCallow: There won't be any non-determinism or inaccuracies with Zstd.

interface/java_binding/src/main/cpp/libktx-jni.cpp
MarkCallow: Thanks for the explanation. Makes sense.

interface/java_binding/src/test/java/org/khronos/ktx/test/KtxTexture2Test.java
```diff
@@ -227,4 +236,127 @@ public void testCreate() {
...
+		// Transcode the resulting texture to RGBA32
+		int outputFormat = KtxTranscodeFormat.RGBA32;
+		int transcodeFlags = 0;
+		t.transcodeBasis(outputFormat, transcodeFlags);
+		byte[] actualRgba = t.getData();
+
+		// Define the expected RGBA pixels. These are the swizzled input
+		// pixels, with slight deviations due to compression artifacts
+		byte[] expectedRgba = new byte[sizeX * sizeY * 4];
+		TestUtils.fillRows(expectedRgba, sizeX, sizeY, 0, 8, 2, 255, 2, 255);
```
MarkCallow, line 284: This is very fragile, subject to breakage when the encoder changes. I don't know how it is working across platforms given the issue I pointed you at yesterday.  I thought the "fixed" solution you were proposing was to also encode the source without swizzling and compare the rows of the transcoded swizzled data with what are expected to be the matching rows of the transcoded unswizzled data.

interface/java_binding/src/main/java/org/khronos/ktx/KtxTexture2.java
MarkCallow: Please fix the documentation in this PR. The overhead of a PR is too big to justify a separate one for these tiny documentation fixes.

MarkCallow:
@javagl I will be away from Thursday morning for 10 days. If you can make the requested changes and merge the changes from main before 22:00 JST (13:00 UTC) today I can merge before I leave otherwise it will have to wait.

interface/java_binding/src/test/java/org/khronos/ktx/test/KtxTexture2Test.java
javagl: This refers to a change that we talked about in the 'swizzle PR'. (Sorry about the confusion here - I'll try to not open parallel/overlapping PRs in the future). 

I summarized my understanding about how the test should be done at https://github.com/KhronosGroup/KTX-Software/pull/876#issuecomment-2022693341 now.

interface/java_binding/src/main/java/org/khronos/ktx/KtxTexture2.java
MarkCallow: Still need the documentation fix.

interface/java_binding/src/test/java/org/khronos/ktx/test/KtxTexture2Test.java
MarkCallow: Resolving this as it has been fixed in #876.

interface/java_binding/src/main/java/org/khronos/ktx/KtxTexture2.java
javagl: I can add this documentation fix here

But note the first bullet point of the things "Not (yet) supposed to be addressed here", at the bottom of https://github.com/KhronosGroup/KTX-Software/pull/886#issue-2208337582 :   

The corresponding functions do not yet (all) exist in the bindings. For example, the Python binding has `deflateZstd`, but not `deflateZLIB`. So this will have to be aligned at a later point anyhow.

javagl: The documentation has been fixed as of https://github.com/KhronosGroup/KTX-Software/pull/879/commits/d3dd37d9f588470e6184beb651052b6a1432448e and https://github.com/KhronosGroup/KTX-Software/pull/879/commits/9a0b280dbb1d04f2d11f9fb7843bdaae2a3385e2

MarkCallow (approved):

PR #883, opened 2024-03-25
fix(cmake): Fix public includes of `obj_basisu_cbind` by adding (only) BasisU include directories to its public interface.

atteneder:
There are two commits.

The first commit adds BasisU specific include directories to the target `obj_basisu_cbind`'s public interface. Now targets that add `obj_basisu_cbind` as dependency automatically have the correct include directories setup.

The second commit makes use of the first's improvements, by simplifying `transcodetests` include directory setup.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -466,6 +466,13 @@ elseif(APPLE OR LINUX)
     endif()
 endif()
 
+set(KTX_BASISU_INCLUDE_DIRS
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/lib/basisu/transcoder>
+    $<INSTALL_INTERFACE:lib/basisu/transcoder>
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/lib/basisu/zstd>
+    $<INSTALL_INTERFACE:lib/basisu/zstd>
+)
+
 # Main library
 add_library( ktx ${LIB_TYPE}
     ${KTX_MAIN_SRC}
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -459,6 +459,13 @@ elseif(APPLE OR LINUX)
     endif()
 endif()
 
+set(KTX_BASISU_INCLUDE_DIRS
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/lib/basisu/transcoder>
+    $<INSTALL_INTERFACE:lib/basisu/transcoder>
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/lib/basisu/zstd>
+    $<INSTALL_INTERFACE:lib/basisu/zstd>
```
MarkCallow, line 466: Do users of obj_basisu_cbind need these include files? Do they really need to be in the INSTALL_INTERFACE?

atteneder: Good question. I did this work a long time ago and will need to take another look.

In my projects I don't need the INSTALL_INTERFACE, so that could be removed.

On second thought, I wonder if we should remove that target altogether. Chances are, I'm the only real user and using BasisU's targets directly (and implement C bindings there) seems to make more sense.

I'll do some research.

Thanks for the feedback!

atteneder: I've reduced the public header directories to just `$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/lib/basisu/transcoder>` now. That's still required as `basisu_c_bindings.h` exposes `basisu_transcoder.h`.

This way it now works on my project, which imports the CMake configuration via `add_subdirectory(KTX-Software)` and `transcodetests` work as well.

I've prototyped a variant where basisu_transcoder.h/any additional header dir is not required at all, but that has several downsides:
1. Class `basis_file` becomes internal and gets passed to the user as a `void *`.
2. `ktx_basisu_get_texture_type`'s return type would need to be changed from the `basis_texture_type` enum to an int.

Imho not as clean, so I'd prefer the current state of the PR.

Happy to hear your thoughts about it.

MarkCallow: Looks good to me.

MarkCallow:
Thank you @atteneder.

PR #884, opened 2024-03-25
feat(cmake): Option `KTX_FEATURE_ETC_UNPACK` to exclude ETC decoding.

atteneder:
The option defaults to `ON`, so existing setups won't change.

To quote [LICENSE.md](https://github.com/KhronosGroup/KTX-Software/blob/main/LICENSE.md#special-cases):

> The file lib/etcdec.cxx is not open source. It is made available under the terms of an Ericsson license, found in the file itself.

This option allows for easy exclusion of `etcdec.cxx` in cases where strict open source compliance is required.


MarkCallow:
Thank you for this.

Issue #880, opened 2024-03-24
Improvements for the Java/JNI bindings


javagl:
The **tl;dr** of this issue could be:

I'd try to allocate some time to clean up and refactor the JNI part in some ways. 

Anybody opposed?

---

Details:

I recently wanted to use the Java bindings that are offered via https://github.com/KhronosGroup/KTX-Software/tree/main/interface/java_binding . I noticed some issues ( https://github.com/KhronosGroup/KTX-Software/issues/624,  https://github.com/KhronosGroup/KTX-Software/issues/875 , https://github.com/KhronosGroup/KTX-Software/issues/877 ...), and started fixing some of them ( https://github.com/KhronosGroup/KTX-Software/pull/876 , https://github.com/KhronosGroup/KTX-Software/pull/879 ...). But while diving into the implementation and the JNI part, I noticed that there are some more general questions that might warrant a dedicated discussion. Some of this might have to be broken down into dedicated issues, but I'll start with an overview here.

---

The current approach for the JNI bindings aims at offering the functionality of the KTX-Software - mainly, the `ktxTexture2` class - with a 1:1 mapping of the original API. This is a reasonable approach. Any attempt to introduce a "convenience layer", abstractions, or functionality that is more "Java-idiomatic" raises a bunch of questions, and bears the risk of introducing bugs in the translation layer or limitations compared to the original API. However, there _are_ some questions that have to be answered even for the 1:1 translation (and I think that they should be solved differently than they are right now). 

Some questions are fine-grained and specific. For example: I think that the `createFromNamedFile` and `writeToNamedFile` methods should not be offered as part of the `KtxTexture2` class in the current form. File systems and IO are too language- and platform-specific. (There may be convenience methods for stuff like that, but the underlying C/C++-functions should not be routed through via JNI). Some of these points may have to be discussed on a case-by-case basis, though.

--- 

One broader, general point is that of ...

## Error handling

To get this straight: Error handling in JNI is a pain in the back. It bloats the code by a factor of 5, and is "frustrating" insofar that the developer knows in 99% of the cases that the error will not happen 99% of the time.

But right now, there is **zero** error handling for the Java bindings.

A call like
`KtxTexture2 t = KtxTexture2.create(null, KtxCreateStorage.ALLOC);`
will crash the JVM.

A call like
`t.setImageFromMemory(0, 0, 0, null);`
will crash the JVM.

A call like
`t.writeToNamedFile(null);`
will crash the JVM.

[...]

Issue #624, opened 2022-08-23
JAR release to Maven Central


tlf30:
Hello,
I saw some prior discussion on the matter, but would it be possible to get the jni binding jars released to maven central?
Maven central is the key location for libraries to be accessible via build tools. 

Thanks,
Trevor

MarkCallow:
I'm not against it. I don't know how to accomplish it. If someone can provide a PR, if changes are necessary to source or build system, and clear instructions for what I need to do then this can be done.

tlf30:
Does Khronos have a maven central account that is approved? The process is actually very easy it just looks daunting at first.

MarkCallow:
I'll check.

MarkCallow:
Khronos does not currently have an account. 

What is involved in syncing with the Maven repository? My colleague took a quick look, and it seems we can either host with Sonotype OSS repository hosting, or we can self host the repository and sync with Smart Proxy.

https://central.sonatype.org/publish/large-orgs/sync/
[OSSRH](https://central.sonatype.org/publish/publish-guide/)
[Smart Proxy](https://help.sonatype.com/repomanager2/smart-proxy)

What is the right approach here?

tlf30:
You will probably want to setup the OSSRH to let them host it. The link you provided are the instructions. It is a bit clunky making a jira account to request a 'project'.

Depending on how your organization is setup, you will want to request your highest level domain (no subdomain) for your packages. You can add the appropriate subdomain classpaths to each package you release.

MarkCallow:
Turns out Khronos does have a Maven Central account and an established group id: _org.khronos_. We have now set up an account I can use for publishing the KTX jar files to this group id.

Please look at [pom.xml](https://github.com/KhronosGroup/KTX-Software/blob/9bd2f9bbf39834beafad130b695ea36634c4f2a5/interface/java_binding/pom.xml#L170) starting at the linked line. First there is stuff to do with signing and then, commented out, with publishing to Maven Central. 

Regarding signing, I couldn't spot where I'm supposed to provide the GPG signature. I also wonder if we can use our SSL.com-issued software signing certificate instead of a GPG signature.

Regarding deployment, as well as fixing this up for our group id and figuring out how to provide the account name and pw to use, I need to know how to call this part of the POM from a CI deployment section.

Can you make a PR with the necessary changes and  clear indications of where I should provide signing key and account credentials?

[...]

PR #861, opened 2024-02-14
Allow A8B8G8R8 formats.

MarkCallow:
Fixes issue #860.

Updates the switches used by the the GL loader to those generated by `generate_format_switches.rb` in the KTX-Specification repo as a way to get them updated for these and some other recently added Vulkan formats. The glloader now loads the newly allowed formats.

VkFormat declarations for Java and Python are now generated by `mkvkformatfiles` which has been moved to ci_scripts since it now generates files outside the `lib` directory. The alternative was to hand edit these files to add the newly allowed formats.

CI testing of vkformat-related file generation (`mkvk`) has been moved to a dedicated GitHub workflow and removed from the main CI builds. This is because testing now requires a clone of the KTX-Specification repo.

The change to `createdfd.c` is to remove a duplicated line of code and is unrelated to the purpose of this PR.

https://github.com/KhronosGroup/KTX-Software-CTS/pull/19 removes the tests for these as prohibited and adds new tests for the formats.

PR #886, opened 2024-03-26
JNI improvements

javagl:
Addresses https://github.com/KhronosGroup/KTX-Software/issues/880 

I know, the title is too generic. It would be better to clearly lay out a work plan and create a clean sequence of self-contained, _specific_ PRs. But there are just _too many (unrelated) things to do_. "You can't cross a chasm in two small jumps". The first pass has to be tackled somewhat holistically. The result may not be as 'good' as I'd like it to be, but I hope that it will be 'better' along many dimensions. If this is not considered to be a viable approach, then that's OK. Then I'll just create my own repo with JNI bindings for KTX, with blackjack and error handling.

---

The preliminary task list that I tried to create in the [linked issue](https://github.com/KhronosGroup/KTX-Software/issues/880 ) is now replicated here, and I'll try to update it and extend it as we move on:

- [x] Wrap up https://github.com/KhronosGroup/KTX-Software/pull/876 
- [x] Wrap up https://github.com/KhronosGroup/KTX-Software/pull/879 
- [x] Expose `CreateFromMemory`
- [x] Improve the use of constants
  - There should be documentation for these constants, and `stringFor` ... methods to convert them to strings
  - Updates:
    - <strike>Aligned the names of all constants with the native version</strike>
    - Aligned the names of constants with the other bindings (see https://github.com/KhronosGroup/KTX-Software/pull/886#issuecomment-2308404357 )
    - Added documentation and string functions for all constants, except for...
       - `KtxInternalformat`
       - `VkFormat` 
       
       (See https://github.com/KhronosGroup/KTX-Software/issues/887 )

    - NOTES: 
      - The `ktxTexture2.supercompressionScheme` does not appear in the native documentation
      - The constants for `KtxPackAstcEncoderMode` are not documented on the native side
- [x] Fix the `KtxTestLibraryLoader` to also work on Windows
- [x] The JNI class- and field initializations should only be done once. Right now, things like `env->GetFieldID(ktx_texture_class, "instance", "J");` are called in each function. The `jfieldID` should be obtained once initially.
- [x] (continuously) Error checks, error checks, error checks ... (no JVM crashes, period!)
  - Updates (I did a first pass for this, but will keep an eye on places where checks have to be added):
    -  Passing `null` to any method should throw a `NullPointerException`. 
    - Trying to use a texture after `destroy()` was called should throw an `IllegalStateException`. 
    - When a JNI function causes an error (like an out-of-memory error), then the native functions have to return **immediately**, because there is a pending exception. This is not done in all places yet...
    - Java methods that do _not_ return error codes (like the `create...` methods) now generally throw a `KtxException` when the implementation caused an error code that was not `KTX_SUCCESS`
- [x] (continuously) Comments, comments, comments...
  - Updates: 
    - I have added _basic_ JavaDocs, and enabled the creation of the JavaDocs in the POM.
- [x] Add bindings for `ktxErrorString`

---

Not (yet) supposed to be addressed here:

- Try to find a clean solution for https://github.com/KhronosGroup/KTX-Software/issues/624 

- There are a few cases where we might have to think about how to handle the absence of `unsigned` types in Java. For example, when calling `createInfo.setBaseWidth(-123); // INVALID!`, then this value is implicitly converted into a `ktx_uint32_t`.

- <strike>Update the documentation of `deflateZstd` and `deflateZLIB`</strike>
  EDIT: This was now done as part of https://github.com/KhronosGroup/KTX-Software/pull/879, but only for the native side and for the Java bindings (other bindings are still missing the doc update or the functions)
  - As pointed out in [a comment](https://github.com/KhronosGroup/KTX-Software/pull/879#discussion_r1537201514), this documentation should include the information that 'the data pointer is updated'. This will go into the [native documentation](https://github.com/KhronosGroup/KTX-Software/blob/5414bd0ebfdfdf4ecb8c613a6edb21372303c2b9/lib/writer2.c#L762) first, but it will also have to be passed through to the [python bindings](https://github.com/KhronosGroup/KTX-Software/blob/7b08daaaa2f1271d0040d9307156c651f55ad58a/interface/python_binding/pyktx/ktx_texture2.py#L168), the Java binding, and maybe to the JS binding. **But...** the `deflateZLIB` function is still missing in the Python binding (and the `deflate*` functions are just about to be added to the Java binding), so ... this may have to be tracked and addressed separately

- Improve thoroughness of `deflate*` tests by comparing the deflated values
  - As pointed out in [another comment](https://github.com/KhronosGroup/KTX-Software/pull/879#discussion_r1537209791), the unit test for the `deflate*` functions should compare the result of _inflating_ the data again to the original input. The `inflate*` methods are not yet exposed (anywhere), so this may have to be deferred for now

- Add test for input swizzling in ASTC
  - The [PR for fixing the swizzling in JNI](https://github.com/KhronosGroup/KTX-Software/pull/876) includes a test for the swizzling for Basis. It does _not_ include one for ASTC. These are different code paths, and should both be tested. But right now, I'm not sure _what_ could be checked for the ASTC result (see [this PR comment](https://github.com/KhronosGroup/KTX-Software/pull/876#discussion_r1539044357) for context).

- Consider consistent formatting. I know, it's a detail, but just auto-formatting everything with the Eclipse default could already be an improvement. I do have strong preferences for my own projects, but here, the main point is that the formatting should be consistent...
 
- The [ktxHash...](https://github.khronos.org/KTX-Software/libktx/classktxHashList.html) functionalities are not yet mapped through from the native side.
  - Consider following the new JS binding here. It adds `findKeyValue`, `addKVPair` and `deleteKVPair` methods to the ktxTexture2 object rather than exposing `ktxHashList`.


javagl:
I locally started a few updates, including some basic error handling, so that things like `KtxTexture2.create(null, ...)` will no longer crash the JVM, but throw a `NullPointerException` instead. (Many more error checks to add...).

One thing that could/should be done next: The `KtxTexture` class has a `destroy()` method, documented like this:
```java
    /**
     * Destroy the KTX texture and free memory image resources
     *
     * NOTE: If you try to use a {@link KTXTexture} after it's destroyed, that will cause a segmentation
     * fault (the texture pointer is set to NULL).
     */
    public native void destroy();
```
Now, the goal is: **No segmentation faults**. And we'll probably not be able to completely avoid this `destroy()` method. So the question is how to handle that.

I think that trying to use a `KtxTexture` after `destroy()` was called should also throw an exception - maybe an `IllegalStateException`. I'll probably draft this (with the type of the exception being easy to change). But it will affect basically _each and every_ native function, which has to check whether the `thiz` parameter has been destroyed...

MarkCallow:
> ```
> * Set the image data - the passed data should be not modified after passing it! The bytes
> * will be kept in memory until {@link KTXTexture#destroy} is called.
> ```

What does "passed data" refer to here? Is it the data at the pointer location that was passed to setImageFromMemory? Not that it matters since you say it isn't true. Indeed the data is copied into the ktxTexture object.

javagl:
> What does "passed data" refer to here? Is it the data at the pointer location that was passed to setImageFromMemory? Not that it matters since you say it isn't true. Indeed the data is copied into the ktxTexture object.

There are multiple "layers" here, and this revolves exactly about the question where data is _copied_. Imagine the following Java **pseudo** code, similar to what _could_ appear in the KTX JNI bindings:

```java
byte input[] = new byte[10];

// Create a texture from the input
KtxTexture2 texture = KtxTexture2.createFrom(input);

// XXX Will this affect the result?
input[5] = 123;

// Encode the input data that was given earlier
texture.encode();
```
The comment suggested that this `input[5] = 123` would be "visible" for the 'encode' function that is called later. This would mean that the native library would store "a pointer" to the Java `byte[]` array. Something like this _is_ actually possible, but with many, many, caveats - and here, it certainly is not the case.

javagl:
A detail @ShukantPal : Currently, the bindings are set up for Java 11. I don't see a strong reason for that. Unless there is a reason, I'd suggest lowering the version requirement to "the lowest version that works" (and I think that 8 should do it).

javagl:
I have added a bit of error handling (updating the list in the first post accordingly).

One point about error handling is that there are still some cases where the JNI layer reports errors to `std::cout`, and then returns `NULL` or so. I'm not yet sure what to do in these cases. (But I'm sure that the errors should not (only) be reported via `std::cout`...). I think that these cases will _usually_ be converted into a thrown exception, and  I'll decide on a case-by-case basis what kind of exception that could be. If someone has strong preferences: Chime in here.

javagl:
Two updates:

### Caching field- and method IDs

All the `jfieldID` and `jmethodID` values are now obtained _once_, when the library is loaded, instead of looking them up again and again in each and every method call. 


### Error handling 

Additional checks for out-of-memory errors and such have been added.

The most significant change refers to the `create...` Methods - or more generally, to the methods where the Java version did **not** return an error code as an `int`:  These methods generally had the pattern of (pseudocode): 
```
KtxTexture2 create(...) {
    ktxTexture2 *instance;
    ktx_error_code_e = create(&instance);
    if (result != KTX_SUCCESS) {
        std::cout << "Something wrong " << result << std::endl;
        return NULL;
    }
    return createJavaObjectFrom(instance);
}
```
Now, there is no sensible method for returning the information about what went wrong here - with the exception of an exception. So these methods will now usually throw the newly added `KtxException` when the internal error code was not `KTX_SUCCESS`. 

One could think about other options here. For example, one could pass an uninitialized `KtxTexture2` object to the `create` methods, as in
```
KtxTexture2 texture = new KtxTexture2(); // NOT INITIALIZED!:
int result = KtxTexture2.create(..., texture); // Try to initialize it
if (result == KtxErrorCode.SUCCESS) { 
    // That worked
} else {
    // Handle this...
}
```
Yeah, that would be a "breaking change" - but I guess nobody really cares about that right now.
I don't have a really strong preference here.

javagl:
The last passes contained a few steps of that dull, repetitive, boring, repetitive, tedius, and repetitive handling of JavaDoc, based on the documentation from the native side.

And apparently, the JavaDocs have never been used or created before. Most of the (few) existing ones had been invalid. Now they are fixed.

I have added the creation of the JavaDoc to the POM. The current state is attached here:

[libktx-4.0.0-SNAPSHOT-javadoc.zip](https://github.com/KhronosGroup/KTX-Software/files/14838479/libktx-4.0.0-SNAPSHOT-javadoc.zip)

There are a few degrees of freedom for reducing (or handling) the redundancy of JavaDoc comments for cases like the `...Param` classes, which correspond to the native `...Param` structs, *but* introduce setters/getters for each field. The current approach is what I consider to be a reasonable middle-ground:

- The field itself receives a _very_ short comment (roughly: The first sentence from the native documentation)
- The getter receives a similarly short comment, but linkts to the setter (with `See [link]`)
- The return value of the getter and the parameter of the setter receive a _very_ short description (sometimes even resorting to `"The setting"` or `"The value"`, just to keep things manageable...)
- The _actual_ documentation is put into the setter

The rationale here is that 
- users don't see the field itself, which is `private`
- users who see the getter have the direct link to the setter
- the setter is where the actual information is most relevant - mainly, the valid _ranges_, constraints, and the effect of the setting.

Here's an example for the `uastcRDOMaxSmoothBlockErrorScale`:

Khronos KTX JavaDoc

(Yeah, .. who's ever gonna change that anyhow? But I see complete JavaDoc as some sort of minimal baseline, to be _at least_ on par with the native docs, even if the docs might have to be extended to be _really_ useful for end-users...)

javagl:
About the current build failures:

What's surprising is that I also don't see these errors _locally_, and they don't happen in the Windows build. Iff I understood that correctly, very roughly: Windows builds are done by GitHub actions. Linux/MacOs are done by Travis - and the error only appears in https://app.travis-ci.com/github/KhronosGroup/KTX-Software/jobs/620089455#L866 but not in https://github.com/KhronosGroup/KTX-Software/actions/runs/8543283946/job/23406762987#step:14:48 

I'll probably have to add some quirky `printf`-logging in the JNI layer to even be able to _guess_ why `CreateFromMemory` might fail on Linux/MacOS...

MarkCallow:
> About the current build failures:
> 
> What's surprising is that I also don't see these errors _locally_, and they don't happen in the Windows build. Iff I understood that correctly, very roughly: Windows builds are done by GitHub actions. Linux/MacOs are done by Travis - and the error only appears in https://app.travis-ci.com/github/KhronosGroup/KTX-Software/jobs/620089455#L866 but not in https://github.com/KhronosGroup/KTX-Software/actions/runs/8543283946/job/23406762987#step:14:48
> 
> I'll probably have to add some quirky `printf`-logging in the JNI layer to even be able to _guess_ why `CreateFromMemory` might fail on Linux/MacOS...

Don't obsess about Travis vs. Actions. Both macOS and Linux builds are failing. These builds are run on different machines, even though both under Travis. Look instead at the environments on the runners, particularly the Java/JVM versions in use.

MarkCallow:
> About the current build failures:

There are 2 circumstances in which `KTX_INVALID_OPERATION` can be returned during `ktxTexture2_CreateFromMemory`:

1. If the offset of the supercompression global data or of the first mip level is greater than the byte length specified to `ktxTexture2_CreateFromMemory`. This comes from `ktxMemStream_setpos`.
2. If you are trying to load the data and it has already been loaded. For example calling `ktxTexture2_LoadImageData` after calling `ktxTexture2_CreateFromMemory` with KTX_TEXTURE_CREATE_LOAD_IMAGE_DATA_BIT set. This comes from `ktxTexture2_LoadImageData`.

Not clear why either of these would be triggered on Linux and macOS but not Windows.

javagl:
Thanks for the details @MarkCallow . 

I added some `printf`-debug-log output (in the hope that it would show up in the build logs), but during that, noticed an embarrassing error in the JNI function for acquiring the input data for `CreateFromMemory`: A sign error for the computation of the `size` of the data caused a value like `-1000` to be interpreted as a `size_t`, so the function was called with something like `size=1844674407370955061` (~ `uint64` max value).

This is fixed now. Let's see whether the build passes.

Two points remain open:

- Why didn't this cause an error on Windows? (It looks like Windows is ignoring the `size` parameter in some way?)
- It is related to the question that I already brought up in the (updated) bullet point list of the first post here a while ago:
  - **TODO**: ... We may have to think about how to handle the absence of unsigned in Java in some cases...

The latter is a more general question. Unsigned types are error-prone, particularly in interfaces - see e.g. https://www.aristeia.com/Papers/C++ReportColumns/sep95.pdf , which points out **exactly** what happened here. And when crossing the language barrier between Java and C++, one has to be _particularly_ careful about that...

javagl:
@MarkCallow We'll have to think about a sensible strategy for reviewing and merging this.

Looking at the _changes_ will hardly make sense. Nearly every file of the JNI bindings was modified. I tried to update the bullet point list in the first post with the progress. Beyond that, I can probably _try_ to give a short summary here, with an overview of the most important changes/commits:

- Added JavaDocs. This has been spread over many commits, only one example: https://github.com/KhronosGroup/KTX-Software/pull/886/commits/2810d2b8012ae205fe0b3dee32249bff1f648e6b 
- Update the `KtxTestLibraryLoader` to work on Windows, via https://github.com/KhronosGroup/KTX-Software/pull/886/commits/bf0c644f0e608d62fe8a18cfe4df04d2ac660e8d 
- [Breaking]: I updated _all_ constants to match the name that they have in the native layer. For example, https://github.com/KhronosGroup/KTX-Software/pull/886/commits/c3c860072d509b1ea4cecac6a68b3d238d1a295d#diff-7e22517bac1468d9f5cff73b4d8e6ecdb2cc11608424890e8f2ac6d9bb06fa19 
- Added `stringFor`-functions that take the respective enum constants, and return their string representation (e.g. https://github.com/KhronosGroup/KTX-Software/pull/886/commits/332510f2b98cc031c7f40b926dc1ba6b56d8c914 ). Note that this only returns the string representation, and not the elaborate description that some of the functions in the native layer return
- [Build process]: Added the step to generate the JavaDocs to the Maven pom.xml at https://github.com/KhronosGroup/KTX-Software/pull/886/commits/b74784930da12f73876e1597ef73d8808480c619 
- The field- and method IDs that are used for accessing the Java structures from the JNI layer are now initialized once, when the library is loaded (via https://github.com/KhronosGroup/KTX-Software/pull/886/commits/2c86122cac0807231aa5504031225edbf2b7b46f )
- One of the most important points: Error handling:
  - Passing in `null` to any function will **not** cause a JVM crash any more, but throw a `NullPointerException` instead (e.g. https://github.com/KhronosGroup/KTX-Software/pull/886/commits/8f581095765168a194bac0ebc403d942b6a1a486 )
  - When an `OutOfMemoryError` is caused in the native layer, then the functions return immediately (and throw this pending error), e.g. https://github.com/KhronosGroup/KTX-Software/pull/886/commits/a5adb8abad12e89f20ebe0436d200663e51da125 
  - When trying to use a texture on which `destroy` was already called, then this will **not** cause a JVM crash, but throw an `IllegalStateException` instead (mainly done in https://github.com/KhronosGroup/KTX-Software/pull/886/commits/3673c0aa8c043aabbff2bfe759d22d9b9929bdce )

Some of this is just "baseline stuff" to follow the recommendations from https://developer.ibm.com/articles/j-jni/ . But some points still have subjective elements to them.

---

I have actively been _using_ the resulting library in a small dummy application where you can drag-and-drop a PNG file, see a preview of the PNG and its KTX version, drag around some sliders for the compression/quality (and see a live preview). So I'm confident that this "mostly works". But there are still a few things to wrap up. 
- Some of the JNI implementations still have to be reviewed to ensure that they don't perform operations with pending exceptions, and maybe run with `-verbose:jni` to spot further possible pitfalls that already existed in the code
- There should be a version of `setImageFromMemory` that does not accept a `byte[]` array, but a `ByteBuffer` as the last parameter (mentioned in https://github.com/KhronosGroup/KTX-Software/pull/886#issuecomment-2020869640 ) 
- The open points of the bullet point list from the first post

But maybe this can already go through a first review pass, based on the current state.

MarkCallow:
How do we make further progress on this?

I find having the full KTX_ prefixes rather clunky. Can't we follow the WebGL model which removed all the GL_ prefixes?

javagl:
That's the question: What could be a sensible way to review this, given that it is ... _nearly_ a "rewrite" of the JNI bindings? 

Is the bullet point list above sufficient? Should I invite further potential reviewers? I'm pretty flexible here. If you wish, we could also arrange a short call to go through the changes, what's left to do, and what could be the next steps.

> I find having the full KTX_ prefixes rather clunky. Can't we follow the WebGL model which removed all the GL_ prefixes?

From quickly skimming over some Python binding files, it looks like the Python bindings are also just omitting the `KTX_`, so the question is then: Should the naming be consistent with the C-layer or should it be consistent with the bindings? I think that changing
`KtxTextureCreateStorage.KTX_TEXTURE_CREATE_NO_STORAGE` to
`KtxTextureCreateStorage.TEXTURE_CREATE_NO_STORAGE` 
does not make a crucial difference here, but when you say that the `KTX_` prefixes should be omitted, I'll remove them.

MarkCallow:
@javagl, PR #874 extends the Javascript binding to provide full libktx functionality. As part of it I went over the naming. I made the names consistent with the libktx names and I ended up removing all the `ktx` and `KTX_` prefixes. This is because the module that contains the binding is typically called `ktx`. Please take a look. Perhaps it can inform the naming here too. If we change any names that exist in the current binding we should provide the old names as aliases.

I suggest asking @ShukantPal to review this as he is the author of the original binding.

The bullet point list at the top is a good guide. Obviously there are items to be completed. I think a short call would be good but I want to first familiarize myself with the code. I'll let you know when I've done that and we can then plan the call.

javagl:
Thanks @MarkCallow , I'll try to allocate some time looking over the changes from the linked PR. _This_ PR should be a step in the direction of aligning the JNI bindings with KTX itself. We can decide later whether some missing parts (like bindings for the `ktxHash...` family of functions) will become part of _this_ PR or a dedicated, new one. 

> Perhaps it can inform the naming here too. If we change any names that exist in the current binding we should provide the old names as aliases.

I'm in favor of aligning names insofar that they should be _recognizable_ as their native counterpart. There are some conventions in the respective target languages, though. For example, I wouldn't use something else than a `UPPER_SNAKE_CASE` for constant names or something else than `CamelCase` for class names in Java, but the details can still be sorted out. (E.g. the removal of the `KTX_` prefix mentioned above)

MarkCallow:
@javagl, I've been reviewing the discussion here and some of the code including the commits you specifically called out in [comment of April 18th](https://github.com/KhronosGroup/KTX-Software/pull/886#issuecomment-2062022093). The code I've seen looks very good.

I added a note to the [description](https://github.com/KhronosGroup/KTX-Software/pull/886#issue-2208337582) suggesting following the JS binding handling of metadata queries.

> Added stringFor-functions that take the respective enum constants, and return their string representation (e.g. https://github.com/KhronosGroup/KTX-Software/commit/332510f2b98cc031c7f40b926dc1ba6b56d8c914 ). Note that this only returns the string representation, and not the elaborate description that some of the functions in the native layer return

Can't you call the native function(s) from the stringFor function to get the elaborate description. I'd particularly like to have the full descriptions of the error codes. Are there any beyond `ktx_error_code_e` where the native side has a convert-to-string function?

I am happy you have replaced the crashes with exceptions.

> I have actively been using the resulting library in a small dummy application where you can drag-and-drop a PNG file, see a preview of the PNG and its KTX version, drag around some sliders for the compression/quality (and see a live preview). So I'm confident that this "mostly works". But there are still a few things to wrap up.

Any plan to release this? Does the preview show you both pre- and post-compression images so you can easily compare them?

>
>    Some of the JNI implementations still have to be reviewed to ensure that they don't perform operations with pending exceptions, and maybe run with -verbose:jni to spot further possible pitfalls that already existed in the code

Which implementations?

>    There should be a version of setImageFromMemory that does not accept a byte[] array, but a ByteBuffer as the last parameter (mentioned in https://github.com/KhronosGroup/KTX-Software/pull/886#issuecomment-2020869640)

I agree a version that accepts ByteBuffer is needed.

>    The open points of the bullet point list from the first post

Yes.

>But maybe this can already go through a first review pass, based on the current state.

There is certainly enough here to review.  I want to spent a little more time familiarizing myself with the code before we arrange that call.

Note that all GitHub Actions Windows builds are currently broken due a bad GHA runner image. Many tests, including the Java binding tests fail with segmentation violations.

javagl:
Thanks for the detailed review.

---

> I added a note to the https://github.com/KhronosGroup/KTX-Software/pull/886#issue-2208337582 suggesting following the JS binding handling of metadata queries.

I haven't looked at the detals of the hash-related functions. If it's "only" about exposing 3 functions, then it _might_ be added here, to be on par with other bindings, but if it's too much, it might make more sense to carve it out into a follow-up PR.

---

> Can't you call the native function(s) from the stringFor function to get the elaborate description. I'd particularly like to have the full descriptions of the error codes. Are there any beyond `ktx_error_code_e` where the native side has a convert-to-string function?

There are two levels here: 

1. A function of obtaining a string version of the constant itself
2. A function for the actual _description_ of the _meaning_ of the constant

Until now, I only addressed 1.: Each "enum" has a `stringFor` function. This is particularly useful for the `KtxErrorCode`, where you could have
`System.out.println("Error " + errorCode); // Prints '16' - what is that?`
vs
`System.out.println("Error " + KtxErrorCode.stringFor(errorCode)); // Prints 'UNSUPPORTED_TEXTURE_TYPE'`
But similarly applies to all other "enum" types.

Now, I saw that there is the [`ktxErrorString`](https://github.com/KhronosGroup/KTX-Software/blob/29aeddefd6d02630a3e8bcda7c6202aac4a58c77/lib/strings.c#L59) function in the native layer. (And I think that this the only function of this kind in the native layer, but will have to check again to be sure). And one _could_ make a case for exposing that via JNI, to be able to do something like
`System.out.println("Error " + KtxErrorCode.errorString(errorCode)); // Prints "Texture type not supported."`
I think that this does not add soooo much value beyond `UNSUPPORTED_TEXTURE_TYPE`, but can add this, if desired.

---

> > I have actively been using the resulting library in a small dummy application 
> 
> Any plan to release this? Does the preview show you both pre- and post-compression images so you can easily compare them?

Until now, this was mainly for internal experiments and testing. I think that it _might_ eventually have some aspects that are "useful", and that it could be worth being published. We'll probably have to sort out some of the "Maven Release"-related questions before this can be done, but maybe there can be an "early sneak preview" release. 

The goal was to have some sort of "interactive preview". (As 'interactive' as it can be, when encoding an image takes 5 seconds...). I did create a short screengrab of that a while ago (yeah... as a GIF... but), maybe it shows the overall idea:

Khronos KTX Sneak Preview 00002

---

> > Some of the JNI implementations still have to be reviewed to ensure that they don't perform operations with pending exceptions, and maybe run with -verbose:jni to spot further possible pitfalls that already existed in the code
> Which implementations?

I did try to keep an eye on that for every function that I "touched", but did not systematically go through all function from top to bottom. It boils down to the following pattern: 

When any JNI function that is called with `env->...` causes an exception, then it is **not** allowed to call **any** other JNI function. Instead, the function must return immediately (and leave the handling of the pending exception to the Java layer). 

(The only functions that may be called with a pending exception are the ones like `env->ExceptionCheck()` or certain `env->Release...` functions. Details in the [specification](https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/design.html#exception_handling)).

To check this manually, I'd have to read _every_ function and see whether it exposes this pattern (calling a function when an exception _might_ be pending). It can be automated to _some_ extent, with the `‑Xcheck:jni` `verbose`, but this only covers the case where such an exception _actually_ appears....

---

I still have to look closer at the other PR, but will try to at least pick up the work here (syncing with `master`, addressing some of the smaller points mentioned above...) in the next days.

MarkCallow:
> There is certainly enough here to review. I want to spent a little more time familiarizing myself with the code before we arrange that call.
> 

I am now back to familiarizing myself with the more of this code. Sorry for the looooong delay. Can we arrange a call towards the end of next week, FaceTime or Skype? I want to get this wrapped up.

Regarding constant naming. both the JS and Python bindings remove KTX_ and all other parts of a name that are reflected in the class or enum name. (The sole slight exception I've spotted is e.g. `KTX_PACK_ASTC_BLOCK_DIMENSION_4x4`, where dimension is in the enum name (JS) or class name (Python) but due to the numerical tail an alphabetic character is needed. Python is using `D4x4`. JS is using `d4x4`. I will change the latter since it is not yet released. I wonder though if it should be `D_4x4` or `DIMENSION_4x4`.) Java should follow the same model for constants.

Another aspect of naming is the enum or class name. In the JS binding I have used the enum or struct name from the C API minus the `ktx` or `KTX_` prefix. enums are therefore `lower_snake_case` and structs, classes in JS, are `CamelCase`. Python has used `KtxCamelCase` for everything. We need to discuss and regularize how to treat these names.

Please merge the latest main into this branch.

You will almost certainly get a failure in the first Travis-CI job of a build which is a Mac build. This is due to now using vcpkg for dependencies for the load test apps.  The dependencies will not be cached in the initial build. The resulting extended log length from building them seems to be hitting a Travis-CI limit.

javagl:
I have updated with the `main` state, and added a few exception checks that had been pointed out by running it with `Xcheck:jni`. 

---

For the constants: I'm OK with either pattern. 
- the full name `KTX_<whatever>_<value>` or
- omitting KTX, as in `<whatever>_<value>`, or
- the short one, just `<value>`

As long as it is consistent, I don't have a strong opinion. It can either be aligned with the C name (i.e. the long form), or the short name that is used in other bindings. 

Java does not differentiate between "struct" and "class". Java does have `enum`, but this is currently not used. From the perspective of the user, the difference between a class (with `public static final int CONSTANTS = ...`) and an `enum` is negligible. For the names of classes, there is a strong convention to use `CamelCase`, and we should not deviate from that.

---

> Can we arrange a call towards the end of next week, FaceTime or Skype? I want to get this wrapped up.

If you don't mind, I'll send you a mail to arrange the details.

MarkCallow:
> If you don't mind, I'll send you a mail to arrange the details.

Certainly. I did not intend to arrange that through this public forum.

javagl:
The `ktxErrorString` function that was mentioned in an earlier comment ( https://github.com/KhronosGroup/KTX-Software/pull/886#issuecomment-2160458217 ) should also be part of the Java API. It should be simple to add, and can become part of this PR as well.

javagl:
The last sequence of commits are...

- _"Update ... names"_ commits: These are changing the names of constants, to omit any redundant prefixes. For example, `KtxPackAstcBlockDimension.KTX_PACK_ASTC_BLOCK_DIMENSION_D4x4` is now just called `KtxPackAstcBlockDimension.D4x4`, which is aligned with the naming in the JavaScript- and Python bindings
- Added a binding for `ktxErrorString`. It's a judgement call where and how to add this. In the native layer, it's a "free-floating" function that does not belong anywhere. On the Java side, I thought that it would make sense to offer it as `KtxErrorCode.createString(int)`.
- A minor fix for `ktxErrorString` on the native side

The last one _appears_ to be related to the (open) question about how to handle the absence of `unsigned` types in Java. But strictly speaking, it isn't: Even in the native layer, something like
`const char* s = ktxErrorString(-1234);`
would have caused ... Undefined Behavior, I guess. 

Visual Studio even points this out:

Khronos Ktx Error enum

So I changed that check to 
```
    if (error < 0 || error > KTX_ERROR_MAX_ENUM)
        return "Unrecognized error code";
    ...
```

javagl:
I _think_ that the build failure should already have been sorted out - right now, it marks some actions as 'failed' because they have been _cancelled_. In doubt, I could push ... "some change" ... to trigger a re-run (because I apparently cannot request a re-run manually). 

Apart from that, this PR should be 'Ready for review' (which might be short, because intermediate states have already been reviewed and discussed). 

Miscellaneous:
- At some point, the change log may have to include a hint that the Java bindings have been ~"incompatibly refactored". I think this is not a big deal, because there probably haven't been so many people been using the previous state (and those who did hopefully appreciate the changes, even though some of them may be "breaking" in the strictest sense)
- The open points from the first comment of this PR (Maven release, formatting, ....) will be promoted into dedicated issues where appropriate

MarkCallow:
In GitHub Actions when a job in a workflow fails the remaining jobs are cancelled. The VS2019 job in the Windows workflow is failing. I just tried re-running this job but it still fails in the same place: problems creating the Javadoc. I will try rerunning the cancelled jobs. Some of the Linux build jobs are failing for a similar reason. I did not try to re-run them. I'm don't know yet why some macOS jobs are failing.

When you add more commits to a PR, existing queued builds are cancelled, running builds may be cancelled. A new build is started for the new commit when an existing job is cancelled or finishes. All this means in this case that the errors remain in the code of the most recent commit because a build was done using it.

MarkCallow:
The failures in the Emscripten jobs look to be due to an updated clang in the latest Emscripten Docker image. Research is needed.

javagl:
> I just tried re-running this job but it still fails in the same place: problems creating the Javadoc. 

Yes, I didn't dive deeply enough into the travis output for that one. This was caused by a `{@link #...}` in the JavaDoc that referred to a renamed constant. (I probably should have run that step locally to rule that out). Should be fixed now, but I'll keep an eye on possible further errors.

MarkCallow:
**Status update.**

I worked around the new spam warning from clang 20 fixing the Emscripten build but GitHub just deployed a new version of its Windows runner which breaks builds with python due to https://github.com/python-cffi/cffi/issues/117. I don't think I can do anything until a new version of cffi with PR https://github.com/python-cffi/cffi/pull/118 merged is released. There is no way to tell GitHub Actions to use a particular version of the runner otherwise I'd make it use the previous working version.

This is incredibly frustrating and is adding to the issues that have stalled successful builds for going on 4 weeks now. I have at least 5 fixes backed up in local branches waiting until CI is unblocked. The time taken to deal with all of this is why I have not completed review of this PR.

javagl:
Thanks for the update. 

I've recently heard about a bunch of trouble in CIs, due to ~"updates of the GitHub Windows/MSVC version". It should _really_ be possible to pinpoint that to a specific version in CI. 
(Even though I'm well aware of the _other_ issues that _this_ might cause - if this was available since the dawn of time, GitHub would still have to maintain a Windows 3.1 build infrastructure 😆 ) 

That reminds me of one thing regarding the JNI that is _also_ related to _versioning_ (mentioned as a side note in an [earlier comment](https://github.com/KhronosGroup/KTX-Software/pull/886#issuecomment-2022891797) : Right now, the JNI part assumes Java 11. Unless there is a _strong_ reason to do this, I'd change that to "the lowest version that works" (likely Java 8). 

However, I assume that this PR will eventually be merged without many modifications (because the current hassle is pretty unrelated to JNI). So I'll probably start addressing things like the Maven issue _locally_, based on the state of this PR, and ... when the dust has settled here, open a PR for that.

MarkCallow:
I reviewed all the recent commits. They look good except that I was expecting you to use the new ktxErrorString function in the `stringFor` method of `KtxErrorCode`. Why not use it?

I don't understand why you saw the warning or error about "Unchecked lower bound" and we didn't see it in CI and I don't see it in Intellicode in the IDE. We have warning level W4 set and warnings as errors. I have Microsoft Visual Studio Community 2022 Version 17.10.1 VisualStudio.17.Release/17.10.1+34928.147 and Visual C++ 2022   00482-90000-00000-AA364. Anyway thank you for the fix.

I think I have found a workaround for the python issue on GitHub. Build in progress. Still having issues with Travis.

I am appalled to see this PR has been open for 6 months. Sorry for the loooong delay.

javagl:
> I reviewed all the recent commits. They look good except that I was expecting you to use the new ktxErrorString function in the stringFor method of KtxErrorCode. Why not use it?

At this point, it's only a matter of consistency. The `stringFor` function exists in _all_ "enum-classes". And it _always_ returns the _string representation of the enum value_. I didn't want to deviate from that pattern, insofar that this would imply that there is no (reasonable) way to translate an error code like `42` into the string that actually reflects the `"NAME_OF_THE_CONSTANT"`. So I thought that

- `stringFor` consistently returning the constant name
- `KtxErrorCode.createString` returning the human-readable string 

made sense. But if you prefer, I can change it to replace the `stringFor` function here.

MarkCallow:
Please merge the latest changes from main, which includes the fix for clang 20 in the Emscripten build. After a successful build this PR will be ready for merge.

Please provide a summary of the changes, particularly any breaking changes, that I can use as the commit message and can include in the Release Notes.

javagl:
Technically, there is a whole bunch of "breaking changes", because certain parts have nearly been a "rewrite". But as I mentioned earlier: I think that hardly anybody really _used_ the previous state, so it should not cause too much trouble. The build is currently running. I'll try to write a summary of the changes soon.

MarkCallow (approved):

javagl:
I had a short look at the current `RELEASE_NOTES.md`, but am not sure in how far any existing "pattern" is applicable here. This is not a single bugfix or feature. But something like the following might be added in the release notes, as a summary:

---

### Java Wrapper

There was a larger refactoring of the Java Wrapper, focussing on error handling, documentation, completeness, and usability improvements ( [#886](https://github.com/KhronosGroup/KTX-Software/pull/886) ). This includes several breaking changes. Not all breaking changes can be listed here explicitly. But in all cases, it should be easy for clients to update their code to the new state. The most important changes for the Java Wrapper are:

- Made `deflateZstd`, `deflateZLIB`, and `createFromMemory` available in the `KtxTexture2` class
- Fixed a bug where the `KtxBasisParams#setInputSwizzle` function caused data corruption
- Increased the consistency of the naming and handling of constants:
  - All constant names are `UPPER_SNAKE_CASE`, omitting common prefixes
  - Classes that define constants offer a `stringFor` function that returns the string representation of a constant
  - The `KtxCreateStorage` class was renamed to `KtxTextureCreateStorage`
  - The `KtxPackUASTCFlagBits` class was renamed to `KtxPackUastcFlagBits`
- Improved error handling:
  - When functions receive a parameter that is `null`, then this will no longer crash the JVM, but throw a `NullPointerException` instead
  - When `setInputSwizzle` receives invalid arguments, then an `IllegalArgumentException` will be thrown
  - When one of the `create...` family of functions causes an error, meaning that the respective `KtxTexture2` object cannot be created, then a `KtxException` will be thrown
- Added JavaDocs, and enabled the generation of JavaDocs as part of the Maven build process
- Internal improvements (like JNI field caching, and avoiding to call JNI functions with pending exceptions)

---

(Feel free to edit or adjust that (or request changes) as you see fit)

---

An aside: Is there a reason why the `RELEASE_NOTES.md` contains lines like

> `Allow A8B8G8R8 formats. (#861) (5ac5cf126) (@MarkCallow)`

that are not linked, as in

> Allow A8B8G8R8 formats. ([#861](https://github.com/KhronosGroup/KTX-Software/pull/861)) ([5ac5cf126](https://github.com/KhronosGroup/KTX-Software/commit/5ac5cf126)) (@MarkCallow)

?

PR #889, opened 2024-03-29
WIP prototype clang-format support

aqnuep:
This is largely WIP at this point.


aqnuep:
@MarkCallow, this is still WIP as I still have to update the auto-generated files and their scripts to include `/* clang-format off */` in order to avoid auto-formatting them, and add clang-tidy support. However, this branch already includes the following:

  * Moving external dependencies to the `external` directory
  * Adding a set of clang-format files specific for the directories in question (e.g. `lib` has a special style file to allow return types on separate lines per the traditional convention)
  * Running clang-format across all files

Please check the code and let me know whether the new structure and code format seems reasonable to you.

aqnuep:
@MarkCallow, I did not yet get feedback from you about the proposed code formatting conventions applied in this branch.

Should I maybe create a separate PR first that just moves the external components into the `external` directory (and update build scripts accordingly) to do this step-by-step?

MarkCallow:
> @MarkCallow, I did not yet get feedback from you about the proposed code formatting conventions applied in this branch.
> 
Sorry. I am a little under the weather so working when I feel up to it. I reckon reviewing this is going to take some extended concentration so I've been putting it off. Hopefully I'll feel up to taking a look tomorrow.

> Should I maybe create a separate PR first that just moves the external components into the `external` directory (and update build scripts accordingly) to do this step-by-step?

That is not necessary.

I remain somewhat concerned that formatting is done as a workflow rather than via a clean filter run when files are commited. I have to think more about potential ramifications of the remove branch becoming different from what one has just pushed.

Are there any editors that can apply the same formatting during editing?

MarkCallow:
What is the difference between "format" and "tidy"?

aqnuep:
> I remain somewhat concerned that formatting is done as a workflow rather than via a clean filter run when files are commited. I have to think more about potential ramifications of the remove branch becoming different from what one has just pushed.

This is the practice used in other Khronos projects: modify, format (automatically or through explicit call), then commit. CI only checks that the modifications match the formatting rules.

> Are there any editors that can apply the same formatting during editing?

Many editors have clang-format built-in, AFAICT, but I always do this manually so I don't have much experience with them.

> What is the difference between "format" and "tidy"?

clang-format is only dealing with formatting, while clang-tidy is a "linter" and provides higher level analysis, see: https://clang.llvm.org/extra/clang-tidy/

interface/js_binding/ktx_wrapper.cpp
```diff
@@ -457,36 +419,22 @@ EMSCRIPTEN_BINDINGS(ktx)
...
-    enum_<ktxOrientationY>("OrientationY")
-        .value("UP", KTX_ORIENT_Y_UP)
-        .value("DOWN", KTX_ORIENT_Y_DOWN)
-    ;
-    enum_<ktxOrientationZ>("OrientationZ")
-        .value("IN", KTX_ORIENT_Z_IN)
-        .value("OUT", KTX_ORIENT_Z_OUT)
-    ;
+        .value("TRANSCODE_ALPHA_DATA_TO_OPAQUE_FORMATS", KTX_TF_TRANSCODE_ALPHA_DATA_TO_OPAQUE_FORMATS);
+    enum_<ktxOrientationX>("OrientationX").value("LEFT", KTX_ORIENT_X_LEFT).value("RIGHT", KTX_ORIENT_X_RIGHT);
```
MarkCallow, line 425: Why has it run these `.value`s together? It hasn't in other cases. Can you prevent it here? This is much harder to read and understand than the original formatting with each .value on a separate line.

lib/etcdec.cxx
```diff
@@ -1,9 +1,9 @@
 /*
```
MarkCallow, line 1: We should move this file to `external/etcdec` or similar so it is not reformatted. Strict adherence to the terms of the license required not modifying the file.

lib/etcunpack.cxx
```diff
@@ -22,256 +22,247 @@
...
+    unsigned int width, height;
+    unsigned int block_part1, block_part2;
+    unsigned int x, y;
+    /*const*/ GLubyte* src = (GLubyte*)srcETC;
+    // AF_11BIT is used to compress R11 & RG11 though its not alpha data.
+    enum { AF_NONE, AF_1BIT, AF_8BIT, AF_11BIT } alphaFormat = AF_NONE;
+    int dstChannels, dstChannelBytes;
+
+    switch (srcFormat) {
+    case GL_COMPRESSED_SIGNED_R11_EAC:
```
MarkCallow, line 63: Can you indent the cases by 2 spaces? Since I'm always fighting editors to do this we probably shouldn't continue but I'm curious.

lib/.clang-format
```diff
@@ -0,0 +1,13 @@
+# Copyright 2024 The Khronos Group Inc.
+# SPDX-License-Identifier: Apache-2.0
+---
+# Use defaults from the Google style with the following exceptions:
+Language: Cpp
```
MarkCallow, line 5: Does this apply to the c files as well?

lib/gl_funcs.c
```diff
@@ -69,21 +69,21 @@ HMODULE ktxOpenGLModuleHandle;
 #define GetOpenGLModuleHandle(flags) dlopen(NULL, flags)
 #define LoadProcAddr dlsym
 #define LIBRARY_NAME NULL
-void* ktxOpenGLModuleHandle;
+void *ktxOpenGLModuleHandle;
```
MarkCallow, line 72: I saw the opposite change being made in basisu_c_binding.{cpp,h}, "foo *bar" became "foo* bar". Why the difference here? Is it a c vs C thing? I prefer "foo* bar" because to my mind the type of "bar" is "pointer to foo".

lib/vkformat_check.c
```diff
@@ -15,72 +15,71 @@
 #include "vkformat_enum.h"
 
 bool
-isProhibitedFormat(VkFormat format)
-{
+isProhibitedFormat(VkFormat format) {
```
MarkCallow, line 18: Has the generator been modified to match the clang format or has clang modified this file? I don't see anything to tell clang to ignore it.

MarkCallow (commented):
Here are some comments to start. Some of the comments refer to global things which I've listed here.

- [ ] 1. Narrower line width
- [ ] 2. { on same line as function name. There is no specific comment about this. I want to think about it. I've always put the { on a separate line except on inline functions defined within a class definition. I suppose it is better to be consistent.
- [ ] 3. nesting (indent) of #if and #define
- [ ] 4. case not indented
- [ ] 5. namespace content not indented.


After my experience with this first pass, I think it would be better to make a separate PR for moving all the files. Note there is a comment about also moving etcdec.cxx.

examples/vkload.cpp
```diff
@@ -43,20 +42,14 @@ Texture::Texture(const std::string ktxfile)
     // with the expectation that app's will typically load many textures.
     ktxVulkanDeviceInfo_Construct(&kvdi, gpu, device, queue, commandPool, nullptr);
 
-    ktxresult = ktxTexture_CreateFromNamedFile(
-                                        (getAssetPath() + ktxfile).c_str(),
-                                        KTX_TEXTURE_CREATE_NO_FLAGS,
-                                        &kTexture);
+    ktxresult = ktxTexture_CreateFromNamedFile((getAssetPath() + ktxfile).c_str(), KTX_TEXTURE_CREATE_NO_FLAGS, &kTexture);
```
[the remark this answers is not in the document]
aqnuep: Yes, 132 is the line width. Note that this (and many other rules) are the default ones used across other Khronos repositories. I think we should use at least 100 characters though as 80 is just too short for modern standards.

examples/vkload.cpp
```diff
@@ -43,20 +42,14 @@ Texture::Texture(const std::string ktxfile)
...
 
-        message << "Creation of ktxTexture from \"" << getAssetPath()
-                << ktxfile << "\" failed: " << ktxErrorString(ktxresult);
+        message << "Creation of ktxTexture from \"" << getAssetPath() << ktxfile << "\" failed: " << ktxErrorString(ktxresult);
         throw std::runtime_error(message.str());
     }
-    ktxresult = ktxTexture_VkUploadEx(kTexture, &kvdi, &texture,
-                                      VK_IMAGE_TILING_OPTIMAL,
-                                      VK_IMAGE_USAGE_SAMPLED_BIT,
+    ktxresult = ktxTexture_VkUploadEx(kTexture, &kvdi, &texture, VK_IMAGE_TILING_OPTIMAL, VK_IMAGE_USAGE_SAMPLED_BIT,
```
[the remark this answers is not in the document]
aqnuep: Not sure, I'll have to look into that, but I would be surprised if clang-format wouldn't have an option for such a common style.

examples/vkload.cpp
```diff
@@ -116,9 +102,7 @@ Texture::cleanup()
     */
 }
 
-void
-Texture::prepareSamplerAndView()
-{
+void Texture::prepareSamplerAndView() {
```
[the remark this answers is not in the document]
aqnuep: So this is an on/off option. I used return type on separate line for the libktx code because that was the style, but used return type on same line for newer things like the ktx CLI tools which use more modern style. If you could provide a list of directories where you'd like us to retain the legacy "return type on separate line" style, we can just configure things accordingly.

interface/basisu_c_binding/inc/basisu_c_binding.h
```diff
@@ -4,36 +4,32 @@
 #pragma once
 
 #if defined(KHRONOS_STATIC)
-  #define KTX_BASISU_API
+#define KTX_BASISU_API
```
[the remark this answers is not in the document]
aqnuep: I'll look into whether there's a style option for that.

interface/basisu_c_binding/src/basisu_c_binding.cpp
```diff
@@ -107,37 +95,31 @@ basis_texture_type basis_file::getTextureType() {
...
+    } else {
         // Compressed formats are 2D arrays of blocks.
         const uint32_t bytes_per_block = basis_get_bytes_per_block_or_pixel(transcoder_format);
 
-        if (transcoder_format == transcoder_texture_format::cTFPVRTC1_4_RGB || transcoder_format == transcoder_texture_format::cTFPVRTC1_4_RGBA)
-        {
-            // For PVRTC1, Basis only writes (or requires) total_blocks * bytes_per_block. But GL requires extra padding for very small textures: 
-            // https://www.khronos.org/registry/OpenGL/extensions/IMG/IMG_texture_compression_pvrtc.txt
+        if (transcoder_format == transcoder_texture_format::cTFPVRTC1_4_RGB ||
+            transcoder_format == transcoder_texture_format::cTFPVRTC1_4_RGBA) {
```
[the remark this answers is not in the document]
aqnuep: I'll have to look into that too.

interface/js_binding/ktx_wrapper.cpp
aqnuep: We'll check whether there's an option for this, but please consider that even clang-format's configuration options are limited, so I cannot promise anything.

lib/etcdec.cxx
aqnuep: Thanks for the info, I didn't notice this is also an external file.

lib/gl_funcs.c
aqnuep: No, this is one of those configurable things. We can go either way. The reason why this style is the default in the Google style because technically the "*" belongs to the variable name (imagine when multiple variables are declared in a comma separated list like `int *a_ptr, b_value, *c_ptr;`).

lib/vkformat_check.c
aqnuep: clang-format modified it.

For generated files we really have various options:

  1. Not format them at all (would require moving generated files into a dedicated folder marked to not format)
  2. Format them (by explicitly calling e.g. `git clang-format` before your commit) - kind of the current proposal
  3. Integrate calling clang-format into the generation process

lib/gl_funcs.c
MarkCallow: My question here is why the difference in `basisu_c_binding.*` vs within the libktx code?

lib/vkformat_check.c
MarkCallow: I thought you were going to modify the generators to add a line that tells clang_format not to touch the file and to output in whatever format we eventually settle on.

How does clang-format get installed on one's system? My concern is for when we eventually move to generating the VkFormat related files from vk.xml in the Vulkan-Docs repo. Will #3 require installing clang-format in that repo.

aqnuep: So it is an option to not auto-format the generated files, but then they also need to be moved to a dedicated directory, because clang-format styles can only be changed (in this case disabled) per directory.

It's up to us whether we want to do that instead, but with this prototype I proposed that it may be best to just format them as other files (and after they are regenerated they are also reformatted using `git clang-format`, per normal).

MarkCallow: > but then they also need to be moved to a dedicated directory, because clang-format styles can only be changed (in this case disabled) per directory.

What happened to this?

> I still have to update the auto-generated files and their scripts to include /* clang-format off */ in order to avoid auto-formatting them,

aqnuep: Oh you're right. I completely forgot that's what we agreed on with @lexaknyazev.

MarkCallow:
Shall we close this now that #913 exists?

aqnuep:
Yes, closing.

PR #891, opened 2024-04-09
Fixes to CLI error messages

### LINKED ISSUE #867: Preceding `ktx` command name with option leads to error
The command

```
ktx --testrun extract --all --raw <input> <output>
```

produces the error

```
ktx: Unrecognized command: "extract"
```

There should not be an error in this case. Should one be unavoidable it should definitely not be saying that "extract" is an unrecognized command.

aqnuep:
Fixes #867.

CTS PR: https://github.com/KhronosGroup/KTX-Software-CTS/pull/24

### DIFF
--- a/tools/ktx/ktx_main.cpp
+++ b/tools/ktx/ktx_main.cpp
@@ -100,7 +100,7 @@ public:
 
 int Tools::main(int argc, char* argv[]) {
     cxxopts::Options options("ktx", "");
-    options.custom_help("[--version] [--help] <command> <command-args>");
+    options.custom_help("[<command>] [OPTION...]");
     options.set_width(CONSOLE_USAGE_WIDTH);
     options.add_options()
             ("h,help", "Print this usage message and exit")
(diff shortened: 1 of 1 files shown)


tools/ktx/ktx_main.cpp
```diff
@@ -100,7 +100,7 @@ class Tools : public Command {
 
 int Tools::main(int argc, char* argv[]) {
     cxxopts::Options options("ktx", "");
-    options.custom_help("[--version] [--help] <command> <command-args>");
+    options.custom_help("[<command>] [OPTION...]");
```
MarkCallow, line 103: Has the command really become optional, as implied by "[]"?

Is this message only printed for cases where there are no positional arguments?

aqnuep: The command is optional, as you can call `ktx --version` and `ktx --help` alone (potentially with `--testrun` to output the default version.

This has been the case always, it's just that the custom help did not correctly reflect this, so I unified the syntax output to match the format used by the subcommands.

aqnuep:
> Please update CTS ref.

Done. Thanks for the review!

MarkCallow:
Please rebase and remove draft status.

aqnuep:
> Please rebase and remove draft status.

Done.

PR #894, opened 2024-04-14
Add function for app to explicitly make libktx load its GL function pointers.

### LINKED ISSUE #708: ktxTexture_GLUpload fails to find GL symbols with NVIDIA driver on Linux
This is a followup for https://github.com/KhronosGroup/KTX-Software/discussions/707

I've been seeing "Could not load OpenGL command: glBindTexture!" when calling `ktxTexture_GLUpload`, although by using GLAD I can call `glBindTexture` just fine.

I reproduced the issue here: https://github.com/viseztrance/ktx-loader-issue

I also compiled the code on ubuntu 20.04 in a virtual machine and it worked there, and then tried the binary on my machine where it failed complaining about `glBindTexture`. 

Then I tried the binary on several other machines (full breakdown bellow), and I think it might be an issue related to the nvidia drivers (ver. 530.41.03)?

| Operating System | Hardware       | Status |
|------------------|----------------|--------|
| Ubuntu 20.04     | virtual machine| OK     |
| Fedora 38        | nvidia rtx 3060       | FAIL   |
| Fedora 38        | nvidia gtx 1060 (3GB)       | FAIL   |
| Fedora 38        | intel          | OK     |
| Fedora 38        | intel          | OK     |
| Fedora 33        | intel          | OK     |

Some other things I tried:
* removing glad had no effect.
* compiling KTX dynamically also had no effect

MarkCallow:
Fixes #708.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -454,6 +454,7 @@ set(BASISU_ENCODER_CXX_SRC
 
 if(KTX_FEATURE_GL_UPLOAD)
     list(APPEND KTX_MAIN_SRC
+        lib/gl_funclist.inl
         lib/gl_funcs.c
         lib/gl_funcs.h
         lib/glloader.c
--- a/cmake/docs.cmake
+++ b/cmake/docs.cmake
@@ -122,6 +122,7 @@ function( CreateDocLibKTX )
         lib/basis_transcode.cpp
         lib/miniz_wrapper.cpp
         lib/strings.c
+        lib/gl_funcs.c
         lib/glloader.c
         lib/hashlist.c
         lib/filestream.c
--- a/include/ktx.h
+++ b/include/ktx.h
@@ -905,6 +905,18 @@ struct ktxStream
  * functions.
  */
 
+typedef void (KTX_APIENTRY* PFNVOIDFUNCTION)(void);
+/**
+ * @~English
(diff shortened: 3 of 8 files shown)


MarkCallow:
@viseztrance, @Raffaello please review.

Raffaello:
@MarkCallow I have done a quick test, compiling and reinstalling this branch as a RPM on fedora.

patch libktx usage in my own code adding this line before GL_Upload: `result = ktxLoadOpenGL((PFNGLGETPROCADDRESS)SDL_GL_GetProcAddress);`


and That makes on Fedora passing my tests!  💯 👍 

So it looks very good to me! 

Problem solved from my side.

Thanks Mark.

viseztrance:
Thank you! Can also confirm it works. This works nicely (no segfault, no `glBindTexture` message):

```c
ktxLoadOpenGL(glfwGetProcAddress);
```

I noticed that if I don't call `ktxLoadOpenGL`, the code no longer segfaults even though `Could not load OpenGL command: glBindTexture!` still gets printed like before. 
(I'll still use `ktxLoadOpenGL` just to be on the safe side in my code, though).

MarkCallow:
Thank you for the confirmations @viseztrance and @Raffaello. Glad to hear it doesn't segfault anymore when the new function is not called though I don't understand exactly what is fixed in that regard.

Issue #747, opened 2023-07-31
New equivalent of ktxsc


vlucendo:
Hi,

I am trying to supercompress a texture with `VK_FORMAT_R32G32B32A32_SFLOAT` format but can't find the way to do it with the latest alpha release (`ktx encode` requires a codec parameter and the format is not compatible), therefore I still need to resort to ktxsc.

Is there a way to have the same functionality with the new releases?

Thanks!

MarkCallow:
Is this an existing texture that you want to supercompress? If so, yes the only way currently is `ktxsc`. `ktx encode` is intended for encoding to a Basis Universal format hence the required codec parameter.

We are having discussions about providing `ktxsc-like` functionality in the new tool set. This issue helps show there is demand. I'm leaning toward making a `ktx deflate` tool for supercompressing existing textures with zstd or zlib. However it might be a while. The development contract for the new tools has finished. I'd happily accept a PR to add such a tool.

vlucendo:
Thanks. Yes, this is an existing texture with some custom data I generated using [KTX-Parse](https://github.com/donmccurdy/KTX-Parse). Supercompression with `zstd` reduces it's size significantly, so maybe a `ktx deflate` tool would be a nice addition for these less common usecases.

MarkCallow:
The deflate functionality of `ktxsc` is now available in `ktx deflate` and `ktx encode` can now be used to encode ASTC files as well as BasisU files. Both are part of the v4.4.0 release.

PR #896, opened 2024-04-17
Add `ktx deflate` command

MarkCallow:
For deflating an existing ktx 2.0 file.

Fixes part of #747. A complete fix requires either `ktx encode` support ASTC or a new ktx `encode-astc` command.

### DIFF
--- a/cmake/docs.cmake
+++ b/cmake/docs.cmake
@@ -157,6 +157,7 @@ function( CreateDocTools )
         tools/ktx/ktx_main.cpp
         tools/ktx/command_compare.cpp
         tools/ktx/command_create.cpp
+        tools/ktx/command_deflate.cpp
         tools/ktx/command_encode.cpp
         tools/ktx/command_extract.cpp
         tools/ktx/command_help.cpp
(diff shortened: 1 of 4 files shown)


MarkCallow:
Questions for reviewers:

1. The command raises an error if neither --zlib or --zstd is specified. `CompressOptions` does not check for this since it is fine to not specify a scheme in either `create` or  `encode`. Should I move the check to `CompressOptions` with some kind of option to enable it or is it fine in `deflate`'s `processOptions`.
2. The command preserves previously applied options in `KTXwriterScParams` by either maintaining the original KTXwriter, if it was a member of the `ktx` suite, or appending the original writer and params to the end of the new `KTXwriterScParams`. See comment at https://github.com/KhronosGroup/KTX-Software/blob/e162e3017096869a896b98d43d5708ae1d2b4b6f/tools/ktx/command_deflate.cpp#L202 for complete details.

MarkCallow:
@lexaknyazev please review. For some reason your id doesn't show up in the potential reviewers list.

aqnuep (approved):

PR #900, opened 2024-04-25
Fix various issues surfaced by fuzzer

florczyk-fb:
This collects some fixes for issues found while fuzzing KTX parsing code through `ktxTexture_CreateFromStream`
* Integer truncation in `memstream.c` that could lead to out-of-bounds access
* Memory leaks in `texture2.c` caused by early function exits without freeing memory
* Out-of-bound read/write caused by using data read from the file without checking if the buffer has enough data
* `valueLen` becoming negative in `hashlist.c` if `keyLen==keyAndValueByteSize`


florczyk-fb (commented):
Fixed wrong ordering of check in if statement

lib/texture2.c
```diff
@@ -2556,25 +2572,27 @@ ktxTexture2_inflateZstdInt(ktxTexture2* This, ktx_uint8_t* pDeflatedData,
...
-    return KTX_SUCCESS;
-
 cleanup:
-    ZSTD_freeDCtx(dctx);
-    free(nindex);
+    if(dctx) {
+        ZSTD_freeDCtx(dctx);
+        dctx = NULL;
+    }
+    if(nindex) {
```
MarkCallow, line 2592: Another tool (can't remember which) told us it is unnecessary to check for null pointers before calling `free`. So the `if(nidex)` is not needed. Please remove it.

I think the same is true of `ZSTD_freeDCtx`.

MarkCallow:
Thank you very much for these fixes. Please add a GitHub Action to run the fuzzer. Then we can have continuous checks.

lib/texture2.c
```diff
@@ -244,6 +244,12 @@ ktxFormatSize_initFromDfd(ktxFormatSize* This, ktx_uint32_t* pDfd)
     if (KHR_DFDVAL(pBdb, VERSIONNUMBER) != KHR_DF_VERSIONNUMBER_1_3) {
         return false;
     }
+    // pDfd[0] contains the size in words as per createdfd.c writeHeader()
+    // Check if DFD buffer is large enough to hold all the declared samples
+    uint32_t samplesSize = sizeof(uint32_t) * (1 + KHR_DFDSIZEWORDS(KHR_DFDSAMPLECOUNT(pBdb)));
```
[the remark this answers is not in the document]
florczyk-fb: Oh, looks like I'm calculating and checking total size, according to createdfd.c, just named the variable incorrectly.

lib/texture2.c
```diff
@@ -246,8 +246,8 @@ ktxFormatSize_initFromDfd(ktxFormatSize* This, ktx_uint32_t* pDfd)
     }
     // pDfd[0] contains the size in words as per createdfd.c writeHeader()
```
MarkCallow, line 247: pDfd[0] contains the size in bytes.

lib/texture2.c
```diff
@@ -246,8 +246,8 @@ ktxFormatSize_initFromDfd(ktxFormatSize* This, ktx_uint32_t* pDfd)
...
     // Check if DFD buffer is large enough to hold all the declared samples
-    uint32_t samplesSize = sizeof(uint32_t) * (1 + KHR_DFDSIZEWORDS(KHR_DFDSAMPLECOUNT(pBdb)));
-    if (pDfd[0] < samplesSize) {
+    uint32_t totalSize = sizeof(uint32_t) * (1 + KHR_DFDSIZEWORDS(KHR_DFDSAMPLECOUNT(pBdb)));
```
MarkCallow, line 249: This test is not doing anything hugely useful because `KHR_DFDSAMPLECOUNT` finds the number of samples by dividing the number of words in the descriptor block, `descriptorBlockSize`,  minus the `SAMPLESTART` offset by the number of words needed by a single sample. Therefore the reported number of samples is dependent on the data size. 

It is also not robust because it is possible for a descriptor to contain more than 1 descriptor block, though, we don't have any such today, each with its own `descriptorBlockSize`. `totalSize` is the total of all those blocks plus itself. The only way to know if you have more than one descriptor block is to compare `totalSize` with `descriptorBlockSize + 1`. If more there is another sample.

With the hardcoded +1 the test only ensures `totalSize` claims at least enough space to hold a Basic Descriptor Block with possibly 0 samples.

Given the data format specification I don't see any way to make a more robust test.

Please note the limitations of the test in a comment.

MarkCallow: More clearly and simply this test is exactly equivalent to

```
if (pDfd[0] < KHR_DFDVAL(pBdb, DESCRIPTORBLOCKSIZE) + 1)
```

Before using `KHR_DFDVAL` you should check that `pDfd[0]` is at least as large as `KHR_DF_WORD_SAMPLESTART` words so you know it won't crash.

This is a minimal sanity test given the possibility of multiple descriptor blocks. A full check requires looping over possible descriptor blocks (basically as long as pDfd[0] > 4) deducting the `descriptorBlockSize` of each descriptor found. As just noted care must be taken that there is enough data left when starting to process a descriptor such that `KHR_DFDVAL(pBdb, DESCRIPTORBLOCKSIZE)` can succeed without crashing. It should also verify there is at least 1 sample in each descriptor.

It is probably a good idea to add the full test.

florczyk-fb: Thanks for the detailed explanation and pseudo-code, it makes sense to add more robust test here. I'm working on adding this test, however adding it causes few tests to faile
```
	354 - ktxToolsTest.info.invalid_file (Failed)
	409 - ktxToolsTest.validate.warning_6003_UnknownDFDBlock (Failed)
	411 - ktxToolsTest.validate.warning_6005_MultipleBDFD (Failed)
	412 - ktxToolsTest.validate.error_6006_DescriptorBlockSizeTooSmall (Failed)
	413 - ktxToolsTest.validate.error_6007_DescriptorBlockSizeTooBig (Failed)
	681 - ktxToolsTest.compare.ignore_dfd_all (Failed)
	684 - ktxToolsTest.compare.ignore_dfd_extended (Failed)
	685 - ktxToolsTest.compare.ignore_dfd_none (Failed)
	686 - ktxToolsTest.compare.ignore_bdfd_bytesplane (Failed)
```
Do you have some ideas how to tackle this?

MarkCallow: Thanks for the more robust test.

The CTS tests all compare the output from the program against golden files. I expect the more robust test is causing libktx to return a different error code and hence the tools to output different messages.

Use `ctest` to run the failing tests locally. Something like

```
ctest --test-dir build --output-on-failure -R ktxToolTests.validate
```

This will show you the output from the failed tests which will be the name of the file with the captured output and that of the golden file. You can then look at the files and see what is different.See [Conformance Test Suite](https://github.com/KhronosGroup/KTX-Software/blob/main/BUILDING.md#conformance-test-suite) for more information about setting up and running the tests from `ctest`.

If the captured output from the commands looks reasonable for the error condition you can regenerate the golden files by running the `clitest.py` command in `tests/cts/clitests`. You can also use this to run individual tests rather than using `ctest`. See [Regenerating Golden Files](https://github.com/KhronosGroup/KTX-Software-CTS/blob/main/README.md#regenerating-golden-files) for more information. Omitting `--regen-golden` from the example `clitest.py` command runs the test. You will see the same output as from `--output-on-failure`.

One you have regenerated the golden files you will need to make a PR in the [KTX-Software-CTS repo](https://github.com/KhronosGroup/KTX-Software-CTS) with the new tests. Make a new branch in `tests/cts` for the updated tests. I do not think you will be able to push this branch to KTX-Software-CTS so I think you will have to fork KTX-Software-CTS, push your new branch to your fork and make the PR from there.

Lastly you need to update the KTX-Software-CTS commit reference in KTX-Software and push it to this PR. You do this by committing `tests/cts` and pushing the result.

lib/texture2.c
```diff
@@ -235,12 +235,31 @@ bool
 ktxFormatSize_initFromDfd(ktxFormatSize* This, ktx_uint32_t* pDfd)
 {
     uint32_t* pBdb = pDfd + 1;
-
-    // Check the DFD is of the expected type and version.
-    if (*pBdb != 0) {
+    // pDfd[0] contains totalSize in bytes, check if it has at least 8 bytes
```
MarkCallow, line 238: Prefer

```
(KHR_DF_WORD_DESCRIPTORBLOCKSIZE + 1) * sizeof(uint32_t)
```

over a hard-coded 8.

lib/texture2.c
```diff
@@ -235,12 +235,31 @@ bool
...
+    if (pDfd[0] < 2 * sizeof(uint32_t) || *pBdb != 0) {
         // Either decriptorType or vendorId is not 0
         return false;
     }
+    // Iterate through all block descriptors and check if sum of their sizes
+    // is equal to the totalSize in pDfd[0]
+    uint32_t descriptorSize = pDfd[0] - sizeof(uint32_t);
+    while(descriptorSize > KHR_DF_WORD_DESCRIPTORBLOCKSIZE + 1) {
```
MarkCallow, line 246: This is wrong. My bad. Should be
```
descriptorSize > (KHR_DF_WORD_DESCRIPTORBLOCKSIZE + 1) * sizeof(uint32_t)
```

Suggest defining a constant or macro with the value `(KHR_DF_WORD_DESCRIPTORBLOCKSIZE + 1) * sizeof(uint32_t)`.

lib/texture2.c
florczyk-fb: Sorry for the delay, I regenerated test files and addressed the rest of the comments

florczyk-fb:
> Thank you very much for these fixes. Please add a GitHub Action to run the fuzzer. Then we can have continuous checks.

I'm not familiar with GitHub Actions, however this libfuzzer based harness could be used for setting up the fuzzer.

```
#include <ktx.h>

extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
  ktxTexture* texture = nullptr;
  ktxTexture_CreateFromMemory(data, size, KTX_TEXTURE_CREATE_LOAD_IMAGE_DATA_BIT, &texture);
  if (texture != nullptr) {
    ktxTexture_Destroy(texture);
  }
}
```

MarkCallow:
I've merged the test updates. Please update the CTS reference within this PR to the new HEAD commit in KTX-Software-CTS.

MarkCallow:
Thank you for your hard work on this.

MarkCallow:
> 
> I'm not familiar with GitHub Actions, however this libfuzzer based harness could be used for setting up the fuzzer.
> 
> ```
> #include <ktx.h>
> 
> extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
>   ktxTexture* texture = nullptr;
>   ktxTexture_CreateFromMemory(data, size, KTX_TEXTURE_CREATE_LOAD_IMAGE_DATA_BIT, &texture);
>   if (texture != nullptr) {
>     ktxTexture_Destroy(texture);
>   }
> }
> ```

Where is the `main()` that runs this?

How does the fuzzer determine the failures? This function is not checking the return code from ktxTexture_CreateFromMemory.  Where is the information logged?

For an Action we need a script to download the dependencies, presumably LLVMFuzzer, and a command that runs the fuzzer, displays the output and has a non-zero return code if there are issues.

florczyk-fb:
> Where is the main() that runs this?

The snippet is just a harness for testing KTX library, `main()` and the rest of fuzzer is implemented by LibFuzzer. Harness would have to be compiled by clang with `-fsanitize=fuzzer,address` options to generate fuzzer binary that could later be run.

> How does the fuzzer determine the failures? This function is not checking the return code from ktxTexture_CreateFromMemory. Where is the information logged?

Fuzzer is looking for crashes when running the binary, therefore we don't need to check for the result of `ktxTexture_CreateFromMemory` as we are only interested in crashes, if the function returns error that's still a valid program behavior from fuzzer's perspective. Additionally the binary is instrumented with AddressSanitizer to make memory safety issues easier to trigger.

https://llvm.org/docs/LibFuzzer.html

PR #913, opened 2024-05-29
Add clang-format support

MathiasMagnus:
This work if the continuation of #889.

## Notes for the reviewers

Some comments from the original work addressed here, most notably:

- Limit column size of all formatted sources to [100](https://github.com/KhronosGroup/KTX-Software/pull/889#discussion_r1573654200).

Aside from explicit comments:

- A git-blame ignore file has been committed to the repo which instructs GitHub's blame functionality on the web interface to keep attribution of lines affected by formatting. Local clones may be [configured told similarly](https://www.stefanjudis.com/today-i-learned/how-to-exclude-commits-from-git-blame/).
- The functional changes are all outside the singular "Apply initial formatting" commit.

Should other formatting changes need refining, discussions may continue here.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,4 @@
+# Copyright 2024 The Khronos Group Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+# Some note about why this hash is ignored (placeholder for initial formatting)
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -25,7 +25,11 @@ on:
   workflow_dispatch:
 
 jobs:
+  #formatting:
+  #  uses: ./.github/workflows/formatting.yml
   windows:
+    # Shortcircuit and don't burn CI time when formatting will reject
+    #needs: formatting
     strategy:
       matrix:
         os: [ windows-latest ]
--- a/external/.clang-format
+++ b/external/.clang-format
@@ -1,0 +1,7 @@
+# Copyright 2024 The Khronos Group Inc.
+# SPDX-License-Identifier: Apache-2.0
+---
+# Disable clang-format in this directory
+DisableFormat: true
+SortIncludes: false
+...
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,13 @@
+# Copyright 2024 The Khronos Group Inc.
+# SPDX-License-Identifier: Apache-2.0
+---
+# Use defaults from the Google style with the following exceptions:
+Language: Cpp
+BasedOnStyle: Google
+IndentWidth: 4
+IndentCaseLabels: false
+AccessModifierOffset: -2
+ColumnLimit: 100
+SortIncludes: false
+IndentPPDirectives: BeforeHash
+...
--- a/.github/workflows/formatting.yml
+++ b/.github/workflows/formatting.yml
@@ -1,0 +1,20 @@
+# Copyright 2024 The Khronos Group Inc.
+# Copyright 2024 RasterGrid Kft.
+# SPDX-License-Identifier: Apache-2.0
+name: Formatting
+
+on:
+  workflow_call:
+
+permissions:
+  contents: read
+
+jobs:
+  clang-format:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - name: Run clang-format
+      uses: jidicula/clang-format-action@v4.13.0
+      with:
+        clang-format-version: '17'
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -638,5 +638,17 @@ directory of your KTX-Software workarea or set the value of the
 `KTX_SPECIFICATION` CMake cache variable to the location of your specification
 clone.
 
+Formatting
+------------
+
+The KTX repository is transitioning to enforcing a set of formatting guides, checked during CI.
+The tool used for this is [ClangFormat](https://clang.llvm.org/docs/ClangFormat.html).
+To minimize friction, it is advised that one configure their environment to run ClangFormat in an automated fashion,
+minimally before committing to source control, ideally on every save.
+
+### Visual Studio Code
+
+Set the [`editor.formatOnSave`](https://code.visualstudio.com/docs/editor/codebasics#_formatting) option and use one of the C/C++ formatting extensions available, most notably [ms-vscode.cpptools](https://marketplace.visualstudio.com/items?itemName=ms-vscode.cpptools) or [llvm-vs-code-extensions.vscode-clangd](https://marketplace.visualstudio.com/items?itemName=llvm-vs-code-extensions.vscode-clangd).
+
 
 {# vim: set ai ts=4 sts=4 sw=2 expandtab textwidth=75:}
--- a/external/fmt/.clang-format
+++ b/external/fmt/.clang-format
@@ -1,8 +1,0 @@
-# Run manually to reformat a file:
-# clang-format -i --style=file <file>
-Language: Cpp
-BasedOnStyle: Google
-IndentPPDirectives: AfterHash
-IndentCaseLabels: false
-AlwaysBreakTemplateDeclarations: false
-DerivePointerAlignment: false
--- a/include/.clang-format
+++ b/include/.clang-format
@@ -1,0 +1,7 @@
+# Copyright 2024 The Khronos Group Inc.
+# SPDX-License-Identifier: Apache-2.0
+---
+# Disable clang-format in this directory
+DisableFormat: true
+SortIncludes: false
+...
--- a/lib/.clang-format
+++ b/lib/.clang-format
@@ -1,0 +1,14 @@
+# Copyright 2024 The Khronos Group Inc.
+# SPDX-License-Identifier: Apache-2.0
+---
+# Use defaults from the Google style with the following exceptions:
+Language: Cpp
+BasedOnStyle: Google
+IndentWidth: 4
+IndentCaseLabels: false
+AccessModifierOffset: -2
+ColumnLimit: 100
+SortIncludes: false
+AlwaysBreakAfterDefinitionReturnType: TopLevel
+IndentPPDirectives: BeforeHash
+...
--- a/lib/vkformat_check.c
+++ b/lib/vkformat_check.c
@@ -9,6 +9,7 @@
 ** SPDX-License-Identifier: Apache-2.0
 */
 
+// clang-format off: CI is complicated if formatting checks on generated files are enforced.
 #include <stdint.h>
 #include <stdbool.h>
 
@@ -211,3 +212,4 @@ isValidFormat(VkFormat format)
     }
 }
 
+// clang-format on
--- a/lib/vkformat_enum.h
+++ b/lib/vkformat_enum.h
@@ -1,3 +1,4 @@
+// clang-format off: CI is complicated if formatting checks on generated files are enforced.
 #if !defined(_VKFORMAT_ENUM_H_) && !defined(VULKAN_CORE_H_)
 #define _VKFORMAT_ENUM_H_
 
@@ -312,3 +313,4 @@ typedef uint64_t VkFlags64;
 #define VK_FORMAT_MAX_STANDARD_ENUM 184
 
 #endif /* _VKFORMAT_ENUM_H_ */
+// clang-format on
--- a/lib/vkformat_str.c
+++ b/lib/vkformat_str.c
@@ -9,6 +9,7 @@
 ** SPDX-License-Identifier: Apache-2.0
 */
 
+// clang-format off: CI is complicated if formatting checks on generated files are enforced.
 
 #include <stdint.h>
 #include <ctype.h>
@@ -1292,3 +1293,4 @@ stringToVkFormat(const char* str)
         return VK_FORMAT_R16G16_SFIXED5_NV;
     return VK_FORMAT_UNDEFINED;
 }
+// clang-format on
--- a/lib/vkformat_typesize.c
+++ b/lib/vkformat_typesize.c
@@ -9,6 +9,7 @@
 ** SPDX-License-Identifier: Apache-2.0
 */
 
+// clang-format off: CI is complicated if formatting checks on generated files are enforced.
 
 #include <stdint.h>
 
@@ -582,3 +583,4 @@ vkFormatTypeSize(VkFormat format)
         return 0;
     }
 }
+// clang-format on
--- a/other_include/.clang-format
+++ b/other_include/.clang-format
@@ -1,0 +1,7 @@
+# Copyright 2024 The Khronos Group Inc.
+# SPDX-License-Identifier: Apache-2.0
+---
+# Disable clang-format in this directory
+DisableFormat: true
+SortIncludes: false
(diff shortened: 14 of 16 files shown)


.github/workflows/windows.yml
```diff
@@ -25,7 +25,11 @@ on:
   workflow_dispatch:
 
 jobs:
+  formatting:
+    uses: ./.github/workflows/formatting.yml
   windows:
+    # Shortcircuit and don't burn CI time when formatting will reject
+    needs: formatting
```
MarkCallow, line 32: Is this forcing CI to always run formatting before a build? If so, how? The workflow name is "Formatting".

.git-blame-ignore-revs
```diff
@@ -0,0 +1,5 @@
+# Copyright 2024 The Khronos Group Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+# Apply initial formatting
+4ceab17e94e73c918130a49ad1b21d5e4c14e7ec
```
MarkCallow, line 5: Isn't this ignoring just this single commit? What about future reformatting that takes place when commit(s) are pushed?

interface/basisu_c_binding/inc/basisu_c_binding.h
```diff
@@ -4,36 +4,32 @@
 #pragma once
 
 #if defined(KHRONOS_STATIC)
-  #define KTX_BASISU_API
+#define KTX_BASISU_API
```
MarkCallow, line 7: I prefer to keep the PP directives indented. I see from fmt's .clang_format that it is possible:

`IndentPPDirectives: BeforeHash`

will do it.

.github/workflows/windows.yml
MathiasMagnus: The workflow name is "Formatting" indeed, but that is only the display name of the workflow if it were to run by itself, which it doesn't, because the only `on` flavor is [`on.workflow_call`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_call), which means only other workflows will call it (aka. [reusable worfklow](https://docs.github.com/en/actions/using-workflows/reusing-workflows)). The calling workflow will give it a new name (if it doesn't, it'll be an ugly yml filename). The `needs` part has to refer to to the [`jobs.<job_id>`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_id) of the workflow being called in the _caller's_ yaml file.

For an example of how such a workflow executes, check [this](https://github.com/MathiasMagnus/Test-Packaging/actions/runs/9414364871) run from a toy repo of mine. Hope this clears it up.

.git-blame-ignore-revs
MathiasMagnus: Thanks for reminding me, this commit hash needs update after rebases I done.

Ideally no further commits need to be added here, because now the entire repo is formatted and all formatting violations will trip CI, hence no pure formatting-related commits need to be made. Contributors are highly encouraged to configure their development environment to format on save. (eg. VS Code has [`editor.formatOnSave`](https://code.visualstudio.com/docs/editor/codebasics#_formatting) option and multiple C/C++ formatting extensions available, most notably [ms-vscode.cpptools](https://marketplace.visualstudio.com/items?itemName=ms-vscode.cpptools) or [llvm-vs-code-extensions.vscode-clangd](https://marketplace.visualstudio.com/items?itemName=llvm-vs-code-extensions.vscode-clangd))

interface/basisu_c_binding/inc/basisu_c_binding.h
MathiasMagnus: Added.

MathiasMagnus:
I think this work is complete. The current CI failure doesn't seem to be related to the changes made here.

.github/workflows/windows.yml
MarkCallow: I think I see. `needs:` refers to the job name `formatting` within the windows.yml file. Correct? The workflow name given in formatting.yml is unused here. Correct?

MathiasMagnus: Yes.

.git-blame-ignore-revs
MarkCallow: Please resolve this when you've updated the commit hash.

MathiasMagnus: It's been updated with the latest rebase.

external/.clang-format
```diff
@@ -0,0 +1,7 @@
+# Copyright 2024 The Khronos Group Inc.
+# SPDX-License-Identifier: Apache-2.0
+---
+# Disable clang-format in this directory
+DisableFormat: true
+SortIncludes: false
```
MarkCallow, line 6: Why is this needed when formatting is disabled?

interface/basisu_c_binding/src/basisu_c_binding.cpp
```diff
@@ -150,56 +137,53 @@ uint32_t basis_file::getImageTranscodedSizeInBytes(uint32_t image_index, uint32_
...
 
-        if (transcoder_format == transcoder_texture_format::cTFPVRTC1_4_RGB || transcoder_format == transcoder_texture_format::cTFPVRTC1_4_RGBA)
-        {
-            // For PVRTC1, Basis only writes (or requires) total_blocks * bytes_per_block. But GL requires extra padding for very small textures: 
+        if (transcoder_format == transcoder_texture_format::cTFPVRTC1_4_RGB ||
+            transcoder_format == transcoder_texture_format::cTFPVRTC1_4_RGBA) {
+            // For PVRTC1, Basis only writes (or requires) total_blocks * bytes_per_block. But GL
+            // requires extra padding for very small textures:
             // https://www.khronos.org/registry/OpenGL/extensions/IMG/IMG_texture_compression_pvrtc.txt
             // The transcoder will clear the extra bytes followed the used blocks to 0.
```
MarkCallow, line 188: Unrelated to this PR but given the overhead of running a PR please fix the typo. s/followed/following/.

lib/astc_encode.cpp
```diff
@@ -35,19 +35,19 @@
...
-#include <pthread.h>
+    #include <pthread.h>
 #else
 // Provide pthreads support on windows
-#define WIN32_LEAN_AND_MEAN
-#include <Windows.h>
+    #define WIN32_LEAN_AND_MEAN
+    #include <Windows.h>
 
 typedef HANDLE pthread_t;
```
MarkCallow, line 44: Can you get it to indent this code up to the #endif?

tests/loadtests/geom/cube.h
```diff
@@ -3,70 +3,47 @@
  * SPDX-License-Identifier: Apache-2.0
```
MarkCallow, line 3: Either don't format this file or fix the formatting for it so the vertices remain separated in the cube face - the formatting has made the comments nonsensical plus it is much harder to see the vertices when they are all run together.

tests/loadtests/geom/cube_data.h
```diff
@@ -10,146 +10,92 @@
 // Mesh and VertexFormat Data
```
MarkCallow, line 10: Also don't format this.

tests/loadtests/geom/frame.h
```diff
@@ -6,18 +6,6 @@
 /*
```
MarkCallow, line 6: Or this.

tests/loadtests/geom/frame.h
```diff
@@ -6,18 +6,6 @@
...
  * Frame definition
```
MarkCallow, line 7: Or this.

tests/loadtests/geom/quad.h
```diff
@@ -9,26 +9,10 @@
 /*
```
MarkCallow, line 9: Or this.

tests/unittests/unittests.cc
```diff
@@ -85,31 +84,30 @@ class CheckHeader1Test : public ::testing::Test {
...
     KTX_header testHeader;
+
   public:
     static ktx_uint8_t ktxId[12];
 };
 
-ktx_uint8_t CheckHeader1Test::ktxId[12] = {
-    0xAB, 0x4B, 0x54, 0x58, 0x20, 0x31, 0x31, 0xBB, 0x0D, 0x0A, 0x1A, 0x0A
-};
+ktx_uint8_t CheckHeader1Test::ktxId[12] = {0xAB, 0x4B, 0x54, 0x58, 0x20, 0x31,
```
MarkCallow, line 92: Is there a way to fix the format of a few lines in a file? This represents the signature of a KTX v1 file so it much easier for humans to follow as originally written.

.git-blame-ignore-revs
MarkCallow: Sorry. I edited my original comment that said "please resolve" to add the request about documentation which you probably missed. Probably `BUILDING.md` is the best place for the documentation.

MarkCallow:
You need to rebase or merge `main` again to pick up the workaround for the build issue on GitHub Actions. The latest runner image is only partially fixed.

MarkCallow:
@MathiasMagnus please address the outstanding issues and complete this work.

.git-blame-ignore-revs
MathiasMagnus: Documented VS Code. I looked into (Neo)vim, but the phase space is too large.

external/.clang-format
MathiasMagnus: It's probably not strictly necessary.

MathiasMagnus:
@MarkCallow We created this PR out of courtesy, because this task did not fit in the original project timeline, so we are unable to address all the specific personal styling and tooling requests. I added documentation on the minimal requirement to reduce CI friction and noted how to configure VS Code to not trip CI. Vim is too deep a rabbit a hole and I don't have the time to give anything I document in XCode a test drive.

As for the rest of the outstanding formatting related inquiries: instead of whack-a-mole-ing the formatting issues one-by-one and aligning them to the desired format, I reverted the "Initial formatting" commit and just commit the machinery enabling the enforced formatting.

- For now, CI job definition is there, but its invocation is commented out.
- The GitHub-friendly git blame ignore file is there with a note but not hash.

Because clang-format can't run on an entire source tree, all the tools drive it file-by-file. Should you want to tweak the `.clang-format` files to perfection, I used the following one-liner on Windows:

```pwsh
# list folders excluding build and .venv | recurse into all | filter files matching regex | for each of the matches { run clang-format inplace using style files and full path to file }
gci -exc build,.venv | gci -rec | where -prop Name -match '^.*\.((((c|C)(c|pp|xx|\+\+)?$)|((h|H)h?(pp|xx|\+\+)?$))|(ino|pde|proto|cu))$' | % { clang-format.exe -i --style=file $_.FullName }
```

I leave implementing similar logic in the shell of their choice to the end-user. (The regex I borrowed from the [script run by the CI action](https://github.com/jidicula/clang-format-action/blob/4dacb6087962f49cedccf872a80c610db1e0e02a/check.sh#L77).)

In summary, all the infrastructure for clang-format is there, please tweak it for your needs as you see fit.

MarkCallow:
I am not happy ...

> We created this PR out of courtesy, because this task did not fit in the original project timeline, so we are unable to address all the specific personal styling and tooling requests.

I thought providing a PR with initial styling was part of the contract. How else would you deliver the project work, if not by a PR?

> Instead of whack-a-mole-ing the formatting issues one-by-one and aligning them to the desired format, I reverted the "Initial formatting" commit and just commit the machinery enabling the enforced formatting.

There is no whack-a-mole on formatting issues here. The related open conversations are everything. I count 2 format change requests and requests to exclude 4 files from formatting. That is not a huge amount of work.

Please restore the formatting making the small number of requested changes.

I understand that documenting this for vim and xcode is potentially time-consuming, if you don't already have the knowledge. Omitting that part  of my request is okay.

aqnuep:
> I thought providing a PR with initial styling was part of the contract. How else would you deliver the project work, if not by a PR?

Yes, but the contract ended at the end of March with the understanding that due to the additional requests we ran out of time and budget to do the clang-format change. We simply offered to put together a PR anyway afterwards, out of courtesy, as we had free time for it, but we cannot go back-and-forth for months to make the styling fit anybody's personal preferences.

All the infrastructure work is there. Actual styling is subjective and a usual rabbit-hole. You should easily tweak the configurations for your needs, and continuing this back-and-forth is not something that we can continue doing forever.

lib/astc_encode.cpp
aqnuep: You can exclude lines of code from being formatted by surrounding them with:

```cpp
// clang-format off
...
// clang-format on
```

There are examples of this in the generated files.

external/.clang-format
aqnuep: This is the convention used in other Khronos repositories, but it should have no effects (at least on clang-format versions I've used).

tests/loadtests/geom/cube.h
aqnuep: There is no way to exclude individual files from formatting. That's why we had to move everything external that is not expected to be formatted by us into a separate subdirectory where we disabled formatting.

You can exclude lines of code from being formatted (or entire files) by surrounding them with:

```cpp
// clang-format off
...
// clang-format on
```

There are examples of this in the generated files.

tests/loadtests/geom/cube_data.h
aqnuep: See above.

tests/loadtests/geom/frame.h
aqnuep: See above.

tests/loadtests/geom/quad.h
aqnuep: See above.

tests/unittests/unittests.cc
aqnuep: You can exclude lines of code from being formatted by surrounding them with:

```cpp
// clang-format off
...
// clang-format on
```

There are examples of this in the generated files.

aqnuep:
> There is no whack-a-mole on formatting issues here. The related open conversations are everything. I count 2 format change requests and requests to exclude 4 files from formatting. That is not a huge amount of work.

Please also consider that rebasing a formatting change is a nightmare. We've undone the one-shot formatting, you can adjust the styling as needed, and re-run clang-format on all C++ files then merge it once you're happy with the output. It should be trivial from this point.

MarkCallow:
> the understanding that due to the additional requests we ran out of time and budget to do the clang-format change

Were these "additional requests" related or unrelated to clang-format? If the former what where they?

> There is no way to exclude individual files from formatting. 

> Because clang-format can't run on an entire source tree, all the tools drive it file-by-file.

These statements conflict. What am I missing?

MarkCallow:
> There is no way to exclude individual files from formatting.

What about .clang-format-ignore which I've just discovered at https://clang.llvm.org/docs/ClangFormat.html.

aqnuep:
> > the understanding that due to the additional requests we ran out of time and budget to do the clang-format change
> 
> Were these "additional requests" related or unrelated to clang-format? If the former what where they?
> > There is no way to exclude individual files from formatting.
> 
> > Because clang-format can't run on an entire source tree, all the tools drive it file-by-file.
> 
> These statements conflict. What am I missing?

PR #917, opened 2024-06-05
docs: clarify pointer lifetimes for ktxTexture_SetImageFromMemory

### LINKED ISSUE #914: ktxTexture_SetImageFromMemory should document pointer lifetime guarantees
`ktxTexture_SetImageFromMemory` (and friends) expect a `const ktx_uint8_t*` to the image source in memory. It's unclear if this takes over ownership of the pointer or if it creates a copy of the underlying image memory. Upon further investigation, it appears it [creates a copy](https://github.com/KhronosGroup/KTX-Software/blob/5fc739c8fc577946cf91509509ac189e81143d2a/lib/writer2.c#L232) of the image memory. 

To better help users avoid memory leaks (by e.g. assuming a memory ownership transfer), corresponding KTX APIs should be updates to include a note about pointer lifetime assumptions.

matthew-rister:
**Problem** 
`ktxTexture1_SetImageFromMemory` and `ktxTexture2_SetImageFromMemory` accept a pointer to image memory but do not clearly document if ownership of those pointers is transferred or copied. 

**Solution**
This commit adds a Doxygen note informing users that the underlying image data is copied and that they are responsible for freeing the original copy of image memory.

**Testing**
Documentation updates were manually verified by building the project using the `KTX_FEATURE_DOC=ON` CMake option and viewing the generated Doxygen output using a local server.

Closes #914

### DIFF
--- a/lib/writer2.c
+++ b/lib/writer2.c
@@ -299,6 +299,9 @@ ktxTexture2_SetImageFromStdioStream(ktxTexture2* This, ktx_uint32_t level,
  * Level, layer, faceSlice rather than offset are specified to enable some
  * validation.
  *
+ * @note The caller is responsible for freeing the original image memory
+ *       referred to by @p src.
+ *
  * @param[in] This      pointer to the target ktxTexture object.
  * @param[in] level     mip level of the image to set.
  * @param[in] layer     array layer of the image to set.
--- a/lib/writer1.c
+++ b/lib/writer1.c
@@ -206,6 +206,9 @@ ktxTexture1_SetImageFromStdioStream(ktxTexture1* This, ktx_uint32_t level,
  *
  * @warning Do not use @c memcpy for this as it will not pad when necessary.
  *
+ * @note The caller is responsible for freeing the original image memory
+ *       referred to by @p src.
+ *
  * @param[in] This      pointer to the target ktxTexture object.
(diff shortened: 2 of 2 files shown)


lib/writer2.c
```diff
@@ -292,6 +292,9 @@ ktxTexture2_SetImageFromStdioStream(ktxTexture2* This, ktx_uint32_t level,
  * Level, layer, faceSlice rather than offset are specified to enable some
  * validation.
  *
+ * @warning Client applications are responsible for freeing the original copy of
+ *          image memory referred to by \p src.
```
MarkCallow, line 296: These requests apply both to here and `writer1.c`.

Please change `\p` to `@p` and change the text to

"The caller is responsible for freeing the original image memory referred to by @p src."

Both changes are for consistency with the rest of the documentation.

I think this should be `@note` not `@warning` because the `const` already implies that the function will not change the pointed-at content and therefore the caller is responsible for freeing it.

MarkCallow:
Sorry for my delay reviewing this. Work was hampered for more than a week by a error in the GitHub Actions Windows runner image.

lib/writer2.c
matthew-rister: Thank you for the review. I've updated the docs with your suggested changes.

PR #936, opened 2024-09-05
Fixing build on Ubuntu 24.04 w/ Clang 20

Honeybunch:
Some compiler implementations seem to be very adamant that `std::unique_ptr` lives in `<memory>` and without this change I was unable to build.


MarkCallow:
Thank you for this fix.

PR #950, opened 2024-10-04
Attempt to workaround venv failure in Actions runner 20240929.1.

MarkCallow:


ShukantPal:
Hi @MarkCallow,

It seems like the new images don't have virtualenv pre-installed. You can install it using `pip` before the build step.

https://virtualenv.pypa.io/en/latest/installation.html

MarkCallow:
> It seems like the new images don't have virtualenv pre-installed. You can install it using `pip` before the build step.

That is what I thought so the last but one commit installs virtualenv via `pipx`. The installation was successful but I still get the error. The latest commit tries to invoke `activate` but it is not working. I think I don't have the right path for Windows. I look further into it tomorrow.

BUT, the [change log](https://github.com/actions/runner-images/releases/tag/win22%2F20240929.1) for the Windows runner image shows no changes whatever to Python components.

ShukantPal:
Try using `pip` instead of `pipx` so that `virtualenv` is installed as a global Python module. You can see it being referenced using `python3 -m virtualenv` here: https://github.com/KhronosGroup/KTX-Software/blob/797098b4b9c5d415906a292365eae7ef0958443e/external/fmt/doc/build.py#L23

MarkCallow:
`pip` worked. Thanks for the tip.

These are the commands in the build that are triggering the virtualenv requirement:
```
${PYTHON_EXECUTABLE} -m build --sdist --outdir ${DIST_DIR}

${PYTHON_EXECUTABLE} -m build --wheel --outdir ${DIST_DIR}
```

Building the requirements listed in `requirements.txt` does not require virtualenv.

Issue #812, opened 2023-12-04
ktx encode and create lost --normalize option


wasimabbas-arm:
Not sure why `--normalize` was removed from `encode` and `create` but its still referred to in the docs within `--normal-mode`

```
     --normal-mode      Optimizes for encoding textures with normal data. If the input texture has
                         three or four linear components it is assumed to be a three component
                         linear normal map storing unit length normals as (R=X, G=Y, B=Z). A fourth
                         component will be ignored. The map will be converted to a two component
                         X+Y normal map stored as (RGB=X, A=Y) prior to encoding. If unsure that
                         your normals are unit length, use --normalize. If the input has 2 linear
                         components it is assumed to be an X+Y map of unit normals.
                         The Z component can be recovered programmatically in shader code by using
                         the equations:
                             nml.xy = texture(...).ga;              // Load in [0,1]
                             nml.xy = nml.xy * 2.0 - 1.0;           // Unpack to [-1,1]
                             nml.z = sqrt(1 - dot(nml.xy, nml.xy)); // Compute Z
                         ETC1S / BasisLZ encoding, RDO is disabled (no selector RDO, no endpoint
                         RDO) to provide better quality.
```

MarkCallow:
I missed this. @lexaknyazev was this a deliberate omission? If so, why? The normalize code is still in the image processing library.

MarkCallow:
> was this a deliberate omission?

Ping @lexaknyazev. Please respond. Do you have any objection to it being restored?

MarkCallow:
> > was this a deliberate omission?
> 
> Ping @lexaknyazev. Please respond. Do you have any objection to it being restored?

Ping @lexaknyazev again. I will ask @wasimabbas-arm to restore it. If you have an objection, speak now.

lexaknyazev:
The spec for the CLI tool had a limited scope so some features were excluded.

To add it now, one would have to go through all currently possible input/output parameter and payload configurations and define reasonable normalization behavior for them or explicitly declare certain combinations invalid.

MarkCallow:
> To add it now, one would have to go through all currently possible input/output parameter and payload configurations and define reasonable normalization behavior for them or explicitly declare certain combinations invalid.

Yes we need to do this. It is mostly about the input I think. For example it should be invalid for any non-linear input.

MarkCallow:
[...]

PR #952, opened 2024-10-19
Encode astc support

wasimabbas-arm:
Updated version of #810.

Since #810 was adding a separate tool. Its easier to start another cleaner PR.

This PR has:
    Commit to add --format option and make it exclusive with --codec
    Add ASTC support to encode


lib/astc_encode.cpp
```diff
@@ -807,3 +890,232 @@ ktxTexture2_CompressAstc(ktxTexture2* This, ktx_uint32_t quality) {
 
     return ktxTexture2_CompressAstcEx(This, &params);
 }
+
+struct decompression_workload
```
MarkCallow, line 894: I don't like having the decoder in a file called `astc_encode.cpp`. I know we talked about this before but I don't recall the answer to "is there a lot of code in this file shared between encode and decode?" If not then I'd prefer the decoder in a separate file. If there is then consider renaming the file to `astc_codec.cpp`.

MarkCallow (commented):
Thanks @wasimabbas-arm. Can you now get metrics for ASTC compressed files? 

You will need to add tests of the new functionality to the tools CTS and possibly need to update some existing tests.

lib/astc_encode.cpp
wasimabbas-arm: > is there a lot of code in this file shared between encode and decode?
The answer was/is yes. I think better rename the file.

tools/ktx/metrics_utils.h
```diff
@@ -101,11 +101,22 @@ class MetricsCalculator {
...
-        const auto tSwizzleInfo = determineTranscodeSwizzle(texture, report);
+        // Start with a default swizzle
+        TranscodeSwizzleInfo  tSwizzleInfo{};
+        tSwizzleInfo.defaultNumComponents = 4;
+        tSwizzleInfo.swizzle = "rgba";
 
         ktx_error_code_e ec = KTX_SUCCESS;
         // Decode the encoded texture to observe the compression losses
-        ec = ktxTexture2_TranscodeBasis(texture, KTX_TTF_RGBA32, 0);
+        if (isFormatAstc((VkFormat)texture->vkFormat))
```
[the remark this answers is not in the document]
wasimabbas-arm: Fixed in https://github.com/KhronosGroup/KTX-Software/pull/952/commits/8f4b5a65c9fc29f67b1cf14be36802dbfa4f5358

lib/astc_encode.cpp
wasimabbas-arm: Fixed in https://github.com/KhronosGroup/KTX-Software/pull/952/commits/8f4b5a65c9fc29f67b1cf14be36802dbfa4f5358

tools/ktx/command_encode.cpp
```diff
@@ -138,13 +143,35 @@ int CommandEncode::main(int argc, char* argv[]) {
 
 void CommandEncode::OptionsEncode::init(cxxopts::Options& opts) {
     opts.add_options()
-        ("codec", "Target codec."
+        (kFormat, "KTX format enum that specifies the KTX file output format."
+                  " The enum names are matching the VkFormats without the VK_FORMAT_ prefix."
+                  " The VK_FORMAT_ prefix is ignored if present."
+                  "\nIt can't be used with --codec and is only valid for ASTC encoding."
```
MarkCallow, line 149: Delete "is only valid for ASTC encoding" as the start of the next sentence basically repeats this.

tools/ktx/command_encode.cpp
```diff
@@ -220,9 +251,18 @@ void CommandEncode::executeEncode() {
...
+
+    if (options.vkFormat != VK_FORMAT_UNDEFINED) {
+       options.mode = KTX_PACK_ASTC_ENCODER_MODE_LDR; // TODO: Fix me for HDR textures
+       ret = ktxTexture2_CompressAstcEx(texture, &options);
+       if (ret != KTX_SUCCESS)
+           fatal(rc::IO_FAILURE, "Failed to encode KTX2 file with codec \"{}\". KTX Error: {}", ktxErrorString(ret));
+    } else {
+       ret = ktxTexture2_CompressBasisEx(texture, &options);
+       if (ret != KTX_SUCCESS)
+           fatal(rc::IO_FAILURE, "Failed to encode KTX2 file with codec \"{}\". KTX Error: {}", ktxErrorString(ret));
```
MarkCallow, line 263: The codec type argument is missing.

tools/ktx/command_encode.cpp
```diff
@@ -220,9 +251,18 @@ void CommandEncode::executeEncode() {
...
     metrics.saveReferenceImages(texture, options, *this);
-    ret = ktxTexture2_CompressBasisEx(texture, &options);
-    if (ret != KTX_SUCCESS)
-        fatal(rc::IO_FAILURE, "Failed to encode KTX2 file with codec \"{}\". KTX Error: {}", ktxErrorString(ret));
+
+    if (options.vkFormat != VK_FORMAT_UNDEFINED) {
+       options.mode = KTX_PACK_ASTC_ENCODER_MODE_LDR; // TODO: Fix me for HDR textures
+       ret = ktxTexture2_CompressAstcEx(texture, &options);
+       if (ret != KTX_SUCCESS)
+           fatal(rc::IO_FAILURE, "Failed to encode KTX2 file with codec \"{}\". KTX Error: {}", ktxErrorString(ret));
```
MarkCallow, line 259: Since this is for ASTC the message should say '... encode KTX2 file to \"{}\"'. Also the codec type (in this case format type) argument is missing.

MarkCallow (changes requested):
Some minor nits.

MarkCallow:
I've merged the CTS changes. They only update some messages. Please add tests that use encode to encode KTX2 files to ASTC.

abbaswasim:
@MarkCallow thanks. I will make the rest of the changes.

Can I just ask. I know you have mentioned 

> Thanks @wasimabbas-arm. Can you now get metrics for ASTC compressed files?
> 
> You will need to add tests of the new functionality to the tools CTS and possibly need to update some existing tests.

But every time I have come back to this PR I wasn't sure how many things needs doing (apart from fixing those tests of course). And thats put me off a bit, and ended up procrastinating.

Is there a complete list of things that I need to do for this? I am hoping to go through all at once to reduce lots of back and forth.

At the moment I have 
1. Write tests for new functionality (How many tests are enough? It's already tested in a way, as PR#31 in CTS). Please provide details of what else needs testing.
2. Get metrics for ASTC compress files (Could you describe what this exactly means?)
3. what else needs doing? (Maybe the doxygen docs are wrong?)

tools/ktx/command_encode.cpp
wasimabbas-arm: I didn't think so. This says `--codec` and `--format` are mutualy exclusive options. The following says:

```
                  "\nThe format must be an ASTC format. When specified the ASTC encoder specific"
                  " options becomes valid."
```

This to me just says the format must be ASTC format, it doesn't talk about mutual exclusion with `--codec`

MarkCallow: Yes. I am not suggesting you remove "It can't be used with --codec", only the latter half of the sentence, which is what is repeated in the next sentence.

abbaswasim: I see. Happy to remove what you suggest.

Just adding my 2ps. I still think saying that the `--format` is only valid for ASTC encoding is still useful. Because here we are talking about the "option" itself while later we are talking about the "value" of the option. I am sure one can deduce this from "format needs to be only ASTC format means its not relevant for anything else bit" but thinking about someone who doesn't know this could trip on that.

MarkCallow: When I read it, it feels duplicative. The whole piece is talking about the --format option.

Change "The format must be" to "The value must be". I think then it is fine.

wasimabbas-arm: Fixed in https://github.com/KhronosGroup/KTX-Software/pull/952/commits/3645a7aff46fda0206c2bd32f4f38beaec26b26a

tools/ktx/command_encode.cpp
wasimabbas-arm: Fixed in https://github.com/KhronosGroup/KTX-Software/pull/952/commits/3645a7aff46fda0206c2bd32f4f38beaec26b26a

tools/ktx/command_encode.cpp
wasimabbas-arm: Fixed in https://github.com/KhronosGroup/KTX-Software/pull/952/commits/3645a7aff46fda0206c2bd32f4f38beaec26b26a

wasimabbas-arm:
Not sure if the last commit is correct or not. Maybe I need to wait for the https://github.com/KhronosGroup/KTX-Software-CTS/pull/33 to be merged before I can update the submodule

tools/ktx/command_encode.cpp
MarkCallow: "and is only valid for ASTC encoding" is still present. Please remove it.

MarkCallow:
> I'll address the rest of the questions in your comment tomorrow.

Sorry. It's two days later. Back to your questions.

> Get metrics for ASTC compress files (Could you describe what this exactly means?)

I was asking if you had implemented the feature of printing metrics on the loss from the encoding. Looking at the code, you seem to have done so. The important thing is decoding the ASTC-compressed images so they can be compared with the reference images. You've done that. The rest of the code in metrics_utils.h is the same regardless whether the images came from BasisU or ASTC encoding. We need tests though. Please add tests that encode to ASTC and print out the metrics. (It is possible you have done so and I missed them when reviewing the CTS PR.)

> what else needs doing? (Maybe the doxygen docs are wrong?)

At the moment I do not think there is anything else beyond addressing the issues here and in the CTS PR. I'll review the docs again but what I looked at so far seemed fine.

MarkCallow:
> Please add tests that encode to ASTC and print out the metrics.

Please modify `command_create.cpp` as necessary so the metrics options work for ASTC as well as BasisU and add tests for the functionality. It previously only supported metrics for BasisU.

MarkCallow:
I just found one more thing that needs to be done. The documentation  for `--normal-mode` in encode_utils_common.h says "use `--normalize` if unsure that the input has unit normals". However there is no normalize option (that I could find. Something weird is going on with my Xcode find at present). The image class has implementations so it should just be a question of adding the option and when set calling the function on the image.

wasimabbas-arm:
> > 
> 
> > Not sure if the last commit is correct or not. Maybe I need to wait for the [KhronosGroup/KTX-Software-CTS#33](https://github.com/KhronosGroup/KTX-Software-CTS/pull/33) to be merged before I can update the submodule
> 
> If you updated the submodule reference to the commit that is in the PR, it should be fine.

MarkCallow:
List to track the requests I've made that aren't comments on existing code.

- [ ] Uncomment references in `ktx create` documentation to encode being able to encode to ASTC, now that it is.
- [ ] Enable printing metrics when encoding to ASTC in `ktx create`
- [ ] Add `--normalize` option to `ktx create`
- [ ] Add test to ensure passing an ASTC encoded KTX2 file to `ktx encode` is a no-op.

tools/ktx/command_create.cpp
MarkCallow, line 1: You also need to modify https://github.com/KhronosGroup/KTX-Software/blob/a0405a033d3669a4cc18c33a21276d58610ac5d0/tools/ktx/command_create.cpp#L1228 so it calls

```
    metrics.saveReferenceImages(texture, options, *this);
```

and

```
    metrics.decodeAndCalculateMetrics(texture, options, *this);
```

abbaswasim: yea sorry, not done yet, there are tests etc coming up.

wasimabbas-arm: Fixed in https://github.com/KhronosGroup/KTX-Software/pull/952/commits/15ba0e6c1e190c7da080acca159d1f5e05433243

wasimabbas-arm:
>     * [ ]  Add `--normalize` option to `ktx create`


@MarkCallow Please refer to https://github.com/KhronosGroup/KTX-Software/issues/812. Its currently is still the case that both `create` and `encode` has lost the option. 

I would argue this is out of scope for this PR. Lets conclude 812 before we proceed.

tools/ktx/command_create.cpp
wasimabbas-arm: I have to revert this change in https://github.com/KhronosGroup/KTX-Software/pull/952/commits/b0ce40330ac6cded2405224fa6707b8605dc1715 and add it at call site. It was creating a bug.

MarkCallow:
The reuse issue has cleared up on its own. I've no idea what happened. The latest version ran fine for me locally.

MarkCallow (approved):

MarkCallow:
> > ```
> > * [ ]  Add `--normalize` option to `ktx create`
> > ```
> 
> @MarkCallow Please refer to #812. Its currently is still the case that both `create` and `encode` has lost the option.
> 
> I would argue this is out of scope for this PR. Lets conclude 812 before we proceed.

I agree it is out of scope for this PR. I've pinged Alexey again in #812. My view is normalize should be restored.

MarkCallow:
Please update the CTS reference then I can merge this.

MarkCallow:
Thanks @wasimabbas-arm.

PR #957, opened 2024-11-12
Fix JNI function name for `getDataSizeUncompressed`

javagl:
A small fix for the JNI bindings:

 The function signature for `getDataSizeUncompressed` was given as 
`Java_org_khronos_KTXTexture_getDataSizeUncompressed` but must be
`Java_org_khronos_ktx_KtxTexture_getDataSizeUncompressed` (matching the full package- and class name)

Currently, calling this function would result in a
`java.lang.UnsatisfiedLinkError: org.khronos.ktx.KtxTexture.getDataSizeUncompressed()J`

I also added a test that _might_ look useless at the first glance: It just creates a texture, and calls all the `native` "getter" functions on that (without any assertion). The point of that test is that it would have _prevented_ this kind of error...

### DIFF
--- a/interface/java_binding/src/test/java/org/khronos/ktx/test/KtxTexture2Test.java
+++ b/interface/java_binding/src/test/java/org/khronos/ktx/test/KtxTexture2Test.java
@@ -607,14 +607,24 @@ public class KtxTexture2Test {
                 .toAbsolutePath()
                 .normalize();
 
+        // The purpose of this test is to check the bindings for the 'native'
+        // functions that only return a value. When the binding for one of
+        // these functions is not implemented properly, then trying to call
+        // it will cause an 'UnsatisfiedLinkError'.
+        // This does not cover all 'native' functions: Some of them can only
+        // sensibly be called in the context of the other tests.
+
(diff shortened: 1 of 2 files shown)


interface/java_binding/src/test/java/org/khronos/ktx/test/KtxTexture2Test.java
```diff
@@ -599,4 +599,37 @@ public void testSupercompressionZLIB() throws IOException {
 
         t.destroy();
     }
+
+    @Test
```
MarkCallow, line 603: Please add a comment similar to what you wrote in the PR description about this test.

javagl: That certainly makes sense (otherwise, people might wonder what the purpose of the test is). 

(An aside: In another context, I automated this kind of test, [using some reflection magic](https://github.com/jcuda/jcuda/blob/master/JCudaJava/src/test/java/jcuda/test/BasicBindingTest.java) - but here, we only have ~20 `native` functions, whereas for JCuda, it was about (literally) thousands of `native` methods in dozens of classes...)

MarkCallow:
Thanks @javagl.

PR #958, opened 2024-11-13
Cast a macro argument to `int` to fix a C++20 error.

Cons-Cat:
Clang 20 errors for me on several expansions of the `KHR_DFDSETVAL` macro because it attempts to bitwise-and two different types of enums. For example:
```
/home/conscat/vulkan-experiments/KTX-Software/lib/basis_encode.cpp:256:5: error: invalid bitwise operation between different enumeration types ('_khr_df_vendorid_e' and '_khr_df_mask_e')
  256 |     KHR_DFDSETVAL(nbdb, VENDORID, KHR_DF_VENDORID_KHRONOS);
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/conscat/vulkan-experiments/KTX-Software/include/KHR/khr_df.h:109:13: note: expanded from macro 'KHR_DFDSETVAL'
  109 |     (((val) & (KHR_DF_MASK_ ## X)) << (KHR_DF_SHIFT_ ## X)))
      |       ~~~~~ ^ ~~~~~~~~~~~~~~~~~~~
```
This PR simply casts the value to an int inside of the macro.

### DIFF
--- a/include/KHR/khr_df.h
+++ b/include/KHR/khr_df.h
@@ -106,7 +106,7 @@ typedef enum _khr_df_mask_e {
     ((BDB)[KHR_DF_WORD_ ## X] = \
      ((BDB)[KHR_DF_WORD_ ## X] & \
       ~((KHR_DF_MASK_ ## X) << (KHR_DF_SHIFT_ ## X))) | \
-     (((val) & (KHR_DF_MASK_ ## X)) << (KHR_DF_SHIFT_ ## X)))
+     (((int)(val) & (KHR_DF_MASK_ ## X)) << (KHR_DF_SHIFT_ ## X)))
 
 /* Offsets relative to the start of a sample */
 typedef enum _khr_df_sampleword_e {


MarkCallow:
CI passed a few days ago using clang 20.0.0git (https:/github.com/llvm/llvm-project d6344c1cd0d099f8d99ee320f33fc9254dbe8288. Are you using a newer version than this, @Cons-Cat.

Cons-Cat:
My version was 2 weeks old.

Cons-Cat:
The issue was not due to my version of Clang, it was due to my `CMAKE_CXX_FLAGS` containing `std=gnu++26`. Apparently C++20 deprecated implicit enum conversions in this context. Arguably I shouldn't have had that flag there, but this is the only error I have that prevents building the library this way.

MarkCallow:
@Cons-Cat Thank you for clearing up the confusion of c++20 vs clang v20.

I am looking into the `reuse` failure. It clearly has nothing to do with the change you have made. I just tried the latest `reuse` locally and it passes so I don't have any idea of the cause at present. Please confirm your branch contains the file `REUSE.toml`.

MarkCallow:
I re-ran the `reuse` job and it passed. I have no idea what caused an apparently temporary issue before.

Issue #969, opened 2024-12-31
Java bindings: Invalid check for ASTC input swizzle parameter


javagl:
(Just some sort of "TODO" for me, to keep track of this: )

The [`KtxAstcParams#setInputSwizzle`](https://github.com/KhronosGroup/KTX-Software/blob/b6190046bbf71e0fbf9e7a964f6ee39d726b16be/interface/java_binding/src/main/java/org/khronos/ktx/KtxAstcParams.java#L245) function currently does a basic validity check for the given argument. This includes a check that the input only contains the characters `rgba01`. This breaks for the case where this value is initialized with its _default_ value, which is a `char[]` array that contains only `0` values (in contrast to `'0'` _characters_). 

For example:
```
KtxAstcParams a = new KtxAstcParams(); // All default
KtxAstcParams b = new KtxAstcParams(); // All default
b.setInputSwizzle(a.getInputSwizzle()); // This should work!
```
This will currently throw an `IllegalArgumentException`.  

I'll try to fix this soon, either as a mini-PR, or as part of a larger PR.

javagl:
Completely unrelated to the issue above, but as another ~"small thing that I'll probably fix soon": 

There currently is a [constant `KtxPackAstcBlockDimension.D6x4`](https://github.com/KhronosGroup/KTX-Software/blob/b6190046bbf71e0fbf9e7a964f6ee39d726b16be/interface/java_binding/src/main/java/org/khronos/ktx/KtxPackAstcBlockDimension.java#L32). I think that this should be called `D6x5`. I'll rename that, unless someone disagrees...

MarkCallow:
Please fix these soon. We'll be making a release in early February, hopefully.

javagl:
@MarkCallow OK, given that this is only a small fix, I created a PR at https://github.com/KhronosGroup/KTX-Software/pull/972 . 

(Otherwise, I would have added this to https://github.com/KhronosGroup/KTX-Software/pull/968 , but it's hard to give an exact timeline for that. It might be that in that process, further "small fixes" might be necessary, but we can address them when they come up).

[...]

PR #972, opened 2025-01-01
Java swizzle and constant fixes

javagl:
Fixes https://github.com/KhronosGroup/KTX-Software/issues/969 (see the issue for details).

The fix for the `inputSwizzle` (for both the `KtxAstcParams` and `KtxBasisParams`) is implemented by checking whether the input equals the default, which is a `char[] { 0, 0, 0, 0 }` array. If this is the case, then this value will be accepted, passed to the native layer, and handled there accordingly. Only when the input is _not_ that default, the elements will be checked to be in `rgba01`.


interface/java_binding/src/main/java/org/khronos/ktx/KtxAstcParams.java
```diff
@@ -235,21 +237,28 @@ public char[] getInputSwizzle() {
...
+	 * Otherwise, this swizzle must match the regular expression 
+	 * <code>/^[rgba01]{4}$/</code>.<br>
 	 * <br>
-	 * When the given swizzle is <code>null</code>, then no swizzling will be
-	 * applied to the input.
 	 *
 	 * @param inputSwizzle The swizzle
 	 */
 	public void setInputSwizzle(char[] inputSwizzle) {
+	    char defaultSwizzle[] = new char[4];
```
MarkCallow, line 250: Does Java initialize new arrays to all 0? Otherwise initialization is needed.

interface/java_binding/src/main/java/org/khronos/ktx/KtxBasisParams.java
```diff
@@ -410,29 +412,36 @@ public char[] getInputSwizzle() {
...
-			throw new IllegalArgumentException("The inputSwizzle must contain 4 characters");
-		}
-		String valid = "rgba01";
-		for (int i = 0; i < inputSwizzle.length; i++) {
-			char c = inputSwizzle[i];
-			if (valid.indexOf(c) == -1) {
-				throw new IllegalArgumentException("The inputSwizzle may only consist of 'rgba01', but contains " + c);
-			}
-		}
+	    char defaultSwizzle[] = new char[4];
```
MarkCallow, line 425: Same question as for the AstcParams case.

interface/java_binding/src/main/java/org/khronos/ktx/KtxAstcParams.java
javagl: **Yes**, all array components are initialized to be 0. 

https://docs.oracle.com/javase/specs/jls/se23/html/jls-4.html#jls-4.12.3 : 

>  Array components are unnamed variables that are created and initialized to default values 
> (link->) For type char, the default value is the null character, that is, `'\u0000'`.

interface/java_binding/src/main/java/org/khronos/ktx/KtxBasisParams.java
javagl: Same answer as above. 

With the addendum: I also didn't like the duplication, and considered to pull out this block into a function, so that the 'setter' could boil down to something like
`this.inputSwizzle = Swizzling.validate(inputSwizzle);`
This would also allow to avoid the duplication of the unit tests (which may be a stronger case to make). 

I somehow hestiate(d) to create such a `class Swizzling` just for this purpose, but will have another look: There might be other suplicated blocks that could justify moving this into some `class Utils` that does some of the parameter checks.

javagl: So I think that, in this case, it makes sense to move this to a common function. So I just did that, and there now is a `KtxUtilities.validateSwizzle` function. This avoids the duplication in the `BasisParams`/`AstcParams` themself, _and_ it avoids the duplication of the unit tests.

MarkCallow: > This avoids the duplication in the BasisParams/AstcParams themself, and it avoids the duplication of the unit tests.

Sounds good. I'll be away the next 3 days. I'll review again on my return.

PR #973, opened 2025-01-06
Fix crash on loading EXR 2.0 input

### LINKED ISSUE #966: Attempt to load EXR deep image crashes `ktx create`
An attempt to load an EXR deep image such as the test image at https://raw.githubusercontent.com/AcademySoftwareFoundation/openexr-images/main/v2/LeftView/Leaves.exr crashes `ktx create`. Debug config produces an error from the memory allocator "pointer being freed was not allocated" or possibly "attempting to free already freed pointer."

The root of the problem is that `LoadEXRImageFromMemory`, or more precisely `DecodeEXRImage` which it ends up calling, frees the exr_image it allocated via `FreeEXRImage` (see line 11350 in tinyexr.h) when `DecodeChunk` returns false due to finding invalid data in the EXR file.  LoadEXRImageFromMemory then returns an error which causes ExrInput::ReadImage to throw an exception. As part of the exception handling the ExrInput destructor is called and it too calls `FreeEXRImage`

The reason for the invalid data is because loading a deep image requires calling a different function, `LoadDeepEXR`. Whether an image is a "deep image" can be determined from the file version. If `version.non_image` is set it is a deep image.

Therefore 2 fixes are required here:

1. Do not call `FreeEXRImage` in `~ExrInput()` if the image has not been allocated or has been freed.
2. Check version for `non_image`and if set, exit with an error message to say such images are not supported.

MathiasMagnus:
Fixes  #966

### DIFF
--- a/tools/imageio/exr.imageio/exrinput.cc
+++ b/tools/imageio/exr.imageio/exrinput.cc
@@ -37,7 +37,9 @@ public:
     }
     ~ExrInput() {
         // FreeEXRVersion(&version); // No need to call, no such function
-        FreeEXRImage(&image);
+        if (image.width != 0 && image.height != 0) {
+            FreeEXRImage(&image);
+        }
         FreeEXRHeader(&header);
         FreeEXRErrorMessage(err);
     }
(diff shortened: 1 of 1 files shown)


MarkCallow:
Thank you @MathiasMagnus.

PR #977, opened 2025-01-27
Add normalise option to command create

wasimabbas-arm:
Fixes https://github.com/KhronosGroup/KTX-Software/issues/812 partially. At the moment only adds support to create. 

NOTE: I can't seem to work out how to add support to encode because its not using `Image` from `./tools/imageio/image.hpp` its using streams to load this.

I have also tried to add this normalize option to `./tools/ktx/encode_utils_common.h` first. You can see this in the first commit  https://github.com/KhronosGroup/KTX-Software/commit/53f728f9aa549a255d28aee3b5f0de747e351242. I am not sure if thats the way to go to see this option in both `encode` and `create` but it felt like its not an encoder (basisu, ASTC) specific option so feels misplaced. 

The other option is to add another "general_util.h" which stores other commonly used options amongst the tools because it doesn't fit in all of the following categories either.

`ktx create [OPTION...] <input-file...> <output-file>`
`Encode ASTC options:`
`Encode BasisLZ options:`
`Encode UASTC options:`
`Encode common options:`
`Generate Mipmap options:`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -303,6 +303,7 @@ if(MSVC)
     # Enable UTF-8 support
     add_compile_options( $<$<C_COMPILER_ID:MSVC>:/utf-8> )
     add_compile_options( $<$<CXX_COMPILER_ID:MSVC>:/utf-8> )
+    add_compile_options($<$<OR:$<COMPILE_LANGUAGE:C>,$<COMPILE_LANGUAGE:CXX>>:-Wno-cast-function-type-mismatch>)
 elseif(${CMAKE_CXX_COMPILER_ID} STREQUAL "GNU"
        OR ${CMAKE_CXX_COMPILER_ID} MATCHES "Clang")
     add_compile_options( -Wall -Wextra $<$<BOOL:${KTX_WERROR}>:-Werror>)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -307,7 +307,6 @@ elseif(${CMAKE_CXX_COMPILER_ID} STREQUAL "GNU"
        OR ${CMAKE_CXX_COMPILER_ID} MATCHES "Clang")
     add_compile_options( -Wall -Wextra $<$<BOOL:${KTX_WERROR}>:-Werror>)
     add_compile_options( $<IF:$<CONFIG:Debug>,-O0$<SEMICOLON>-g,-O3> )
-    add_compile_options($<$<OR:$<COMPILE_LANGUAGE:C>,$<COMPILE_LANGUAGE:CXX>>:-Wno-cast-function-type-mismatch>)
     if(EMSCRIPTEN)
         add_link_options( $<IF:$<CONFIG:Debug>,-gsource-map,-O3> )
     else()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -303,11 +303,11 @@ if(MSVC)
     # Enable UTF-8 support
     add_compile_options( $<$<C_COMPILER_ID:MSVC>:/utf-8> )
     add_compile_options( $<$<CXX_COMPILER_ID:MSVC>:/utf-8> )
-    add_compile_options($<$<OR:$<COMPILE_LANGUAGE:C>,$<COMPILE_LANGUAGE:CXX>>:-Wno-cast-function-type-mismatch>)
 elseif(${CMAKE_CXX_COMPILER_ID} STREQUAL "GNU"
        OR ${CMAKE_CXX_COMPILER_ID} MATCHES "Clang")
     add_compile_options( -Wall -Wextra $<$<BOOL:${KTX_WERROR}>:-Werror>)
     add_compile_options( $<IF:$<CONFIG:Debug>,-O0$<SEMICOLON>-g,-O3> )
+    add_compile_options($<$<OR:$<COMPILE_LANGUAGE:C>,$<COMPILE_LANGUAGE:CXX>>:-Wno-cast-function-type-mismatch>)
     if(EMSCRIPTEN)
         add_link_options( $<IF:$<CONFIG:Debug>,-gsource-map,-O3> )
     else()
--- a/tools/ktx/command_create.cpp
+++ b/tools/ktx/command_create.cpp
@@ -1616,10 +1616,8 @@ void CommandCreate::executeCreate() {
                 if (target.format().transfer() != KHR_DF_TRANSFER_UNSPECIFIED && target.format().transfer() != KHR_DF_TRANSFER_LINEAR) {
                     fatal(rc::INVALID_FILE,
                         "Input file \"{}\" The transfer function to be applied to the created texture is neither linear nor none. Normalize is only available for linear images. "
-                        "Use or modify {} or {} settings to assign the transfer function or convert the input image to linear, if required.",
-                        OptionsCreate::kAssignOetf, OptionsCreate::kConvertOetf);
-
-                        fmtInFile(inputFilepath));
+                        "Use or modify \"{}\" or \"{}\" settings to assign the transfer function or convert the input image to linear, if required.",
+                        fmtInFile(inputFilepath), OptionsCreate::kAssignOetf, OptionsCreate::kConvertOetf);
                     }
                 image->normalize();
             }
--- a/tools/ktx/command_create.cpp
+++ b/tools/ktx/command_create.cpp
@@ -1591,6 +1591,9 @@ void CommandCreate::executeCreate() {
                 image->yflip();
             }
 
+            if (options.normalize)
+                image->normalize();
+
             if (options.swizzleInput)
                 image->swizzle(*options.swizzleInput);
 
@@ -2313,6 +2316,9 @@ void CommandCreate::generateMipLevels(KTXTexture2& texture, std::unique_ptr<Imag
         } catch (const std::exception& e) {
             fatal(rc::RUNTIME_ERROR, "Mipmap generation failed: {}", e.what());
         }
+
+        if (options.normalize)
+            image->normalize();
 
         const auto imageData = convert(image, options.vkFormat, inputFile);
 
--- a/tools/ktx/command_create.cpp
+++ b/tools/ktx/command_create.cpp
@@ -1614,10 +1614,21 @@ void CommandCreate::executeCreate() {
 
             if (options.normalize) {
                 if (target.format().transfer() != KHR_DF_TRANSFER_UNSPECIFIED && target.format().transfer() != KHR_DF_TRANSFER_LINEAR) {
-                    fatal(rc::INVALID_FILE,
-                        "Input file \"{}\" The transfer function to be applied to the created texture is neither linear nor none. Normalize is only available for linear images. "
-                        "Use or modify \"{}\" or \"{}\" settings to assign the transfer function or convert the input image to linear, if required.",
-                        fmtInFile(inputFilepath), OptionsCreate::kAssignOetf, OptionsCreate::kConvertOetf);
+                    const auto input_error_message = "Input file \"{}\" The transfer function to be applied to the created texture is neither linear nor none. Normalize is only available for these transfer functions.";
+                    const auto assign_error_message = "Input file \"{}\" Use \"{}\" to assign the linear transfer function to the input image, if required.";
+                    const auto convert_error_message = "Input file \"{}\" Modify \"{}\" settings to convert the input image to linear transfer function, if required.";
+                    const auto inputTransfer =  inputImageFile->spec().format().transfer();
+                    bool is_file_error = (inputTransfer != KHR_DF_TRANSFER_UNSPECIFIED && inputTransfer != KHR_DF_TRANSFER_LINEAR);
+                    bool is_assign_error =  !options.assignOETF.has_value();
+                    bool is_convert_error =  !options.convertOETF.has_value();
+                    if (is_assign_error)
+                        fatal(rc::INVALID_FILE, assign_error_message, fmtInFile(inputFilepath), OptionsCreate::kAssignOetf);
+                    else if (is_convert_error)
+                        fatal(rc::INVALID_FILE, convert_error_message, fmtInFile(inputFilepath), OptionsCreate::kConvertOetf);
+                    else {
+                        assert(is_file_error && "In this branch it must be the input file that has the transfre function issue"); (void)is_file_error;
+                        fatal(rc::INVALID_FILE, input_error_message, fmtInFile(inputFilepath));
+                    }
                     }
                 image->normalize();
             }
--- a/tools/ktx/command_create.cpp
+++ b/tools/ktx/command_create.cpp
@@ -148,11 +148,11 @@ struct OptionsCreate {
                 (kEncode, "Encode the created KTX file. Case insensitive."
                     "\nPossible options are: basis-lz | uastc", cxxopts::value<std::string>(), "<codec>")
                 (kNormalize, "Normalize input normals to have a unit length. Only valid for\n"
-                    "linear normal textures with 2 or more components. For 2-component inputs\n"
-                    "2D unit normals are calculated. Do not use these 2D unit normals\n"
-                    "to generate X+Y normals with --normal-mode. For 4-component inputs\n"
-                    "a 3D unit normal is calculated. 1.0 is used for the value of the\n"
-                    "4th component. Cannot be used with --raw.\n")
+                    "linear normal textures with 2 or more components. For 2-component\n"
+                    "inputs 2D unit normals are calculated. Do not use these 2D unit\n"
+                    "normals to generate X+Y normals with --normal-mode. For 4-component\n"
+                    "inputs a 3D unit normal is calculated. 1.0 is used for the value of\n"
+                    "the 4th component. Cannot be used with --raw.")
                 (kSwizzle, "KTX swizzle metadata.", cxxopts::value<std::string>(), "[rgba01]{4}")
                 (kInputSwizzle, "Pre-swizzle input channels.", cxxopts::value<std::string>(), "[rgba01]{4}")
                 (kAssignTf, "Force the created texture to have the specified transfer function, ignoring"
@@ -468,7 +468,7 @@ struct OptionsCreate {
 
         if (args[kNormalize].count()) {
             if (raw)
-                report.fatal_usage("Option --normalize can't be used with --raw.");
+                report.fatal_usage("Conflicting options: Option --normalize can't be used with --raw.");
             normalize = true;
         }
 
@@ -908,11 +908,11 @@ Create a KTX2 file from various input files.
         </dd>
         <dt>\--normalize</dt>
         <dd>Normalize input normals to have a unit length. Only valid for
-                linear normal textures with 2 or more components. For 2-component inputs
-                2D unit normals are calculated. Do not use these 2D unit normals
-                to generate X+Y normals with @b --normal-mode. For 4-component inputs
-                a 3D unit normal is calculated. 1.0 is used for the value of the
-                4th component. Cannot be used with @b \--raw.</dd>
+                linear normal textures with 2 or more components. For 2-component
+                inputs 2D unit normals are calculated. Do not use these 2D unit
+                normals to generate X+Y normals with @b --normal-mode. For 4-component
+                inputs a 3D unit normal is calculated. 1.0 is used for the value of
+                the 4th component. Cannot be used with @b \--raw.</dd>
         <dt>\--swizzle [rgba01]{4}</dt>
         <dd>KTX swizzle metadata.</dd>
         <dt>\--input-swizzle [rgba01]{4}</dt>
@@ -1613,10 +1613,10 @@ void CommandCreate::executeCreate() {
             }
 
             if (options.normalize) {
-                if (target.format().transfer() != KHR_DF_TRANSFER_LINEAR) {
+                if (colorSpaceInfo.usedInputTransferFunction != KHR_DF_TRANSFER_LINEAR) {
                     fatal(rc::INVALID_FILE,
-                        "Input file \"{}\" transfer functions is not linear. "
-                        "Use --assign-oetf=linear or --convert-oetf=linear to avoid this error.",
+                        "Input file \"{}\" transfer function is not linear. Normalize is only available for linear images. "
+                        "Use --assign-oetf=linear or --convert-oetf=linear to convert to linear if required.",
                         fmtInFile(inputFilepath));
                     }
                 image->normalize();
--- a/tools/ktx/command_create.cpp
+++ b/tools/ktx/command_create.cpp
@@ -152,7 +152,7 @@ struct OptionsCreate {
                     "2D unit normals are calculated. Do not use these 2D unit normals\n"
                     "to generate X+Y normals with --normal-mode. For 4-component inputs\n"
                     "a 3D unit normal is calculated. 1.0 is used for the value of the\n"
-                    "4th component.\n")
+                    "4th component. Cannot be used with --raw.\n")
                 (kSwizzle, "KTX swizzle metadata.", cxxopts::value<std::string>(), "[rgba01]{4}")
                 (kInputSwizzle, "Pre-swizzle input channels.", cxxopts::value<std::string>(), "[rgba01]{4}")
                 (kAssignTf, "Force the created texture to have the specified transfer function, ignoring"
@@ -466,8 +466,11 @@ struct OptionsCreate {
                 mipmapWrap = it->second;
         }
 
-        if (args[kNormalize].count())
+        if (args[kNormalize].count()) {
+            if (raw)
+                report.fatal_usage("Option --normalize can't be used with --raw.");
             normalize = true;
+        }
 
         if (args[kSwizzle].count()) {
             swizzle = to_lower_copy(args[kSwizzle].as<std::string>());
@@ -909,7 +912,7 @@ Create a KTX2 file from various input files.
                 2D unit normals are calculated. Do not use these 2D unit normals
                 to generate X+Y normals with @b --normal-mode. For 4-component inputs
                 a 3D unit normal is calculated. 1.0 is used for the value of the
-                4th component.</dd>
+                4th component. Cannot be used with @b \--raw.</dd>
         <dt>\--swizzle [rgba01]{4}</dt>
         <dd>KTX swizzle metadata.</dd>
         <dt>\--input-swizzle [rgba01]{4}</dt>
@@ -1610,7 +1613,7 @@ void CommandCreate::executeCreate() {
             }
 
             if (options.normalize) {
-                if (colorSpaceInfo.usedInputTransferFunction != KHR_DF_TRANSFER_LINEAR) {
+                if (target.format().transfer() != KHR_DF_TRANSFER_LINEAR) {
                     fatal(rc::INVALID_FILE,
                         "Input file \"{}\" transfer functions is not linear. "
                         "Use --assign-oetf=linear or --convert-oetf=linear to avoid this error.",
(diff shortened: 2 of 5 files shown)


MarkCallow:
I think keeping encode focused on encoding and not adding image manipulations is the right choice. I.e normaiize only in `ktx create` is fine. I don't think `ktxsc` supported normalize either. It would have the same issue as it too doesn't use `Image`.

I can see at some point adding a function to make an `Image` from the data pointer in a `ktxTexture2` texture object as the starting point to being able to manipulate the images in a KTX file. That is future work.

MarkCallow:
> Update create docs but feels like this has gone out of sync with the help text

The help output is a simplified version of the man page. They do need to be kept broadly in sync.

Any time you change the help you have to update the golden file in the CTS otherwise the `crete.help` test will fail, as is happening here. This is separate from keeping help and man page in sync.

tools/ktx/command_create.cpp
```diff
@@ -141,6 +143,12 @@ struct OptionsCreate {
                     " This option is mutually exclusive with --runtime-mipmap and cannot be used with UINT or 3D textures.")
                 (kEncode, "Encode the created KTX file. Case insensitive."
                     "\nPossible options are: basis-lz | uastc", cxxopts::value<std::string>(), "<codec>")
+                (kNormalize, "Normalize input normals to have a unit length. Only valid for\n"
+                    "linear normal textures with 2 or more components. For 2-component inputs\n"
+                    "2D unit normals are calculated. Do not use these 2D unit normals\n"
+                    "to generate X+Y normals for --normal-mode. For 4-component inputs\n"
```
MarkCallow, line 149: I think "with --normal-mode" is more accurate. Isn't it normal-mode that makes the X+Y normals?

tools/ktx/command_create.cpp
```diff
@@ -783,6 +794,13 @@ Create a KTX2 file from various input files.
                 wrap | reflect | clamp.
                 Defaults to clamp.</dd>
         </dl>
+        <dt>\--normalize</dt>
+        <dd>Normalize input normals to have a unit length. Only valid for
+                linear normal textures with 2 or more components. For 2-component inputs
+                2D unit normals are calculated. Do not use these 2D unit normals
+                to generate X+Y normals for --normal-mode. For 4-component inputs
```
MarkCallow, line 801: "with @b \--normal-mode."

Even if you don't agree with changing "for" the other changes are needed.

tools/ktx/command_create.cpp
```diff
@@ -1329,6 +1347,9 @@ void CommandCreate::executeCreate() {
                 image->yflip();
             }
 
+            if (options.normalize)
+                image->normalize();
```
MarkCallow, line 1351: I think we need a fatal error here if the oetf is not linear and a subcase to test it added to `tests/create/cli_errors.json`.

MarkCallow (changes requested):
Thanks @wasimabbas-arm.

tools/ktx/command_create.cpp
wasimabbas-arm: Fixed in https://github.com/KhronosGroup/KTX-Software/pull/977/commits/f1b94c8fd84923b76dcee2c080d9b5735827fb37

tools/ktx/command_create.cpp
wasimabbas-arm: Fixed in https://github.com/KhronosGroup/KTX-Software/pull/977/commits/f1b94c8fd84923b76dcee2c080d9b5735827fb37

tools/ktx/command_create.cpp
wasimabbas-arm: Had a go at adding the fatal error in https://github.com/KhronosGroup/KTX-Software/pull/977/commits/f1b94c8fd84923b76dcee2c080d9b5735827fb37 but I am strugling to understand what's the right way to read the OETF. No one calls image->setOetf() so by that time its unspecified. There are `assign_oetf` and `convert_oetf` which are supposed to be calling `image->transformColorSpace()` at some point which _should_ change the `colorSpaceInfo.usedInputTransferFunction` an/or image::oetf but it doesn't. 

Let's get this right before I add more tests.

MarkCallow: Use `target.format().transfer()`.

tools/ktx/command_create.cpp
```diff
@@ -783,6 +794,13 @@ Create a KTX2 file from various input files.
                 wrap | reflect | clamp.
                 Defaults to clamp.</dd>
         </dl>
+        <dt>\--normalize</dt>
+        <dd>Normalize input normals to have a unit length. Only valid for
+                linear normal textures with 2 or more components. For 2-component inputs
+                2D unit normals are calculated. Do not use these 2D unit normals
+                to generate X+Y normals with @b --normal-mode. For 4-component inputs
+                a 3D unit normal is calculated. 1.0 is used for the value of the
+                4th component.</dd>
```
MarkCallow, line 803: Add `Cannot be used with @b \--raw.` and make sure you test for that in OptionsCreate::process.

tools/ktx/command_create.cpp
wasimabbas-arm: this `target.formate().transfer()` does work because these are set in `determineTargetColorSpace` only if `convertOETF` or `assignOETF` are set. My origional check works so I left it.

tools/ktx/command_create.cpp
wasimabbas-arm: Fixed in https://github.com/KhronosGroup/KTX-Software/pull/977/commits/29ed55299d6a4328daa1ad3a01e72c5fe963d9b6

wasimabbas-arm:
> Error: Unable to create a container for the artifact pyktx-4.3.2-cp311-cp311-win_amd64.whl at https://pipelinesghubeus24.actions.githubusercontent.com/P1YfxNQds3vsiK5yfr414BxgGxrn3zpqF8zRT7xD5N2qBPqBZC/_apis/pipelines/workflows/13117932387/artifacts?api-version=6.0-preview

This doesn't look like a related build issue, is it?

tools/ktx/command_create.cpp
MarkCallow: The test you have does not work in some cases because usedInputTransferFunction is the transfer function of the input file prior to application of `--convert-oetf`. `target.format().transfer` is set initially to the TF that matches the format given with`--format` and is then modified by `--assign-oetf` or `--convert-oetf`, if used.

Probably should look into calling image->setOetf(). `image->transformColorSpace` is not intended to change `colorSpaceInfo.usedInputTransferFunction`. It's meaning is as I just described.

MarkCallow:
> This doesn't look like a related build issue, is it?

No it doesn't. It looks like a change to some part of the GitHub helpers.

tools/ktx/command_create.cpp
wasimabbas-arm: Taking a step back. 

1. What does linear OETF really mean? Does it mean the original OETF of the input data needs to be in Linear (in which case `colorSpaceInfo.usedInputTransferFunction` should be enough.

2. Does it mean whatever the OETF ends up after `--assign-oetf` and/or `--convert-oetf` are applied?. 
    2.1. In assign case the user just says, I am telling you to use that `oetf` and we should respect that, providing correct behaviour accordingly. 
    2.2 In the convert case where does the actual conversion happens? Is this a conversion of `colorSpaceInfo` meta data only or the actual image data (pixels) are also modified in some way(I will find out)? I assume in this case the normalize/linear check should happen after this?

Lets agree on this, then I can find the right combination of checks we need to make.

MarkCallow: > Taking a step back.
> 
>     1. What does linear OETF really mean? Does it mean the original OETF of the input data needs to be in Linear (in which case `colorSpaceInfo.usedInputTransferFunction` should be enough.

It means there is no power function encoding. It does not mean the original OETF of the input data needs to be in linear. `colorSpaceInfo.usedInputTransferFunction` is the transfer function gleaned from the file's metadata or that specified to `--assign-oetf`.

> 
>     2. Does it mean whatever the OETF ends up after `--assign-oetf` and/or `--convert-oetf` are applied?.
>        2.1. In assign case the user just says, I am telling you to use that `oetf` and we should respect that, providing correct behaviour accordingly.

The value from `--assign-oetf` is assigned to both `colorSpaceInfo.usedInputTransferFunction` and `target.format().transfer()`. The latter needs to be valid for the specified `--format` others a fatal error will ensue. `colorSpaceInfo.usedInputTransferFunction` is used as the input to color space conversion. The output is the value specified with `--convert-oetf`. If they are the same there is no conversion.

An implicit conversion is done if neither `--assign-oetf` nor `--convert-oetf` are specified and `colorSpaceInfo.usedInputTransferFunction` does not match what is needed for the target VkFormat.

>        2.2 In the convert case where does the actual conversion happens? Is this a conversion of `colorSpaceInfo` meta data only or the actual image data (pixels) are also modified in some way(I will find out)? I assume in this case the normalize/linear check should happen after this?

The conversion happens here https://github.com/KhronosGroup/KTX-Software/blob/91e61d6fa3f06eb8616fc2eba74fba7703e4fe11/tools/ktx/command_create.cpp#L1307
The image pixels are modified.

The source and destination transfer functions are set in https://github.com/KhronosGroup/KTX-Software/blob/91e61d6fa3f06eb8616fc2eba74fba7703e4fe11/tools/ktx/command_create.cpp#L2123

Testing against `target.format().transfer()` anywhere after `determineTargetColorSpace` has been called will work for the purposes of the error check. Calling normalize must be called after transformColorSpace has been called. It is.

wasimabbas-arm: Ok. Its fixed in https://github.com/KhronosGroup/KTX-Software/pull/977/commits/31efeb39a5ab6eaa74ce0bf4ba2fee6d192223f7. I am now using `target.format().transfer()` after `determineTargetColorSpace` is called and normalize was called after `transformColorSpace` anyways.

MarkCallow: You may be wondering why I asked you to use target.format().transfer() when a check against vkFormat would work. It is because I am about to land a change to `ktx create` which allows the user to `--assign-tf` (new name) any transfer function to a  UNORM texture, except for sRGB if the VkFormat has an SRGB equivalent. This has been permitted in the KTX spec. for some time now. The software is catching up.

MarkCallow:
> > This doesn't look like a related build issue, is it?
> 
> No it doesn't. It looks like a change to some part of the GitHub helpers.

I reran the failed jobs but the same thing happened. I need to investigate. It's a piece of github functionality that is failing and, very typically of software generally, the error message gives no clue why the action is "unable to create a container."

tools/ktx/command_create.cpp
```diff
@@ -1329,6 +1350,16 @@ void CommandCreate::executeCreate() {
                 image->yflip();
             }
 
+            if (options.normalize) {
+                if (target.format().transfer() != KHR_DF_TRANSFER_LINEAR) {
```
MarkCallow, line 1354: For future proofing

```
if (target.format().transfer() != KHR_DF_TRANSFER_UNSPECIFIED && target.format().transfer() != KHR_DF_TRANSFER_LINEAR) {
```

tools/ktx/command_create.cpp
```diff
@@ -1329,6 +1350,16 @@ void CommandCreate::executeCreate() {
...
+                    fatal(rc::INVALID_FILE,
+                        "Input file \"{}\" transfer function is not linear. Normalize is only available for linear images. "
```
MarkCallow, line 1356: "The transfer function to be applied to the created texture is neither linear nor none. Normalize ..."

tools/ktx/command_create.cpp
```diff
@@ -1329,6 +1350,16 @@ void CommandCreate::executeCreate() {
...
+                        "Use --assign-oetf=linear or --convert-oetf=linear to convert to linear if required.",
```
MarkCallow, line 1357:
```
"Use or modify {} or {} settings to assign the transfer function or convert the input image to linear, if required.",
OptionsCreate::kAssignOetf, OptionsCreate::kConvertOetf);
```

I'm not entirely happy with this wording.

Regarding this and the comment above, if you want to get really ambitious you can check `firstInputFile.format().transfer()` to see it is the file or the options and modify the error message accordingly.

MarkCallow (changes requested):
Please make the suggested changes to the error handling and message.

MarkCallow:
> Error: Unable to create a container for the artifact pyktx-4.3.2-cp311-cp311-win_amd64.whl

Please rebase to main to get the fix for this.

MarkCallow:
I am about to land PR #983 which updates the ASTC encoder to 5.2.0. After it lands you will likely have to regenerate the golden files in your tests.

Please finish this PR. I have almost completed the other items I want for the next release of this software and don't want to be held up by this.

tools/ktx/command_create.cpp
wasimabbas-arm: Fixed in https://github.com/KhronosGroup/KTX-Software/pull/977/commits/61556ccf343abe68c01e8497f00c9ca2f786170c

tools/ktx/command_create.cpp
wasimabbas-arm: Fixed in https://github.com/KhronosGroup/KTX-Software/pull/977/commits/61556ccf343abe68c01e8497f00c9ca2f786170c

tools/ktx/command_create.cpp
```diff
@@ -1351,10 +1351,12 @@ void CommandCreate::executeCreate() {
             }
 
             if (options.normalize) {
-                if (target.format().transfer() != KHR_DF_TRANSFER_LINEAR) {
+                if (target.format().transfer() != KHR_DF_TRANSFER_UNSPECIFIED && target.format().transfer() != KHR_DF_TRANSFER_LINEAR) {
                     fatal(rc::INVALID_FILE,
-                        "Input file \"{}\" transfer function is not linear. Normalize is only available for linear images. "
-                        "Use --assign-oetf=linear or --convert-oetf=linear to convert to linear if required.",
+                        "Input file \"{}\" The transfer function to be applied to the created texture is neither linear nor none. Normalize is only available for linear images. "
```
MarkCallow, line 1356: I find it a little confusing saying just "linear" in the second sentence. Suggest changing to "... only available for these transfer functions."

wasimabbas-arm: I have rewored this area with more detailed error message https://github.com/KhronosGroup/KTX-Software/pull/977/commits/ae46489470217b430e1f957aa3c4a5401d90fa69

tools/ktx/command_create.cpp
wasimabbas-arm: I have rewored this area with more detailed error message https://github.com/KhronosGroup/KTX-Software/pull/977/commits/ae46489470217b430e1f957aa3c4a5401d90fa69

wasimabbas-arm:
@MarkCallow I am not sure what's happened to this PR.

Current status is that the following tests are failing:

```
The following tests FAILED:
	160 - ktx2check-test.all (Failed)
	205 - ktxsc-test.zcmp-cubemap (Timeout)
	533 - ktxToolsTest.extract.png_2d (Failed)
	581 - ktxToolsTest.create.basic_png (Failed)
```

All except cubemap ones are giving some `clitest.py:20: SyntaxWarning: invalid escape sequence '\$'` The cubemap one is timeout(ing).

And I have merged this with master, locally I don't have any conflicts. But Why is it not running the CI build? Its confusing to me why the merge conflict (if there is one really) is blocking a CI build (So I can't see what is actually failing if anything related)

tools/ktx/command_create.cpp
```diff
@@ -1329,6 +1350,27 @@ void CommandCreate::executeCreate() {
...
+                    const auto inputTransfer =  inputImageFile->spec().format().transfer();
+                    bool is_file_error = (inputTransfer != KHR_DF_TRANSFER_UNSPECIFIED && inputTransfer != KHR_DF_TRANSFER_LINEAR);
+                    bool is_assign_error =  !options.assignOETF.has_value();
+                    bool is_convert_error =  !options.convertOETF.has_value();
+                    if (is_assign_error)
+                        fatal(rc::INVALID_FILE, assign_error_message, fmtInFile(inputFilepath), OptionsCreate::kAssignOetf);
+                    else if (is_convert_error)
+                        fatal(rc::INVALID_FILE, convert_error_message, fmtInFile(inputFilepath), OptionsCreate::kConvertOetf);
+                    else {
+                        assert(is_file_error && "In this branch it must be the input file that has the transfre function issue"); (void)is_file_error;
```
MarkCallow, line 1367: s/transfre/transfer/

tools/ktx/command_create.cpp
```diff
@@ -1329,6 +1350,27 @@ void CommandCreate::executeCreate() {
...
             }
 
+            if (options.normalize) {
+                if (target.format().transfer() != KHR_DF_TRANSFER_UNSPECIFIED && target.format().transfer() != KHR_DF_TRANSFER_LINEAR) {
+                    const auto input_error_message = "Input file \"{}\" The transfer function to be applied to the created texture is neither linear nor none. Normalize is only available for these transfer functions.";
+                    const auto assign_error_message = "Input file \"{}\" Use \"{}\" to assign the linear transfer function to the input image, if required.";
+                    const auto convert_error_message = "Input file \"{}\" Modify \"{}\" settings to convert the input image to linear transfer function, if required.";
+                    const auto inputTransfer =  inputImageFile->spec().format().transfer();
+                    bool is_file_error = (inputTransfer != KHR_DF_TRANSFER_UNSPECIFIED && inputTransfer != KHR_DF_TRANSFER_LINEAR);
+                    bool is_assign_error =  !options.assignOETF.has_value();
```
MarkCallow, line 1360: I am surprised this is building. There is no longer an `assignOETF` or `convertOETF` in `command_create.cpp` after the merge of PR #982. I think something must have gone strangely in your rebase or merge.

MarkCallow:
> And I have merged this with master, locally I don't have any conflicts. But Why is it not running the CI build? Its confusing to me why the merge conflict (if there is one really) is blocking a CI build (So I can't see what is actually failing if anything related). Its as if the CI build is only checking for CLA requirements now.

I think the CI build is not running because GH cannot merge the PR with main to create the tree it uses for the builds. I don't understand this well so I might be completely wrong.

The first step is that you need to rebase the CTS tests, force push the result to the CTS branch you are using then update the CTS reference (`git commit tests/cts`) in KTX-Software. (That conflict means both the update to KTX-Software that you are merging and your branch have modified the CTS reference.)

I don't know what the conflicts are in `command_create.cpp`. It tells me they are too complex to resolve in the web editor so it won't open it.

I am sorry for the hassle. I needed to merge the TF updates.

tools/ktx/command_create.cpp
wasimabbas-arm: Ok. yes. something did go wrong with the rebase. Now I have done it again and after changing to `assignTF` and `convertTF` 32 tests fail and I am getting error while running CTS as well. Didn't notice this before.

```
The following tests FAILED:
	160 - ktx2check-test.all (Failed)
	372 - ktxToolsTest.info.invalid_file (Failed)
	436 - ktxToolsTest.validate.error_6012_BasicInvalidTransferFunction (Failed)
	437 - ktxToolsTest.validate.error_6013_BasicSRGBMismatch (Failed)
	438 - ktxToolsTest.validate.error_6014_BasicNotSRGBMismatch (Failed)
	453 - ktxToolsTest.validate.warning_6029_TooManySamples (Failed)
	580 - ktxToolsTest.create.cli_errors (Failed)
	586 - ktxToolsTest.create.colorspaces_assign_primaries_exr (Failed)
	587 - ktxToolsTest.create.colorspaces_assign_primaries_png (Failed)
	588 - ktxToolsTest.create.colorspaces_assign_primaries_raw (Failed)
	589 - ktxToolsTest.create.colorspaces_combinations (Failed)
	592 - ktxToolsTest.create.colorspaces_convert_primaries_exr (Failed)
	593 - ktxToolsTest.create.colorspaces_convert_primaries_png (Failed)
	595 - ktxToolsTest.create.colorspaces_error_png_primaries_mismatch (Failed)
	596 - ktxToolsTest.create.colorspaces_error_png_transfer_mismatch (Failed)
	597 - ktxToolsTest.create.colorspaces_error_png_unknown_icc (Failed)
	598 - ktxToolsTest.create.colorspaces_error_linear_oetf (Failed)
	599 - ktxToolsTest.create.colorspaces_error_srgb_oetf (Failed)
	600 - ktxToolsTest.create.colorspaces_error_unknown_icc (Failed)
	601 - ktxToolsTest.create.colorspaces_error_assign_primaries_none_convert_primaries (Failed)
	603 - ktxToolsTest.create.colorspaces_warning_png_primaries_mismatch (Failed)
	604 - ktxToolsTest.create.colorspaces_warning_png_transfer_mismatch (Failed)
	620 - ktxToolsTest.create.help (Failed)
	651 - ktxToolsTest.create.warn_fail_texcoord_origin (Failed)
	668 - ktxToolsTest.encode.help (Failed)
	686 - ktxToolsTest.compare.invalid_file1 (Failed)
	687 - ktxToolsTest.compare.invalid_file2 (Failed)
	688 - ktxToolsTest.compare.invalid_file_both (Failed)
	690 - ktxToolsTest.compare.allow_invalid_input1 (Failed)
	691 - ktxToolsTest.compare.allow_invalid_input2 (Failed)
	692 - ktxToolsTest.compare.allow_invalid_input_both (Failed)
	708 - ktxToolsTest.compare.ignore_dfd_all_except_color_space (Failed)
Errors while running CTest
```

MarkCallow: > The CI builds works now

It is running but it is not building successfully.

wasimabbas-arm: when you say 
> not building
You mean there are build errors?
I see no build errors in any runs except https://github.com/KhronosGroup/KTX-Software/actions/runs/13544009481/job/37851279366?pr=977 where some weird issue here where install NSIS fails.

Every other run builds fins. The above tests are failing in those.

MarkCallow: >   The CI builds works now
>
> It is running but it is not building successfully.

I didn't see your message from 13 hours ago before I sent the above. GitHub does not update PR conversations on an open web page. I have to manually update the page to see additions.

As for the build errors, I think the merge/rebase of the CTS is not correct. For example, I see many  uses of `input/validate/6012/error_BasicInvalidTransferFunction_*.ktx2` files. However all but 1 of those files (`*_0xFF.ktx2`) were removed in when https://github.com/KhronosGroup/KTX-Software-CTS/pull/38 was merged. One new file (`*_0x14.ktx2`) was added. The other files are no longer invalid, hence the tests fail.

The golden files for the `error_BasicSRGBMismatch*` files were all updated in https://github.com/KhronosGroup/KTX-Software-CTS/pull/38. I suspect you don't have those updates either.

Did you rebase your changes after https://github.com/KhronosGroup/KTX-Software-CTS/pull/38 was merged to main?

MarkCallow: > You mean there are build errors?

No. Sorry. I meant that the CI runs are no completing successfully.

wasimabbas-arm: Thanks Mark. Ok I think it must be that the main repo is rebased but the CTS one isn't and I have the wrong submodule link.

wasimabbas-arm: Ok. I think (fingers crossed) I have done it right this time. (TIL you have to rebase each submodule separately if you are doing a main repo rebase if there were changes there too).

Whats up with https://github.com/KhronosGroup/KTX-Software/actions/runs/13562803879/job/37909217848?pr=977 ? How come you didn't see this when you were working on https://github.com/KhronosGroup/KTX-Software/pull/982 and updating astcenc (although it seems like its not related to astcenc, there is similar error in the gl_funcs.c as well)

MarkCallow: The Windows runner was updated on 02.24. Those previous builds use an older runner. No doubt the compiler was updated as part of 02.24 and the new one is now throwing these warnings about incompatible functions.

Please resolve them, if you can. I am in the  middle of a massive update to the CTS because of the change in spec. revision 4 to have sized DFDs in supercompressed files. I do not want to switch attention to something else until it is done.

wasimabbas-arm: Ok that explains it thanks. I will see what I can do.

tools/ktx/command_create.cpp
wasimabbas-arm: Fixed in one of the commits.

CMakeLists.txt
```diff
@@ -303,6 +303,7 @@ if(MSVC)
     # Enable UTF-8 support
     add_compile_options( $<$<C_COMPILER_ID:MSVC>:/utf-8> )
     add_compile_options( $<$<CXX_COMPILER_ID:MSVC>:/utf-8> )
+    add_compile_options($<$<OR:$<COMPILE_LANGUAGE:C>,$<COMPILE_LANGUAGE:CXX>>:-Wno-cast-function-type-mismatch>)
```
MarkCallow: I'd prefer to use the proper casts (whatever that is) rather than disable the warning.

Give that warnings were previously disabled around this code for MSVC how come we started getting them with the latest compiler?

MarkCallow: > Give that warnings were previously disabled around this code for MSVC how come we started getting them with the latest compiler?

Oh! Were the failures in the MingW build using GCC?

CMakeLists.txt
```diff
@@ -303,11 +303,11 @@ if(MSVC)
     # Enable UTF-8 support
     add_compile_options( $<$<C_COMPILER_ID:MSVC>:/utf-8> )
     add_compile_options( $<$<CXX_COMPILER_ID:MSVC>:/utf-8> )
-    add_compile_options($<$<OR:$<COMPILE_LANGUAGE:C>,$<COMPILE_LANGUAGE:CXX>>:-Wno-cast-function-type-mismatch>)
 elseif(${CMAKE_CXX_COMPILER_ID} STREQUAL "GNU"
        OR ${CMAKE_CXX_COMPILER_ID} MATCHES "Clang")
     add_compile_options( -Wall -Wextra $<$<BOOL:${KTX_WERROR}>:-Werror>)
     add_compile_options( $<IF:$<CONFIG:Debug>,-O0$<SEMICOLON>-g,-O3> )
+    add_compile_options($<$<OR:$<COMPILE_LANGUAGE:C>,$<COMPILE_LANGUAGE:CXX>>:-Wno-cast-function-type-mismatch>)
```
MarkCallow, line 310: This has broken the clang builds on macOS/Travis because this elseif is for both GCC and Clang and Clang does not know this warning type.

I've successfully built everything yesterday with GCC 14 (on macOS) without any of the changes in these last commits. I'd better see what version is in use in MingW on the latest Windows runner.

MarkCallow: I need to pay more attention to the logs before writing ...

The build that originally failed used clangcl which in the latest Windows runner is based on Clang 19. It looks like this warning has been introduced in Clang 19. The Emscripten docker image, which is usually the canary for Clang issues is still at Clang 18 and that build was working before you started trying to fix the CLangCL problem. MacOS Xcode is still on Clang 16.

It is best to find the proper cast to use to avoid having to worry about which compilers' warnings to disable.

abbaswasim: I have tried a few things yesterday. Since I don’t have a windows machine a bit painful but I had to commit things to CI to test. 

I thought it was windows so added check for VS. I tried adding a specific warning number based disable but I couldn’t find a number for that warning. As a test I disabled all warnings but still didn’t help and I think also got the syntax wrong in one commit but then I realized, as you also noticed, that it’s actually clangcl 19 not VS this is why these fixes weren’t working and why I couldn’t find the warning number. Before doing this I actually added a reinterpet_cast which I learned, is the right way to disable this warning.

Anyways when I saw it was clang I I removed the reinterpret_cast and added this compiler generator in cmake, I moved this around for VS to clang + Gnu side as you can see in the commits. 

But then I saw the mingw error and some clang 14 error iirc this is where I was last time when I got dragged into other things. And the reason I haven’t provided a fix yet was because I don’t know which version of clang introduced this warning. Looks like clang 14 didn’t recognize it and was complaining. Plan to have a look again today unless you know the fix let me know.

abbaswasim: Right yes it’s the android clang 14 that doesn’t recognize it

wasimabbas-arm:
Phew, that was painful. Anyways looks like everything pass now.

MarkCallow (approved):

PR #981, opened 2025-02-12
Fix python artifact upload

MarkCallow:
Stop using now broken `upload-artifact-as-is` action.


MarkCallow:
Builds cancelled because there is are no changes in main since the branch being merged and the branch build was successful. Merge now.

PR #989, opened 2025-03-04
Set up workflow to build just the documentation.

MarkCallow:
The plan is to use this workflow to build the docs for PRs and pushes to main. Due to the tight deadline for 4.4, there is no time to update the per-platform software build workflows to only build the docs only when building for a release so only the Ubuntu build is enabled now. This is used to deploy the docs to GitHub Pages. For now this only happens for a release tag. Once a location is set-up the plan is to deploy docs for release tags to somewhere on khronos.org and deploy docs to GitHub Pages on any pushes to main.

The following other changes are incorporated in this PR:

- Make a separate `pyktx.doc` target instead of bolting the commands to build the documents on to the `pyktx` target.
- Fix CMake 3.31 warnings in the pyktx CMakeLists.txt.
- Update copyright, author and e-mail address in the pyktx package and documentation.
- Add lists of files to ignore for in each workflow file.


MarkCallow:
@ShukantPal this PR adds a workflow to build just the KTX-Software documentation. For the Python binding I created a new `pyktx.doc` target separate from the pyktx target. But the custom command to build the documentation calls the pyktx `buildscript.py` which causes an attempt to link to libktx. I do not want to have to build libktx in this workflow. I tried commenting out the call to `buildscripts.py` but then I get errors like the following:

```
WARNING: autodoc: failed to import module 'ktx_hash_list' from module 'pyktx'; the following exception was raised:
No module named 'pyktx.native'
```

How can I build the documentation without building libktx?

Issue #996, opened 2025-03-19
Issues in the CMakeLists.txt


DanielGibson:
While trying to figure out how to best integrate libktx in my project (see also https://github.com/KhronosGroup/KTX-Software/discussions/995), I discovered into several problems in the main CMakeLists.txt.

* **`-DASTCENC_DECOMPRESSOR=ON` breaks the build** (at least on Linux, but probably everywhere). CMake will show errors like  
  ```
   CMake Error at CMakeLists.txt:1241 (set_property):
     set_property could not find TARGET astcenc-sse2-static.  Perhaps it has not
     yet been created.
   ```  
  So apparently CMake always references something from astcenc that's only built if `ASTCENC_DECOMPRESSOR=ON` is **not** set.
  - Ugly workaround that however breaks the build *without* `ASTCENC_DECOMPRESSOR=ON`: Comment out lines 1241 - 1250 (`set_property(TARGET ${ASTCENC_LIB_TARGET} PROPERTY POSITION_INDEPENDENT_CODE ON)` until the next empty line) and line 1299 (`list(APPEND KTX_INSTALL_TARGETS ${ASTCENC_LIB_TARGET})`)
* `build/libktx.a` does *not* include the astcenc object files. Those are hidden in `build/external/astc-encoder/Source/libastcenc-sse2-static.a`. Shouldn't they included in `libktx.a`, just like the basisu object files are?
  - If I comment out enough things in CMakeList.txt to get it to build with `ASTCENC_DECOMPRESSOR=ON`, it's basically the same situation, except then it's  ./external/astc-encoder/Source/libastc**dec**-sse2-static.a
* Some of the code integrated in the project requires **`-fno-strict-aliasing`**; at least stb_image_write.h and Basis Universal do *(according to comments in basisu transcoder headers. Unsure if that also affects the encoder? Their own CMakeLists.txt seems to set it for everything)*.  
  However, that option is not set. I verified that the the basisu source files are currently *not* built with `-fno-strict-aliasing`.
  Instead `-Wno-strict-aliasing` is passed to ignore warnings about strict aliasing 🙃 (to be fair, I've never seen that warning actually warn about anything anywhere, so this might be a noop)

Some further things that aren't *issues* as in "obviously broken", but would be quite useful to make integrating libktx easier (which might further the adoption of KTX2 in native code projects - so far my impression is that it's mostly used on the web?):

* It would be nice if one had the option to **build only `libktx_read.a`**
[...]

PR #998, opened 2025-03-22
Fix issues in CMakeLists.txt (see #996)

DanielGibson:
I fixed some of the issues mentioned in #996:
* ~~`-DASTCENC_DECOMPRESSOR=ON` doesn't break the build anymore~~
  `ASTCENC_DECOMPRESSOR` is forced OFF and not shown as an option in cmake-gui anymore.
* Set `-fno-strict-aliasing` for basisu
* remove the unused `${BASISU_ENCODER_C_SRC}` variable from the basisu source list (this one wasn't mentioned in #996, I just noticed this while fixing the other stuff)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -764,7 +764,6 @@ macro(common_libktx_settings target enable_write library_type)
         PRIVATE
             lib/basis_encode.cpp
             lib/astc_codec.cpp
-            ${BASISU_ENCODER_C_SRC}
             ${BASISU_ENCODER_CXX_SRC}
             lib/writer1.c
             lib/writer2.c
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -837,6 +837,30 @@ PRIVATE
     BASISD_SUPPORT_KTX2=0
 )
 
+# helper function to append COMPILE_OPTIONS to source file properties
+# !! if SRCFILES is a list, remember to quote it                              !!
+# !! like add_source_file_compile_options("${MY_SRC_LIST}"  "-Wall")          !!
+# !! or add_source_file_compile_options("dir/foo.cpp;dir/bar.cpp"  "-Wextra") !!
+function(add_source_file_compile_options SRCFILES OPTIONS)
+    foreach(src_file ${SRCFILES})
+        get_source_file_property(cur_options
+            "${src_file}"
+            COMPILE_OPTIONS
+        )
+        if(cur_options)
+            set_source_files_properties(
+                "${src_file}"
+                PROPERTIES COMPILE_OPTIONS "${cur_options};${OPTIONS}"
+            )
+        else() # if cur_options were undefined or empty ("")
+            set_source_files_properties(
+                "${src_file}"
+                PROPERTIES COMPILE_OPTIONS "${OPTIONS}"
+            )
+        endif()
+    endforeach()
+endfunction()
+
 # Turn off these warnings until Rich fixes the occurences.
 # It it not clear to me if generator expressions can be used here
 # hence the long-winded way.
@@ -847,26 +871,22 @@ elseif(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
         # It's too much work to discriminate which files need which warnings
         # disabled.
         ${BASISU_ENCODER_CXX_SRC}
-        PROPERTIES COMPILE_OPTIONS "-Wno-sign-compare;-Wno-unused-variable;-Wno-class-memaccess;-Wno-misleading-indentation;-Wno-extra;-Wno-deprecated-copy;-Wno-parentheses"
+        PROPERTIES COMPILE_OPTIONS "-fno-strict-aliasing;-Wno-sign-compare;-Wno-unused-variable;-Wno-class-memaccess;-Wno-misleading-indentation;-Wno-extra;-Wno-deprecated-copy;-Wno-parentheses"
     )
     set_source_files_properties(
         external/basisu/transcoder/basisu_transcoder.cpp
-        PROPERTIES COMPILE_OPTIONS "-Wno-sign-compare;-Wno-unused-function;-Wno-unused-variable;-Wno-class-memaccess;-Wno-maybe-uninitialized"
+        PROPERTIES COMPILE_OPTIONS "-fno-strict-aliasing;-Wno-sign-compare;-Wno-unused-function;-Wno-unused-variable;-Wno-class-memaccess;-Wno-maybe-uninitialized"
     )
     if (${CMAKE_CXX_COMPILER_VERSION} VERSION_GREATER_EQUAL "11")
         if (${CMAKE_CXX_COMPILER_VERSION} VERSION_LESS "12" )
-        # Version 11 raises several stringop-overflow warnings in some
-        # very hard to decipher code. They appear to be bogus based on
-        # the facts that we have never seen a crash and version 12 no
-        # longer raises the warnings.
-        get_source_file_property(cur_options
-            external/basisu/encoder/basisu_comp.cpp
-            COMPILE_OPTIONS
-        )
-        set_source_files_properties(
-            external/basisu/encoder/basisu_comp.cpp
-            PROPERTIES COMPILE_OPTIONS "${cur_options};-Wno-stringop-overflow"
-        )
+            # Version 11 raises several stringop-overflow warnings in some
+            # very hard to decipher code. They appear to be bogus based on
+            # the facts that we have never seen a crash and version 12 no
+            # longer raises the warnings.
+            add_source_file_compile_options(
+                external/basisu/encoder/basisu_comp.cpp
+                "-Wno-stringop-overflow"
+            )
         endif()
     endif()
     if (${CMAKE_CXX_COMPILER_VERSION} VERSION_GREATER_EQUAL "12.0")
@@ -874,28 +894,16 @@ elseif(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
         # There seems no way for the index calculated by the code at
         # line 326, where the error is raised, to be > the array length.
         # Also we have never seen any crashes.
-        set_source_files_properties(
+        add_source_file_compile_options(
             external/basisu/encoder/basisu_uastc_enc.cpp
-            PROPERTIES COMPILE_OPTIONS "-Wno-stringop-overflow"
+            "-Wno-stringop-overflow"
         )
     endif()
     if (${CMAKE_CXX_COMPILER_VERSION} VERSION_GREATER_EQUAL "14.0")
         # Version 14 raises stringop-overflow on these files.
-        get_source_file_property(cur_options
-            external/basisu/encoder/basisu_comp.cpp
-            COMPILE_OPTIONS
-        )
-        set_source_files_properties(
-            external/basisu/transcoder/basisu_transcoder.cpp
-            PROPERTIES COMPILE_OPTIONS  "${cur_options};-Wno-stringop-overflow"
-        )
-        get_source_file_property(cur_options
-            external/basisu/encoder/basisu_bc7enc.cpp
-            COMPILE_OPTIONS
-        )
-        set_source_files_properties(
-            external/basisu/encoder/basisu_bc7enc.cpp
-            PROPERTIES COMPILE_OPTIONS  "${cur_options};-Wno-stringop-overflow"
+        add_source_file_compile_options(
+            "external/basisu/transcoder/basisu_transcoder.cpp;external/basisu/encoder/basisu_bc7enc.cpp"
+            "-Wno-stringop-overflow"
         )
     endif()
 elseif(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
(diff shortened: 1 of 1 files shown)


DanielGibson:
uhh.. this CLA thing is kinda fishy.

The text of the conditions was in light grey on lighter grey, and when I logged in there with Github (hoping that I could then select what applies to me and see the text black on white) it automatically signs it.
Going back to that page it turns out that it doesn't even have much useful information and the actual CLA is somewhere else, only "linked" there as a non-clickable URL in that unreadable text...

To be clear, I'm fine with the actual CLA conditions, I just find the implementation dubious.

CMakeLists.txt
```diff
@@ -946,6 +945,26 @@ else()
     message(FATAL_ERROR "${CMAKE_CXX_COMPILER_ID} not yet supported.")
 endif()
 
+if(NOT MSVC)
```
MarkCallow, line 948: Simply change lines 851 and 855. In 851 change -Wno-strict-aliasing to `-fno-strict-aliasing`. In 855 add it. There is no need for this whole new set of lines.

CMakeLists.txt
```diff
@@ -1197,9 +1216,15 @@ if(NOT ${universal_build})
         set(ASTCENC_ISA_SSE2 OFF)
     endif()
 
+    if(${ASTCENC_DECOMPRESSOR})
```
MarkCallow, line 1219: If somebody sets this and is building the ktx (libktx) target, the build will break due to undefined functions.

Either leave things and they are and always build the full astcenc library, in the knowledge that only called functions will be linked to the shared library or executable, or set up to build both libraries (with the same ISA) and make `ktx` dependent on `astcenc` and `ktx_read` dependent on `astcdec`. The penalty to this approach is that the decompressor will be compiled twice: once for `astcenc` and once for `astcdec`.

MarkCallow (changes requested):

CMakeLists.txt
DanielGibson: The root problem I'm trying to fix is that the ASTCENC_DECOMPRESSOR option exists, is shown in `cmake-gui` when configuring this project, sounds like something I'd want to use (as long as there is no way to entirely disable any usage of ASTCENC), but setting it causes non-obvious CMake errors.

Your suggestions wouldn't help with that particular problem

CMakeLists.txt
DanielGibson: You mean **851** (or 850 in my file), **889** (`basisu_transcoder.cpp` is the only one used in libktx_read), and for **clang** line **933** for the transcoder and for the **encoder** I'd still have to add **this same block** because for clang source file properties for `${BASISU_ENCODER_CXX_SRC}` is only set `if (${clang_version} VERSION_GREATER_EQUAL "14.0")`.

I'll do this change if you still want it (considering that clang also needs this so it's more than two lines that need to be adjusted), but will wait for your feedback.

CMakeLists.txt
DanielGibson: I could use `set(ASTCENC_DECOMPRESSOR OFF CACHE INTERNAL "")` to make sure that the ASTCENC_DECOMPRESSOR option is disabled (OFF) and not shown to the user in cmake-gui (INTERNAL)

MarkCallow: Yes this is the most straightforward thing to do. Setting a local variable with the same name works to hide the configuration variable too. That is how we hide ASTCENC_UNIVERSAL_BUILD.

Sorry I didn't pick up on the root problem before.

CMakeLists.txt
MarkCallow: Are you getting warnings or errors when building with clang? Our CI is not showing any and it builds with clang on iOS, macOS, with Emscripten and via ClangCL on Windows.

DanielGibson: This is not about warnings or errors, but about disabling "strict aliasing" optimizations, which (according to comments in the basisu source files and [their CMakeLists.txt](https://github.com/BinomialLLC/basis_universal/blob/42c21d8066cd4b3f33e29272b61170e3b2e243bc/CMakeLists.txt#L1) could break their code.

MarkCallow: Keep this block but please remove the now unnecessary `-Wno-strict-aliasing' flag from line 850 as this renders it unnecessary

MarkCallow, line 948: Change to `if(NOT CMAKE_CXX_COMPILER_ID MATCHES "MSVC")`. If I recall correctly MSVC will be set when using ClangCL because it supports the same command line options.

MarkCallow (changes requested):
Our CI for GNU/Linux, iOS, macOS and the web is undergoing remodelling. Once that is complete I will ask you to rebase this so we can run it through all the compilation checks. I will not merge it until it successfully passes CI. Sorry for the delay.

DanielGibson:
I did the requested changes and rebased to the current `main` branch

CMakeLists.txt
DanielGibson: I mentioned most of it in #996 but I guess I should've mentioned that this option was shown in cmake-gui.
Either way, I replaced my prior fix with `set(ASTCENC_DECOMPRESSOR OFF CACHE INTERNAL "")`

DanielGibson:
image
Why does this now pretend that I didn't sign the CLA yet?
I did and that was also reflected in the status here, why is it reset?

MarkCallow:
> Why does this now pretend that I didn't sign the CLA yet?
> I did and that was also reflected in the status here, why is it reset?

I don't know. Although this is the first time I am aware of a reset happening during an open PR my own status has been reset between PRs several times. CLAssistant seems to be a black hole. @KhronosWebservices any thoughts on how to prevent this?

CMakeLists.txt
```diff
@@ -1233,6 +1252,9 @@ if(NOT ${universal_build})
     endif()
 endif()
 
+# setting ASTCENC_DECOMPRESSOR to ON breaks the build, so force ot to OFF
```
MarkCallow, line 1255: s/ot/to/

MarkCallow (commented):
Looks good now. Thanks. Please fix the typos I've pointed out. Then we'll have to wait until the CI remodel is finished at which point another rebase will be needed.

CMakeLists.txt
DanielGibson: should've been "it" actually ("force *it* to off") - will fix, thanks for pointing it out :)

CMakeLists.txt
```diff
@@ -946,6 +945,26 @@ else()
...
+    )
+    set_source_files_properties(
+        external/basisu/transcoder/basisu_transcoder.cpp
+        PROPERTIES COMPILE_OPTIONS "${cur_options};-fno-strict-aliasing"
+    )
+endif()
+
 # Retrieve the final set of properties for use by the transcodetests.
 # We do this because the CMake feature that would allow the transcodetests
 # target to retrieve these from the ktx target are not available until
```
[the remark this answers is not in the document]
DanielGibson: done and pushed

MarkCallow (approved):

MarkCallow:
Please investigate the GCC compile failure in the MingW build. Is this due to the change here or a change in the compiler used in the runner? I do not think it is the latter because the runner is using a version of GCC 12 which is now quite old.

DanielGibson:
```
[75/165] Building CXX object CMakeFiles/ktx.dir/Release/external/basisu/encoder/basisu_uastc_enc.cpp.obj
FAILED: CMakeFiles/ktx.dir/Release/external/basisu/encoder/basisu_uastc_enc.cpp.obj 
C:\mingw64\bin\g++.exe -DBASISD_SUPPORT_FXT1=0 -DBASISD_SUPPORT_KTX2=1 -DBASISD_SUPPORT_KTX2_ZSTD=0 -DBASISU_NO_ITERATOR_DEBUG_LEVEL -DBASISU_SUPPORT_OPENCL=0 -DBASISU_SUPPORT_SSE=1 -DKTX_API=__declspec(dllexport) -DKTX_FEATURE_KTX1 -DKTX_FEATURE_KTX2 -DKTX_FEATURE_WRITE -DLIBKTX -DSUPPORT_SOFTWARE_ETC_UNPACK=1 -DWIN32_HAS_PTHREADS -Dktx_EXPORTS -DCMAKE_INTDIR=\"Release\" -ID:/a/KTX-Software/KTX-Software/include -ID:/a/KTX-Software/KTX-Software/external/basisu/transcoder -ID:/a/KTX-Software/KTX-Software/external -ID:/a/KTX-Software/KTX-Software/external/basisu/zstd -ID:/a/KTX-Software/KTX-Software/utils -ID:/a/KTX-Software/KTX-Software/external/dfdutils -ID:/a/KTX-Software/KTX-Software/external/basisu -ID:/a/KTX-Software/KTX-Software/external/astc-encoder/Source -isystem D:/a/KTX-Software/KTX-Software/other_include -O3 -DNDEBUG -Wall -Wextra -Werror -O3 -ffp-contract=off -Wno-cast-function-type -Wno-pedantic -msse4.1 -Wno-sign-compare -Wno-unused-variable -Wno-class-memaccess -Wno-misleading-indentation -Wno-extra -Wno-deprecated-copy -Wno-parentheses -fno-strict-aliasing -MD -MT CMakeFiles/ktx.dir/Release/external/basisu/encoder/basisu_uastc_enc.cpp.obj -MF CMakeFiles\ktx.dir\Release\external\basisu\encoder\basisu_uastc_enc.cpp.obj.d -o CMakeFiles/ktx.dir/Release/external/basisu/encoder/basisu_uastc_enc.cpp.obj -c D:/a/KTX-Software/KTX-Software/external/basisu/encoder/basisu_uastc_enc.cpp
In file included from C:/mingw64/lib/gcc/x86_64-w64-mingw32/12.2.0/include/c++/bits/stl_pair.h:61,
                 from C:/mingw64/lib/gcc/x86_64-w64-mingw32/12.2.0/include/c++/bits/stl_algobase.h:64,
                 from C:/mingw64/lib/gcc/x86_64-w64-mingw32/12.2.0/include/c++/bits/specfun.h:45,
                 from C:/mingw64/lib/gcc/x86_64-w64-mingw32/12.2.0/include/c++/cmath:1935,
                 from C:/mingw64/lib/gcc/x86_64-w64-mingw32/12.2.0/include/c++/math.h:36,
                 from D:/a/KTX-Software/KTX-Software/external/basisu/transcoder/basisu.h:59,
                 from D:/a/KTX-Software/KTX-Software/external/basisu/encoder/basisu_etc.h:16,
                 from D:/a/KTX-Software/KTX-Software/external/basisu/encoder/basisu_uastc_enc.h:16,
                 from D:/a/KTX-Software/KTX-Software/external/basisu/encoder/basisu_uastc_enc.cpp:15:
In function 'std::_Require<std::__not_<std::__is_tuple_like<_Tp> >, std::is_move_constructible<_Tp>, std::is_move_assignable<_Tp> > std::swap(_Tp&, _Tp&) [with _Tp = unsigned char]',
    inlined from 'void basisu::pack_uastc(basist::uastc_block&, const uastc_encode_results&, const etc_block&, uint32_t, const eac_a8_block&, bool, bool)' at D:/a/KTX-Software/KTX-Software/external/basisu/encoder/basisu_uastc_enc.cpp:326:18:
C:/mingw64/lib/gcc/x86_64-w64-mingw32/12.2.0/include/c++/bits/move.h:205:11: error: writing 2 bytes into a region of size 0 [-Werror=stringop-overflow=]
  205 |       __a = _GLIBCXX_MOVE(__b);
      |           ^
D:/a/KTX-Software/KTX-Software/external/basisu/encoder/basisu_uastc_enc.cpp: In function 'void basisu::pack_uastc(basist::uastc_block&, const uastc_encode_results&, const etc_block&, uint32_t, const eac_a8_block&, bool, bool)':
D:/a/KTX-Software/KTX-Software/external/basisu/encoder/basisu_uastc_enc.cpp:253:25: note: at offset 18 into destination object 'endpoints' of size 18
  253 |                 uint8_t endpoints[18];
      |                         ^~~~~~~~~
cc1plus.exe: all warnings being treated as errors
```

I don't see how my changes should cause this and I don't even think it's a valid warning.

If I understand correctly, the compiler says that 
https://github.com/KhronosGroup/KTX-Software/blob/70ed9b9c5cb80867a045beffb0f1ef92e6648db3/external/basisu/encoder/basisu_uastc_enc.cpp#L326 
can access `endpoints[18]` (*"at offset 18 into destination object 'endpoints' of size 18"*) which would be out of bounds because it only has 18 elements, see https://github.com/KhronosGroup/KTX-Software/blob/70ed9b9c5cb80867a045beffb0f1ef92e6648db3/external/basisu/encoder/basisu_uastc_enc.cpp#L253
But I have no idea where that assumption comes from, because `c < total_comps`, which comes from some global array: https://github.com/KhronosGroup/KTX-Software/blob/70ed9b9c5cb80867a045beffb0f1ef92e6648db3/external/basisu/encoder/basisu_uastc_enc.cpp#L256
I don't know why the compiler assumes that `c` can be `9` (then `endpoints[c * 2 + 0]` would be `endpoints[18]`). It probably can't "see" the values of that array (it's defined in external/basisu/transcoder/basisu_transcoder.cpp and its biggest value is `4` so `c` will be **<= 3**), but if it assumes that it can have invalid values, it could as well assume that `c` is `254` and that it would access `endpoints[508]` and `endpoints[509]` which is even more out of bounds...

MarkCallow:
> Did the compiler version on the runner change? Even if GCC12 is "quite old", if it was an even older version before it could be that the warning only turned up in GCC12

Thank you for the analysis. I agree the warning seems strange. However the compiler version did not change and builds on other branches are compiling without warning using this compiler. See https://github.com/KhronosGroup/KTX-Software/actions/runs/14324522719/job/40147530265?pr=1010. Therefore something here is triggering the compiler into issuing that warning. ~~Please check if it is the removal of the suppression of the strict aliasing warning or the addition of `-fno-strict-aliasing. I want to understand what is going on before, possibly, adding yet another warning suppression.~~ Never mind. I just figured out the cause. This code disables stringop-overflow on 2 encoder files and on the transcoder.
https://github.com/KhronosGroup/KTX-Software/blob/70ed9b9c5cb80867a045beffb0f1ef92e6648db3/CMakeLists.txt#L873-L901

However your change is taking the COMPILE_OPTIONS from `basisu_comp.cpp` adding `-fno-strict-aliasing` and applying those to *all* the encoder source thereby removing the `-Wno-stringop-overflow` from the 2 files where it is set. Please fix.

I just noticed a similar bug in the code above w.r.t `basisu_transcoder.cpp`. It is taking the COMPILE_OPTIONS from basisu_comp.cpp and applying them to `basisu_transcoder.cpp` Please fix this too.

DanielGibson:
Oh, right.

These hyperspecific disabled warnings per source file and per compiler version are confusing and painful to handle :-/
I'll try to figure out a better way to add `-fno-strict-aliasing`.

DanielGibson:
There are other similar bugs in that cmake code, by the way.

https://github.com/KhronosGroup/KTX-Software/blob/v4.4.0/CMakeLists.txt#L885-L892
gets the COMPILE_OPTIONS from `basisu_comp.cpp` and then sets them for `basisu_transcoder.cpp`

https://github.com/KhronosGroup/KTX-Software/blob/v4.4.0/CMakeLists.txt#L878-L881
sets the compile options for `basisu_uastc_enc.cpp` without getting them first, so the generic ones set in https://github.com/KhronosGroup/KTX-Software/blob/v4.4.0/CMakeLists.txt#L847-L852 are lost

DanielGibson:
should be better now :)

MarkCallow (approved):
Thank you for this nice refactoring.

I will wait until Emscripten, iOS, Linux and macOS CI is functional again before merging. I'll let you know when it is functional then you can rebase this and let the updated CI run it.

DanielGibson:
done (you may have to manually approve running github actions for them to run)

PR #1004, opened 2025-03-28
Linux and MacOS workflows

MathiasMagnus:


.github/workflows/macos.yml
```diff
@@ -0,0 +1,431 @@
+# Copyright 2025 The Khronos Group Inc.
+# SPDX-License-Identifier: Apache-2.0
+name: KTX-Software Linux CI
```
MarkCallow, line 3: name and filename don't match

.github/workflows/macos.yml
```diff
@@ -0,0 +1,431 @@
...
+      - .travis.yml
+      - README.md
+      - CODE_OF_CONDUCT.md
+      - CONTRIBUTING.md
+      - LICENSE.md
+      - LICENSES
+      - RELEASE_NOTES.md
+      - REUSE.toml
+      - install-gitconfig*
+      - vcpkg.json
```
MarkCallow, line 37: Remove these vcpkg files from here. Changes to them can affect the loadtests build so it needs to be run.

.github/workflows/macos.yml
```diff
@@ -0,0 +1,431 @@
...
+            echo "CC=clang" >> $GITHUB_ENV
+            echo "CXX=clang++" >> $GITHUB_ENV
+          else
+            echo "Unknown compiler toolchain set"
+            exit 1
+          fi
+
+      - name: before_install
+        run: |
+          echo -n "Running on the following travis CI runner: " && uname -a
```
MarkCallow, line 263: Not travis any more.

.github/workflows/macos.yml
```diff
@@ -0,0 +1,431 @@
...
+          set +e
+
+      - name: before_deploy
+        run: |
+          gem install octokit:'~> 4.15.0' mime-types
+          case "${RUNNER_OS:-Linux}" in
+          Linux)
+            ;;
+          macOS)
+            # Only notarize if we're going to deploy. I.e on a tag. Also check for
```
MarkCallow, line 387: Move this so notarization is done for every package that will be saved as an artifact of the build. It was limited to deployment on Travis because no artifacts were saved.

.github/workflows/macos.yml
```diff
@@ -0,0 +1,431 @@
...
+            # release description.
+            awk '/^Release Notes/,/^## Version/ { next }
+                 ! /<!-- Copyright/ && ! /<!-- SPDX/ { print }' RELEASE_NOTES.md \
+            > $REL_DESC_FILE
+            # dpl v2 alternative
+            #RELNOTES=$(awk '/^Release Notes/,/^## Version/ { next }
+            #     ! /<!-- Copyright/ && ! /<!-- SPDX/ { print }' RELEASE_NOTES.md)
+          fi
+
+      - name: Upload pages artifact
```
MarkCallow, line 411: Don't need docs deployment in this workflow. That is now done in docs.yml. Still need to build the docs in this though for inclusion in the built packages.

MarkCallow:
Please fix the syntax error at line 70 in publish-pyktx.yml. Because the YAML is broken the .yml file is shown strangely in the list of workflows even when not working on the github-actions branch.

.github/workflows/macos.yml
MathiasMagnus: Yes, the PR is draft, not intended for review yet (although some things can be fixed now). (GA remembers workflow names, so I didn't want to give it Linux/MacOS becuase ultimately I wanted to have the platforms be separate workflows, but the 1st crude port behaves much like the Travis version. It's in an interim state and the final version will be clean from this perspective.)

MarkCallow: Understood. Thanks.

.github/workflows/macos.yml
```diff
@@ -384,7 +382,7 @@ jobs:
 
       - name: before_deploy
         run: |
-          gem install octokit:'~> 4.15.0' mime-types
+          sudo gem install octokit:'~> 4.15.0' mime-types
```
MarkCallow, line 385: I made this comment in the commit but it has not appeared in the PR so I am adding it again here.

You don't need octokit. It is a dependency of Travis's deployment package "dpl" and at some point in the past we needed a newer version than they had in their standard package. GitHub release deployment is totally different, as far as I am aware.

.github/workflows/macos.yml
```diff
@@ -0,0 +1,287 @@
...
+
+      BUILD_DIR: build
+      REL_DESC_FILE: "build/rel_desc.md"
+
+      # Map test matrix elements to environmental variables, so that scripts launched from CI scripts may react to them.
+      # Some matrix variables have defaults if unspecified by the matrix
+      CMAKE_GEN: ${{ matrix.generator }}
+      ARCH: ${{ matrix.arch }}
+      ARCHS: ${{ matrix.arch }} # MacOS uses ARCHS
+      CHECK_MKVK: ${{ matrix.check_mkvk && matrix.check_mkvk || 'NO' }}
```
MarkCallow, line 132: You've used this pattern in many places and files, sometimes with && and || as here, sometimes with || and ||. You have changed some existing tests to this pattern too. What is the purpose?

MarkCallow (changes requested):
Since wrote that you think you are finished, I have reviewed even though the PR is still set as draft. I have included in this review the changes I requested  in my e-mail yesterday.

The "big red button" looks fine. The workflows I want in it are android, linux, macos, mingw and windows

.github/workflows/linux.yml
```diff
@@ -0,0 +1,329 @@
...
+        if: matrix.options.package == 'YES' && matrix.options.py == 'ON'
+        uses: actions/upload-artifact@v4
+        with:
+          name: pyktx-${{env.PACKAGE_NAME_SUFFIX}}
+          path: ${{env.BUILD_DIR}}/${{env.PYTHON_DIST_DIR}}
+
+      - name: Upload to Release
+        uses: softprops/action-gh-release@v2
+        if: matrix.options.package == 'YES' && github.event_name == 'push' && github.ref_type == 'tag'
+        with:
```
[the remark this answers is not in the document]
MarkCallow: Figure out a way so this is only done once. It seems wasteful to do it for every job that deploys a package.

scripts/build_wasm_docker.sh
```diff
@@ -94,9 +94,9 @@ fi
 
 echo "Software versions"
 echo '*****************'
-docker exec -it emscripten sh -c "emcc -v; echo '********'"
```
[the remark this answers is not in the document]
MathiasMagnus: It adds interactivity, which in the context of CI and exec in particular doesn't mean much. More importantly, executing interactively in the shell of CI threw a [hard error](https://github.com/KhronosGroup/KTX-Software/actions/runs/14238605195/job/39903080704#step:8:73), arguing the input device isn't a TTY.

.github/workflows/macos.yml
MathiasMagnus: The purpose is to replicate the behavior of Travis env vars. Travis test matrix definition sets env vars directly, but more importantly defaults were set globally, and matrix elements when omitting some variable of the matrix will inherit the default. The `() && () || ()` is an atrocity of a syntax for the [ternary conditional operator](https://en.wikipedia.org/wiki/Ternary_conditional_operator) we all know and love, but this is what [GA uses](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#example). The Windows workflow too has to relay matrix variables to env vars so that PS scripts can see them. GA however doesn't inherit env vars in the same way Travis does, but it [overrides](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#configuration-variable-precedence) them by the most specialized context. The ternary operator is used to replicate the "take the value of the matrix elem if present, otherwise do the default". If the matrix elem is absent dereferencing will result in `null` which is a ["falsy"](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#literals) value evaluating to false, thus forcing the ternary to take the default value provided in the false case. Using the ternary operator allows omission of some matrix vars like it was done in Travis.

scripts/build_wasm_docker.sh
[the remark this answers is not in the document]
MarkCallow: Thanks. I wonder how it ever worked on Travis.

.github/workflows/macos.yml
MarkCallow: And the variant with `() || () || ()`?

MarkCallow:
Please remove all the checkmkvk remnants from linux.yml and macos.yml. The Check mvkk workflow handles that now. The code was still in .travis.yml because I planned to not include it when porting to GHA so I simply disabled it by never setting the relevant environment variable.

.github/workflows/macos.yml
```diff
@@ -0,0 +1,263 @@
...
+            # not copy them from the gh-pages branch to the website. This
+            # file says no Jekyll files here. Treat all as ordinary files.
+            touch $BUILD_DIR/docs/html/.nojekyll
+          fi
+          set +e
+
+      - name: Notarize
+        if: matrix.options.platform == 'macOS'
+        run: |
+          ./scripts/notarize.sh $BUILD_DIR/KTX-Software-*.pkg $APPLE_ID $DEVELOPMENT_TEAM $ALTOOL_PASSWORD && export NOTARIZED="true"
```
MarkCallow, line 226: Notarization is failing unnoticed in the test builds you did. The reason for the failure is that I changed the team environment variable name to `APPLE_DEVELOPMENT_TEAM` to disambiguate because the secrets list now has Windows and Apple stuff in it.

I'm not sure why the failure was unnoticed. I thought it might be because it was called in an `if`, `else` block but you have `set -e` so I'm not sure.

MarkCallow: Just looked at the latest log. The failure is being noticed now it is a separate step.

MarkCallow, line 226: Remove `&& export NOTARIZED="true"`. It was only there because the before_deploy phase in Travis was called because each deploy and there were multiple deploys in the job.

.github/workflows/macos.yml
```diff
@@ -0,0 +1,287 @@
...
+            git clone https://github.com/microsoft/vcpkg
+            cd vcpkg && ./bootstrap-vcpkg.sh -disableMetrics && echo "VCPKG_ROOT=`pwd`" >> $GITHUB_ENV
+            popd
+          fi
+
+      - name: install
+        run: |
+            if [ "$CHECK_REUSE" != "ONLY" -a "$CHECK_MKVK" != "ONLY" ]; then
+              if [ "$FEATURE_DOC" = "ON" ]; then
+                  brew install --formula doxygen
```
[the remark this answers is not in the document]
MathiasMagnus: Done.

.github/workflows/macos.yml
```diff
@@ -0,0 +1,263 @@
...
+          if [ "$DEPLOY_DOCS" = "YES" ]; then
+            # Some files in the pyktx docs have an _ prefix so Jekyll will
+            # not copy them from the gh-pages branch to the website. This
+            # file says no Jekyll files here. Treat all as ordinary files.
+            touch $BUILD_DIR/docs/html/.nojekyll
+          fi
+          set +e
+
+      - name: Notarize
+        if: matrix.options.platform == 'macOS'
```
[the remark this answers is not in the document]
MathiasMagnus: And is `CODE_SIGN_IDENTITY` set? I don't think that's set either as a secret or a variable.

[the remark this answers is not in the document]
MarkCallow: It is only needed for code signing not notarization.

.github/workflows/macos.yml
MathiasMagnus: That's a typo. (Fun fact: it's a ternary operator in the sense that it uses boolean shortcircuit to implement the ternary nature.)

.github/workflows/macos.yml
```diff
@@ -0,0 +1,287 @@
...
+      PLATFORM: ${{ matrix.options.platform }}
+      FEATURE_DOC: ${{ matrix.options.doc }}
+      FEATURE_JNI: ${{ matrix.options.jni }}
+      FEATURE_LOADTESTS: ${{ matrix.options.loadtests }}
+      FEATURE_PY: ${{ matrix.options.py }}
+      FEATURE_TESTS: ${{ matrix.options.tests && matrix.options.tests || 'ON' }}
+      FEATURE_TOOLS: ${{ matrix.options.tools }}
+      FEATURE_TOOLS_CTS: ${{ matrix.options.tools_cts }}
+      PACKAGE: ${{ matrix.options.package || matrix.options.package || 'NO' }}
+      DEPLOY_DOCS: ${{ matrix.options.deploy_docs || matrix.options.deploy_docs || 'NO' }}
```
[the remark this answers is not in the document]
MathiasMagnus: Removed.

.github/workflows/macos.yml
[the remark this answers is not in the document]
MathiasMagnus: @MarkCallow I do not know why, but even though the necessary [env vars are set](https://github.com/KhronosGroup/KTX-Software/blob/0021be17814cfeec558ceb56a5f40bbc0fbc71b4/.github/workflows/macos.yml#L123-L127), it [says that the resulting package does not have valid signature](https://github.com/KhronosGroup/KTX-Software/actions/runs/14383602954/job/40333321773?pr=1004#step:10:38).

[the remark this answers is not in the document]
MarkCallow: One is missing: MACOS_CERTIFICATES_P12. If not set `scripts/before_build_macos.sh` exits immediately as it can't add the certificates to the MacOS keychain.

.github/workflows/macos.yml
```diff
@@ -0,0 +1,260 @@
...
+                 # otherwise subsequent commands will swallow bad exit status.
+          if [ "$PLATFORM" = "macOS" ]; then
+            ./scripts/build_macos.sh
+          else
+            ./scripts/build_ios.sh
+          fi
+          set +e
+
+      - name: Notarize
+        if: matrix.options.platform == 'macOS'
```
MarkCallow, line 222: Only run this step if secrets are available, e.g APPLE_ID or DEVELOPMENT_TEAM is set. PR's from forks, where secrets are not available, must still run.

.github/workflows/macos.yml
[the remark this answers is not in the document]
MarkCallow: > While at it, please add the following at the end of `scripts/before_build_macos.sh`:
> 
> ```
> unset MACOS_CERTIFICATES_P12
> ```

That isn't going to work given the way GHA handles environment variables. Maybe it is better not to set this environment variable and instead pass the value directly from the secret as an argument to `scripts/before_build_macos.sh` which will need modifying to handle it.

.github/workflows/macos.yml
```diff
@@ -0,0 +1,287 @@
...
+          fi
+          if [ "$DEPLOY_DOCS" = "YES" ]; then
+            # Some files in the pyktx docs have an _ prefix so Jekyll will
+            # not copy them from the gh-pages branch to the website. This
+            # file says no Jekyll files here. Treat all as ordinary files.
+            touch $BUILD_DIR/docs/html/.nojekyll
+          fi
+          set +e
+
+      - name: before_deploy
```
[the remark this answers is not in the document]
MathiasMagnus: Removed.

.github/workflows/macos.yml
```diff
@@ -0,0 +1,287 @@
...
+              ./scripts/build_macos.sh
+              ./scripts/notarize.sh $BUILD_DIR/KTX-Software-*.pkg $APPLE_ID $DEVELOPMENT_TEAM $ALTOOL_PASSWORD && export NOTARIZED="true"
+            else
+              ./scripts/build_ios.sh
+            fi
+          fi
+          if [ "$CHECK_MKVK" = "ONLY" ]; then
+            scripts/check_mkvk.sh
+          fi
+          if [ "$DEPLOY_DOCS" = "YES" ]; then
```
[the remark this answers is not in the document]
MathiasMagnus: Removed.

.github/workflows/linux.yml
```diff
@@ -0,0 +1,329 @@
...
+      PLATFORM: ${{ matrix.options.platform }}
+      FEATURE_DOC: ${{ matrix.options.doc }}
+      FEATURE_JNI: ${{ matrix.options.jni }}
+      FEATURE_LOADTESTS: ${{ matrix.options.loadtests }}
+      FEATURE_PY: ${{ matrix.options.py }}
+      FEATURE_TESTS: ${{ matrix.options.tests && matrix.options.tests || 'ON' }}
+      FEATURE_TOOLS: ${{ matrix.options.tools }}
+      FEATURE_TOOLS_CTS: ${{ matrix.options.tools_cts }}
+      PACKAGE: ${{ matrix.options.package || matrix.options.package || 'NO' }}
+      DEPLOY_DOCS: ${{ matrix.options.deploy_docs || matrix.options.deploy_docs || 'NO' }}
```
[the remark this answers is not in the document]
MathiasMagnus: Removed.

.github/workflows/linux.yml
```diff
@@ -0,0 +1,329 @@
...
+            if [ "$WASM_BUILD" = "YES" ]; then
+              ./scripts/build_wasm_docker.sh
+            else
+              ./scripts/build_linux.sh
+            fi
+          fi
+          if [ "$CHECK_MKVK" = "ONLY" ]; then
+            scripts/check_mkvk.sh
+          fi
+          if [ "$DEPLOY_DOCS" = "YES" ]; then
```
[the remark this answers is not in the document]
MathiasMagnus: Removed.

.github/workflows/macos.yml
```diff
@@ -198,6 +198,8 @@ jobs:
           python-version: '3.11.4'
 
       - name: before_script
+        env:
+          MACOS_CERTIFICATES_P12: ${{ secrets.APPLE_CODE_SIGN_IDENTITY }}
```
MarkCallow, line 202: No. The secret is also called `MACOS_CERTIFICATES_P12`.

`APPLE_CODE_SIGN_IDENTITY`, used at signing time, identifies the certificate to use. This item is the actual certificates to be stored in the macOS Keychain.

You also need to set `MACOS_CERTIFICATES_PASSWORD` which is in secrets with the same name to enable `before_build_macos.sh` to import the certificates from MACOS_CERTIFICATES_P12.

.github/workflows/macos.yml
```diff
@@ -197,11 +197,13 @@ jobs:
         with:
           python-version: '3.11.4'
 
-      - name: before_script
+      - name: Add certificate to MacOS keychain
```
MarkCallow, line 200: certificates

.github/workflows/macos.yml
MathiasMagnus: Done.

.github/workflows/macos.yml
```diff
@@ -200,10 +201,13 @@ jobs:
       - name: Add certificate to MacOS keychain
         env:
           MACOS_CERTIFICATES_P12: ${{ secrets.MACOS_CERTIFICATES_P12 }}
+          ALTOOL_PASSWORD: ${{ secrets.ALTOOL_PASSWORD }}
```
MarkCallow, line 203: You need these two at notarize time too.

Actually the call to `notarize.sh` is passing ALTOOL_PASSWORD directly to it. This is because, in the distant past, when `notarize.sh` was being told to fetch the password from the keychain it kept hanging and I could not figure out why. Years later I figured out that the keychain unlock being done by `before_build_macos.sh` was timing out and macOS was displaying a GUI dialog to unlock the Keychain again. `before_build_macos.sh` now unlocks the Keychain without a timeout. This happened when I changed to using vcpkg because initial builds to build the vcpkg cache were taking so long. I forgot about the `notarize.sh` case.  We could change back to passing @keychain:$ALTOOL_PW_LABEL to `notarize.sh` so it retrieves the password from the Keychain.

Let's get it working first. If you do change it back, then only ALTOOL_PW_LABEL will be needed at notarize time.

.github/workflows/macos.yml
MathiasMagnus: Added.

.github/workflows/macos.yml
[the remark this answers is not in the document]
MathiasMagnus: Added step-local.

.github/workflows/macos.yml
MathiasMagnus: Certificates are now added correctly to the key chain.

.github/workflows/macos.yml
MarkCallow: s/MacOS/macOS/

.github/workflows/macos.yml
```diff
@@ -0,0 +1,270 @@
...
+          cat ~/.netrc
+          if [[ -n "$FEATURE_LOADTESTS" && "$FEATURE_LOADTESTS" != "OFF" ]]; then
+            # Install vcpkg.
+            pushd $HOME
+            git clone https://github.com/microsoft/vcpkg
+            cd vcpkg && ./bootstrap-vcpkg.sh -disableMetrics && echo "VCPKG_ROOT=`pwd`" >> $GITHUB_ENV
+            popd
+          fi
+
+      - name: Install Doxygen
```
MarkCallow, line 182: I noticed this and Install Graphviz are running in the iOS job. I think you need

```
if: matrix.options.doc == 'ON'
```

.github/workflows/linux.yml
```diff
@@ -0,0 +1,329 @@
...
+            # Some files in the pyktx docs have an _ prefix so Jekyll will
+            # not copy them from the gh-pages branch to the website. This
+            # file says no Jekyll files here. Treat all as ordinary files.
+            touch $BUILD_DIR/docs/html/.nojekyll
+          fi
+          set +e
+
+      - name: before_deploy
+        # Seemingly unneeded
```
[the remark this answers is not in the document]
MathiasMagnus: Noted.

.github/workflows/macos.yml
MathiasMagnus: Nice catch. Fixed.

MarkCallow:
Have you looked up the "error 65"? If not, see https://circleci.com/blog/xcodebuild-exit-code-65-what-it-is-and-how-to-solve-for-ios-and-macos-builds/ and do `man sysexits`.

I don't think the code signing has anything to do with this error. That happens long after `cmake` instantiates `xcodebuild` and, when secrets are available, nothing related to code signing is passed to xcodebuild on its command line.

```
EX_DATAERR (65)       The input data was incorrect in some way.  This should only be used for
                           user's data and not system files.
```

It also has nothing to do with the CMake warning you can see in the log. I have that too. It's been present since 3.31 and it isn't preventing my builds from working.

that said I don't have any idea what it is. I suggest removing the `| handle_compiler_output` to see if there helpful messages are being swallowed.

MarkCallow:
> I don't think the code signing has anything to do with this error.

Well I was wrong.

MathiasMagnus:
@MarkCallow I was going off on [this](https://github.com/actions/runner-images/issues/6642#issuecomment-1402377143) hunch.

MarkCallow:
Add `build.keychain` on the end of the command to find the installer cert. The command doesn't seem to use the default keychain. Even locally I had to add login.keychain before it found the certs.

I think GHA may be swallowing some of the output in the name of security. The developer cert was clearly found but no output from the command is visible. Locally I see
```
keychain: "/Users/mark/Library/Keychains/login.keychain-db"
version: 512
class: 0x80001000 
attributes:
    "alis"<blob>="Symantec Class 3 Organizational CA - G2"
    "cenc"<uint32>=0x00000003 
    "ctyp"<uint32>=0x00000001 
    "hpky"<blob>=0x75E1E3CEB21FAC8F96C5C6175C4AFDE1E01D3A06  "u\341\343\316\262\037\254\217\226\305\306\027\134J\375\341\340\035:\006"
  ...
```
 
Also add `security default-keychain` to verify the default is build.keychain. Try adding that in the script stage as well to make sure it is still the default.

MarkCallow:
It seems to have the Installer certificate under both names. I'll regenerate the .p12 file and make a new secret.

MarkCallow:
I have to go. I hope you can figure this out now. I have uploaded the APPLE_CODE_SIGN_IDENTITY and APPLE_PKG_SIGN_IDENTITY secrets again in case I made some error last time.

MarkCallow:
The visible parts of the certificates printed in the latest build match what I see locally so I think they've been added to the keychain correctly.

MarkCallow:
It's working now. There must have been an error in either MACOS_CERTIFICATES_P12 or APPLE_CODE_SIGN_IDENTITY before. I just uploaded both again. Sorry about that. It's an infernal nuisance that even the person who set the secrets can't view them.

MarkCallow:
Terrific. It has worked. It was even notarized correctly.

.github/workflows/macos.yml
MathiasMagnus: Changed.

.github/workflows/macos.yml
MathiasMagnus: Working, and most sensitive secrets are only visible in steps that need them.

.github/workflows/linux.yml
```diff
@@ -0,0 +1,329 @@
...
+        if: false
+        run: |
+          if [ ! -f $REL_DESC_FILE -a -f RELEASE_NOTES.md ]; then
+            # Remove lines that are unnecessary in the context of a GitHub
+            # release description.
+            awk '/^Release Notes/,/^## Version/ { next }
+                 ! /<!-- Copyright/ && ! /<!-- SPDX/ { print }' RELEASE_NOTES.md \
+            > $REL_DESC_FILE
+            # dpl v2 alternative
```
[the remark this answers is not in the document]
MathiasMagnus: Temp file is used. Fed to Release properly.

.github/workflows/linux.yml
```diff
@@ -0,0 +1,329 @@
...
+          os: ubuntu-22.04
+          generator: Ninja
+          arch: x86_64
+          compiler: clang
+          vk_sdk_ver: '1.3.290'
+          options: {
+            config: Release,
+            doc: OFF, jni: OFF, loadtests: OpenGL, py: OFF, tools: OFF, tools_cts: OFF,
+            package: YES, sse: OFF, opencl: OFF, wasm: YES
+          }
```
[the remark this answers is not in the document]
MathiasMagnus: If that is known to cause warnings, that matrix elem should be added in #1012 to prove that it doesn't trip the broadened matrix. (I already subsumed #1010 by rebasing on top of that, but I don't want start cherry picking other branches.)

.github/workflows/linux.yml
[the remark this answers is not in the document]
MarkCallow: You are only creating $REL_DESC_FILE once but it is being uploaded every time a package is deployed. There are two packages, x86_64 and aarch64, unlike before. As I can't see a way to set body_path for only one deployment let's accept that it is done twice.

MarkCallow:
One final thing. Move check reuse into its own workflow, which does not need to be callable from `manual.yml`. This workflow will need to run no matter what file in the repo has changed so no `ignore:`.

Please delete the tag you created and the draft release.

.github/workflows/linux.yml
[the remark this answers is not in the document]
MarkCallow: Good idea.

.github/workflows/linux.yml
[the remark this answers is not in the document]
MathiasMagnus: Removed.

.github/workflows/linux.yml
```diff
@@ -215,13 +199,7 @@ jobs:
         run: |
           set -e # Instead of trying to chain everything together with &&
```
MarkCallow, line 200: This and `set +e` are not needed any more as there is now a single command. Please remove.

MarkCallow (changes requested):
Look good. Like the simplification. One very minor change please.

aqnuep (approved):

.github/workflows/web.yml
```diff
@@ -0,0 +1,197 @@
...
+          if [ "$FEATURE_TOOLS_CTS" = "ON" ]; then
+            git submodule update --init --recursive tests/cts
+          fi
+          # Make sure embedded dates are correct.
+          ./install-gitconfig.sh
+          scripts/smudge_date.sh
+
+      - name: script
+        run: |
+          set -e # Instead of trying to chain everything together with &&
```
MarkCallow, line 157: Remove this set +-e pair too.

.github/workflows/web.yml
```diff
@@ -0,0 +1,197 @@
...
+        - name: Package
+          os: ubuntu-22.04
+          generator: Ninja
+          arch: x86_64
+          compiler: clang
+          vk_sdk_ver: '1.3.290'
+          options: {
+            config: Release,
+            doc: OFF, jni: OFF, loadtests: OpenGL, py: OFF, tools: OFF, tools_cts: OFF,
+            package: YES, sse: OFF, opencl: OFF, wasm: YES
```
MarkCallow, line 70: s/wasm: YES/wasm:ON/ to match the Debug case.

MarkCallow (changes requested):
Two more minor changes. @aqnuep's additional review request took me a view of all the changes. I was previously looking at just those made since I last viewed.

.github/workflows/web.yml
MathiasMagnus: Removed.

.github/workflows/web.yml
MathiasMagnus: Changed.

MarkCallow:
Thanks for the excellent work, @MathiasMagnus.

PR #1019, opened 2025-04-17
Vcpkg cache artifacts on macOS

### LINKED ISSUE #1013: macos workflow not caching vcpkg binaries
In .travis.yml vcpkg binaries were being cached between jobs/builds via

```
cache:
  directories:
  -  $HOME/.cache/vcpkg/archives
```

The new `macos.yml` workflow has no equivalent. I missed this during reviews. Sorry. In `windows.yml` this is accomplished by

```
    - name: Export GitHub Actions cache and vcpkg environment variables
      if: matrix.options.loadtests != 'OFF'
      uses: actions/github-script@v7
      with:
        # Caution. JSON in use. "#core" breaks the build.
        script: |
          core.exportVariable('ACTIONS_CACHE_URL', process.env.ACTIONS_CACHE_URL || '');
          core.exportVariable('ACTIONS_RUNTIME_TOKEN', process.env.ACTIONS_RUNTIME_TOKEN || '');
          core.exportVariable('VCPKG_BINARY_SOURCES', "clear;x-gha,readwrite");
          core.exportVariable('VCPKG_ROOT', process.env.VCPKG_INSTALLATION_ROOT || '');
```

I think this should work on macOS too except that the last line is unnecessary as the workflow is already exporting VCPKG_ROOT. The line here is because vcpkg is already installed in Windows runners but only VCPKG_INSTALLATION_ROOT is set.

MathiasMagnus:
Fixes #1013

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -188,6 +188,16 @@ jobs:
             popd
           fi
 
+      - name: Export GitHub Actions cache and vcpkg environment variables
+        if: matrix.options.loadtests != 'OFF'
+        uses: actions/github-script@v7
+        with:
+          # Caution. JSON in use. "#core" breaks the build.
+          script: |
+            core.exportVariable('ACTIONS_CACHE_URL', process.env.ACTIONS_CACHE_URL || '');
+            core.exportVariable('ACTIONS_RUNTIME_TOKEN', process.env.ACTIONS_RUNTIME_TOKEN || '');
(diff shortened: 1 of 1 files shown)


MarkCallow (approved):
Looks good. Very difficult to test. I re-ran one of the jobs but it built all the vcpkg packages again although messages in the log prior to the re-run indicated that the packages had been saved to the cache.

I will merge this and check future builds to see if it is working or not.

Thanks for the fix.

PR #1037, opened 2025-06-24
Update for Vulkan SDK 1.4.313.

MarkCallow:
* Update SDK version number in workflow.
* Change install_macos.sh to handle zip instead of dmg.
* Update loadtests to stop MoltenVK warning.
* Move ARM64 package build to windows-11-arm runner and run tests there while keeping a simpler cross-compile build as a check.
* Set GIT_LFS_SKIP_SMUDGE everywhere to save more bandwidth.
* Put `git lfs pull` commands in workflow files for clarity.
* Remove .netrc setup with GITHUB_TOKEN. Not needed on Actions runners.

### DIFF
--- a/.github/workflows/check-reuse.yml
+++ b/.github/workflows/check-reuse.yml
@@ -24,6 +24,8 @@ on:
 jobs:
   check-reuse:
     runs-on: ubuntu-latest
+    env:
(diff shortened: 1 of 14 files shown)


MarkCallow:
Closing. Will reopen once the Linux and Windows CI builds are updated.

PR #1041, opened 2025-07-09
fix: Enabled Android x86 build.

atteneder:
After upgrading to 4.4.0 building Android x86 failed.

As far as I can x86 was removed in a545f005ff987aa5a4a3c7f938490954ca85c095 so that the load tests build succeeds (presumably some platform like Windows x86 failed). However, libktx itself still builds nicely on Android x86 (and maybe even more x86 platforms).

This PR is not driven by known user demand (heck, I don't know if anyone has use for this), but to maintain compatibility in my downstream product. Removing support for a platform is a breaking change and requires a major version bump, strictly speaking. That's something I want to avoid.

Let me know what you think. Thanks for considering.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -209,7 +209,7 @@ include(GNUInstallDirs) # Must be after project.
 include(CTest)          #           "
 
 set_target_processor_type(CPU_ARCHITECTURE) # Must be after project.
-if (CPU_ARCHITECTURE STREQUAL x86)
+if (CPU_ARCHITECTURE STREQUAL x86 AND NOT ANDROID)
     message(FATAL_ERROR "This project cannot be built for x86 cpu.")
 endif()
 CMAKE_DEPENDENT_OPTION( BASISU_SUPPORT_SSE


MarkCallow:
Please rebase this or PR #1043 against the `fix_pr_builds` branch and force push. That branch has a tentative fix to enable PR builds and retrieve vcpkg cached packages. I do not want to merge it to main until I know it works and I have no other way of testing this than to make thisr equest. When I submit PRs they get the repo secrets.

Once that is sorted I'll have some time to look at the PRs themselves.

MarkCallow:
Partial success. On the macOS and Linux builds some packages were retrieved from the cache. Not clear why only some. On Windows the alternate value for API_TOKEN was not set correctly so it got the same error that "both user name and password are needed" but it did not cause the build to fail. Not clear why there was no failure.

I have pushed a change for Windows which hopefully will set the alternate token value correctly. Please try again.

MarkCallow:
Please merge the latest change from `fix_pr_builds` I need confirmation that it does indeed work before I can merge that branch to `main`.

MarkCallow:
Ping @atteneder please. The lack of final confirmation of the fix is holding me up.

atteneder:
just merged that. sorry for the delay.

MarkCallow:
That is working perfectly with packages retrieved from the cache. I did have a scare thought. The job on the Windows arm64 runner did not retrieve anything. But it was a false alarm. The reason no packages were restored is because the runner has been updated since the last time packages were cached.

As for this PR, since disabled x86 builds I have not been making any effort to ensure code involving all the various 64-bit quantities, such as offsets in the KTX file, will compile and work on an x86 so I cannot make any guarantee that the code will run correctly, even if it compiles. And I don't even know if it compiles. Our CI only builds for arm64 and x86_64.

Have you compiled and run the code @atteneder?

MarkCallow:
> I have compiled successfully for Android x86 (afaik only some Chromebooks have that), but not actually ran the tests. I might be able to run my test app (which does KTX loading/basisU transcoding only) in an emulator.

Please run the tests. I don't think there is any way to run the library tests (unittests, etc) on Android. We really should have a way to test the x86 build so it doesn't get broken by future changes. On the other hand trying to ensure it works is likely to be a burden on future updates. If it is really only some old Chromebooks that have it, it doesn't seem worth the effort.

atteneder:
A little update on my end:
I have not found a way to emulate and test said Android x86 build. I only need to support this to the mid of next year at max (Unity has dropped support for Android x86 on ChromeOS with version 6.0), so I might as well drag along this small fix in a fork until then.

Feel free to close the issue.

PR #1043, opened 2025-07-10
fix: Restored ability to build universal macOS binaries.

atteneder:
I'm maintaining a universal macOS bundle that consumes `ktx_read`. After upgrading to 4.4.0 I was not able to build anymore, as none of the astc-encoder static targets was added as dependency.

Now in case of universal builds (so `arm64` and `x86_64`), multiple astc-encoder targets may get added as dependency (in most cases `astcenc-sse4.1-static` and `astcenc-neon-static`).

bonus fix: I have reason to believe `ktx` was accidentally merged into/overwrote `ktx_read` when combining all static libraries into one when building an Apple static library (in [this line](https://github.com/KhronosGroup/KTX-Software/blob/f9f36940b5f807b3d469e1eab4ad19f76e5e9c70/CMakeLists.txt#L1281), note the hard-coded reference to `ktx` in `$<TARGET_FILE:ktx>`). The fix is in [this new line](https://github.com/atteneder/KTX-Software/blob/77a0a3a5ea9f1464db15bc754b18173faa94a959/CMakeLists.txt#L1300). I presume it never surfaced since `ktx` is a superset of `ktx_read`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,6 +11,7 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/cmake/modules/")
 find_package(Bash REQUIRED)
 
 include(cmake/version.cmake)
+include(cmake/macros.cmake)
 
 if(POLICY CMP0149)
     # Ignore CMAKE_SYSTEM_VERSION and select either latest available
@@ -648,7 +649,7 @@ macro(common_libktx_settings target enable_write library_type)
             # To avoid this set a define to prevent the compiler using
             # constexpr mutex constructors. Remove this eventually after
             # in-use JVM installations have at least this VC runtime. Remove
-            # also from ASTCENC_LIB_TARGET settings around line 1169.
+            # also from ASTCENC_LIB_TARGETS settings around line 1169.
             $<$<AND:${is_msvccl},$<VERSION_GREATER_EQUAL:$<CXX_COMPILER_VERSION>,19.40.33811>>:_DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR>
             $<$<AND:${is_clangcl},$<VERSION_GREATER_EQUAL:$<CXX_COMPILER_VERSION>,17.0.3>>:_DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR>
         PUBLIC # only for basisu_c_binding.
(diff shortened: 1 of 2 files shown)


MarkCallow:
I'll look at this in more detail later. First glance, it is not handling universal build for the Basis encoders or at least doing so without using SIMD on x64. The issue with these is that they use cpp defines to control whether they use SIMD (and only SSE is supported). Meanwhile the standard Xcode way of building a universal binary is to run the compiler front end once then run code generation for each selected architecture. Therefore only one set of cpp defines can be used. I believe the underlying tools support compiling 2 separate binaries and then combining them in the library archive but I don't know how to get Xcode to do that. That is why I disabled universal builds and make two separate release packages. I'd love to have it fixed.

Regarding the CI failure, this is the first PR since I updated vcpkg caching to using NuGet. The username and password for the cache are repository secrets so a Pr does not see them. I need to do some research to see if there is a way for PRs to use the vcpkg package cache.

MarkCallow:
If you wish to proceed with this please fix the build failures. I am not willing to build universal binaries that have SIMD code disabled for x86-64 in the Basis encoders unless you are willing to also fix that we'd better drop this PR.

Note that I expect to pull a big update to Basis encoders soon. I do not expect any change to the way SIMD is enabled (i.e. still via ifdefs) but I expect fairly big changes so any work in the current encoders is likely to be wasted. If you want to do this, it would be best to change the build to do 2 separate compiles which are combined into the final archive. However given the rapid disappearance of x86_64 Macs I am uneasy about forcing on everyone, and the planet, the burden of downloading fat binaries.

CMakeLists.txt
```diff
@@ -1261,32 +1274,40 @@ set(ASTCENC_CLI OFF) # Only build as library not the CLI astcencoder
...
         # Make a single static library to simplify linking.
-        add_dependencies(${target} ${ASTCENC_LIB_TARGET})
+        foreach(ASTCENC_LIB_TARGET ${ASTCENC_LIB_TARGETS})
+            add_dependencies(${target} ${ASTCENC_LIB_TARGET})
+            list(APPEND TARGETS "$<TARGET_FILE:${ASTCENC_LIB_TARGET}>")
+        endforeach()
         add_custom_command( TARGET ${target}
             POST_BUILD
-            COMMAND libtool -static -o $<TARGET_FILE:${target}> $<TARGET_FILE:ktx> $<TARGET_FILE:${ASTCENC_LIB_TARGET}>
+            COMMAND libtool -static -o $<TARGET_FILE:${target}> $<TARGET_FILE:${target}> ${TARGETS}
```
dg0yt, line 1300: FTR I also arrived at the `ktx` -> `${target}` substitution while updating the vcpkg port. It fixes the "duplicate member name" warnings in the output:
~~~
[130/148] : [...] libtool -static -o /Users/vcpkg/Data/b/ktx/x64-osx-rel/libktx_read.a /Users/vcpkg/Data/b/ktx/x64-osx-rel/libktx.a /Users/vcpkg/Data/b/ktx/x64-osx-rel/external/astc-encoder/Source/libastcenc-avx2-static.a
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/libtool: warning for architecture: x86_64h duplicate member name 'astcenc_averages_and_directions.cpp.o' from '/Users/vcpkg/Data/b/ktx/x64-osx-rel/external/astc-encoder/Source/libastcenc-avx2-static.a(astcenc_averages_and_directions.cpp.o)' and '/Users/vcpkg/Data/b/ktx/x64-osx-rel/libktx.a(astcenc_averages_and_directions.cpp.o)'
~~~

Context: I looked at x86_64 macOS build failures while updating the vcpkg port (not universal binaries). The build failures turned out to be from `libastcenc-avx2-static.a` being x86_64h which doesn't seem to be considered when the linker later wants to link x86_64 executables.

MarkCallow: Thank you for all your work on the vcpkg port. Please make a PR to fix this in `lib/CMakeLists.txt`, where it now resides, unless you plan to wrap the fix up in a larger PR for vcpkg-related items.

Do you have any code that does the equivalent of libtool for Linux and Windows?

PR #1049, opened 2025-08-01
ktx create: add premultiply alpha option

jiangzhhhh:


tools/ktx/command_create.cpp
```diff
@@ -2475,6 +2489,9 @@ KTXTexture2 CommandCreate::createTexture(const ImageSpec& target) {
 
     KHR_DFDSETVAL(texture->pDfd + 1, PRIMARIES, target.format().primaries());
     KHR_DFDSETVAL(texture->pDfd + 1, TRANSFER, target.format().transfer());
+    if(options.premultiplyAlpha) {
```
MarkCallow, line 2511: You need to check it was actually pre-multiplied. `premultipleAlpha` rightly does nothing if there aren't 4 components.

tools/ktx/command_create.cpp
```diff
@@ -965,6 +973,9 @@ Create a KTX2 file from various input files.
             normals to generate X+Y normals with @b --normal-mode. For 4-component
             inputs a 3D unit normal is calculated. 1.0 is used for the value of
             the 4th component. Cannot be used with @b \--raw.</dd>
+        <dt>\--premultiply-alpha</dt>
+        <dd>Pre-multiplies the texture by its alpha value before encoding
```
MarkCallow: "Pre-multiplies the color components of the input pixels by the alpha component before encoding and sets the flag in the metadata. Only has an effect for 4-component inputs. Cannot be used with @b --normalize or @b --raw."

Update the option processing to disallow `--normalize` or `--raw` with `--premultiply-alpha`.

tools/imageio/image.hpp
```diff
@@ -1314,6 +1315,24 @@ class ImageT : public Image {
...
 
+    virtual ImageT& premultiplyAlpha() override {
+        if(getComponentCount() < 4) {
+            return *this; // Nothing to do (no alpha channel)
+        }
+
+        uint32_t pixelCount = getPixelCount();
+        static_assert(gc_s[sizeof(componentType)] > 0);
+        double invMax = 1.0 / gc_s[sizeof(componentType)];
+        for (uint32_t i = 0; i < pixelCount; ++i) {
```
MarkCallow, line 1330: If the input is sRGB you need to decode to linear, do the pre-multiplication then reencode.

tools/ktx/command_create.cpp
```diff
@@ -175,6 +177,8 @@ struct OptionsCreate {
                     "normals to generate X+Y normals with --normal-mode. For 4-component\n"
                     "inputs a 3D unit normal is calculated. 1.0 is used for the value of\n"
                     "the 4th component. Cannot be used with --raw.")
+                (kPremultiplyAlpha, "Pre-multiplies the texture by its alpha value before encoding"
```
MarkCallow:
```
"Pre-multiplies the color components of the input pixels by the alpha component\n"
before encoding and sets the flag in the metadata. Cannot be used with --normalize\n"
or --raw."
```

MarkCallow (changes requested):
That was fast! Thanks.

The most important thing needing fixing is properly handling sRGB images and adding test cases to the tools CTS (https://github.com/KhronosGroup/KTX-Software-CTS). You will need to submit the new tests as a PR to the CTS repo and update the submodule ref in the main repo.

Please update vkloadtests and, if time, glloadtests, to properly display pre-multipled images and provide one or more test images for them to load. The most important one to update is`tests/loadtests/Texture.cpp`.

tools/imageio/image.hpp
jiangzhhhh: Execuse me, Could you tell me how to detect the image object is sRGB or linear? It seems there are no propertis to descible within it.
Do I need to add one?

jiangzhhhh: Is it correct to understand that the color space of the data in an image is specified by the user through the transfer function?

MarkCallow: The color space is set from the input image metadata possibly modified by the user via the transfer function options. In `executeCreate`, from which you are calling `image->premultiplyAlpha`, the target transfer function is given by `target.format().transfer()`. The PR is correctly calling `image->premultiplyAlpha` after any color transformation has been performed thus the transfer function will have been set in `image`. You can therefore call the `getTransferFunction` method of the `Image` class in `premultiplyAlpha`.

Look at `ImageT::resample` for an example of how to decode and encode while processing the pixels of the image using the `TransferFunctionSRGB` and `TransferFunctionLinear` classes that are also defined in `image.hpp`.

jiangzhhhh:
@MarkCallow I'm still working on it. Once I finish the revisions, I'll reopen this PR.

jiangzhhhh:
image
I noticed that there is a path in toktx for assigning values to the transfer function of the image object, but it is not present in ktx create. Is it missing in ktx create?

MarkCallow:
> I used a modified ktx create to convert a PNG to a premultiplied alpha RGBA KTX2 image, but the result in the vulkanloadtests renders incorrect (I compared it with another straight alpha RGBA KTX2 image).

vulkanloadtests (actually `vkloadtests`) does not have support for pre-multiplied-alpha images. It does not look at the bit in the DFD.

In what way is the rendering incorrect?

> I suspect the issue is that vulkanloadtests uses a linear workflow, blending semi-transparent images in linear space. However, when I converted the PNG to premultiplied RGBA, ktx create didn’t apply an sRGB-to-linear conversion—meaning the premultiplication was done in sRGB space, causing a mismatch in rendering.

`vkloadtests` and `ktxTexture_VkUploadEx` which it uses` correctly pass the format of the input file to Vulkan, either `_UNORM` or `SRGB`. After that it is up to Vulkan. It will decode sRGB textures to linear when sampling them. Outputs of the fragment shader are always linear. They are encoded to sRGB, if the framebuffer/swapchain is sRGB, which is the case in `vkloadtests`.  The linear outputs of the fragment shader, together with values from the framebuffer, decoded to linear if the buffer is sRGB, are fed to blending and the result converted back to sRGB.

If there is an issue it is because the texture's format is misidentified.

> The problem is, I don’t know what options to pass to ktx create's assign-tf/convert-tf to ensure the image converts to linear space before premultiplication.

See [here](https://github.khronos.org/KTX-Software/ktxtools/ktx_create.html#ktx_create_tf_handling) for an explanation of transfer function handling in `ktx create`.

The line below will be called if a color transform is necessary.
https://github.com/KhronosGroup/KTX-Software/blob/d47320c24e2e20b0abf6f1c9a146fc84fe16011c/tools/ktx/command_create.cpp#L1741-L1742

A transform may not be necessary depending on the input format. You can also see what the intended color space is by looking at the value of `target.format().transfer()`

MarkCallow:
> I noticed that there is a path in toktx for assigning values to the transfer function of the image object, but it is not present in ktx create. Is it missing?

It does indeed appear to be missing which means that resampling has not been working as intended. I will open another bug about this. I think the fix should be made separately from this work. Thanks for noticing this.

MarkCallow:
> However, vlloadtest loads images using sRGB texture format, which converts to linear space when sampled.

No it does not. It loads the images according to the vkFormat field of the file it is loading.

> This means alpha blending should happen in linear space, not sRGB space, which clashes with the premultiply operation.

What do you mean it "clashes?" Data in the fragment shader is always linear unless the texture being sampled is incorrectly labelled as linear when it really sRGB. Alpha blending, regardless of pre-multiply is thus always (and must be) done in linear. As I said earlier fragment shader outputs are then encoded back to sRGB, if the framebuffer is sRGB. This is how graphics APIs work. It has nothing to do with `vkloadtests`.

Oh! Is the clash you refer to between the ktx create bug causing the pre-multiply to be done in sRGB space while the unpremultiply, or whatever, is being done in linear?

I opened #1050 about the failure to set the color space into `image`.

jiangzhhhh:
According to my understanding.
On devices that don't support sRGB textures (like GLES 2.0), sRGB images are created as regular textures, and no color space conversion happens during sampling. In this case, alpha blending actually takes place in sRGB space (since the source data stored in the image is sRGB).

For this scenario, premultiplying these images before rendering must also be done in sRGB space.

In short, Whether to premultiply alpha in color space depends on how the pipeline is set up.

tests/loadtests/vkloadtests.cmake
```diff
@@ -330,12 +332,14 @@ if(APPLE)
                 ktx
             )
             add_custom_command( TARGET vkloadtests POST_BUILD
-                COMMAND ${CMAKE_COMMAND} -E create_symlink $<TARGET_FILE_NAME:ktx> "$<TARGET_BUNDLE_CONTENT_DIR:vkloadtests>/Frameworks/$<TARGET_SONAME_FILE_NAME:ktx>"
-                COMMENT "Create symlink for KTX library (ld name to real name"
+                COMMAND ${CMAKE_COMMAND} -E make_directory "$<TARGET_BUNDLE_CONTENT_DIR:vkloadtests>/Frameworks"
```
MarkCallow: Why did you add this? It has been working fine without it.

tests/loadtests/vkloadtests.cmake
```diff
@@ -330,12 +332,14 @@ if(APPLE)
...
+                COMMAND ${CMAKE_COMMAND} -E create_symlink $<TARGET_FILE:ktx> "$<TARGET_BUNDLE_CONTENT_DIR:vkloadtests>/Frameworks/$<TARGET_SONAME_FILE_NAME:ktx>"
+                COMMENT "Create symlink for KTX library (ld name to real name)"
             )
         endif()
         add_custom_command( TARGET vkloadtests POST_BUILD
-            COMMAND ${CMAKE_COMMAND} -E create_symlink "${Vulkan_LIBRARY_REAL_FILE_NAME}" "$<TARGET_BUNDLE_CONTENT_DIR:vkloadtests>/Frameworks/${Vulkan_LIBRARY_SONAME_FILE_NAME}"
+            COMMAND ${CMAKE_COMMAND} -E make_directory "$<TARGET_BUNDLE_CONTENT_DIR:vkloadtests>/Frameworks"
```
MarkCallow: And why did you add this?

tests/loadtests/vkloadtests.cmake
```diff
@@ -19,13 +19,13 @@ if(APPLE)
...
-                        OUTPUT_VARIABLE Vulkan_LIBRARY_REAL_PATH_NAME
+                        OUTPUT_VARIABLE Vulkan_LIBRARY_REAL_PATH
                         OUTPUT_STRIP_TRAILING_WHITESPACE
         )
         cmake_path(GET
-                  Vulkan_LIBRARY_REAL_PATH_NAME
+                  Vulkan_LIBRARY_REAL_PATH
                   FILENAME
-                  Vulkan_LIBRARY_REAL_FILE_NAME
+                  Vulkan_LIBRARY_REAL_PATH_NAME
```
MarkCallow, line 26: This is not correct. It is supposed to be getting the file name of vulkan library here, that is that part of the path after the last directory separator. Revert to `Vulkan_LIBRARY_REAL_FILE_NAME`.

tests/loadtests/vkloadtests.cmake
```diff
@@ -19,13 +19,13 @@ if(APPLE)
         # Find the name of the actual dylib which includes the version no.
         # readlink -f requires macOS >= 12.3!
         execute_process(COMMAND readlink -f ${Vulkan_LIBRARIES}
-                        OUTPUT_VARIABLE Vulkan_LIBRARY_REAL_PATH_NAME
+                        OUTPUT_VARIABLE Vulkan_LIBRARY_REAL_PATH
```
MarkCallow: Why did you remove the `_NAME` suffix?

tests/loadtests/vkloadtests/Texture.cpp
```diff
@@ -311,6 +315,7 @@ Texture::cleanup()
 
     quad.freeResources(vkctx.device);
     uniformDataVS.freeResources(vkctx.device);
+    uniformDataAlphaMode.freeResources(vkctx.device);
```
MarkCallow: I would prefer uniformDataFS with an alphaMode flag inside it. It is then immediately obvious why there is a second uniform buffer.

tools/ktx/command_create.cpp
```diff
@@ -1802,6 +1823,15 @@ void CommandCreate::executeCreate() {
             if (options.swizzleInput)
                 image->swizzle(*options.swizzleInput);
 
+            if (options.premultiplyAlpha) {
+                if(image->getComponentCount() < 4) {
+                    // This branch is unreachable because all images from loadInputImage already contain alpha channel.
```
MarkCallow: I think you mean "none of the images contain an alpha channel".  But I don't think the comment adds anything you can't tell from reading the error message. I suggest deleting it.

tools/ktx/command_create.cpp
```diff
@@ -1847,6 +1877,12 @@ void CommandCreate::executeCreate() {
 
     compress(texture, options);
 
+    // Set premultiplied alpha metadata.
+    // Must be done after encoding and compression because those operations may change the texture object.
```
MarkCallow: Do you know that `compress` loses the pre-multiplied flag if set before it is called? If it does, that is a bug in compress that should be fixed.

MarkCallow (changes requested):
Thanks for the ongoing work. It looks like it is shaping up well.

I have created a PR to fix the color space issue. I'll give the person I asked to review it more time before I merge it.

tools/ktx/command_create.cpp
```diff
@@ -175,6 +177,9 @@ struct OptionsCreate {
                     "normals to generate X+Y normals with --normal-mode. For 4-component\n"
                     "inputs a 3D unit normal is calculated. 1.0 is used for the value of\n"
                     "the 4th component. Cannot be used with --raw.")
+                (kPremultiplyAlpha, "Pre-multiplies the color components of the input pixels by the alpha component"
```
MarkCallow, line 181: "\n" needed at the end of this line.

tools/ktx/command_create.cpp
jiangzhhhh: In the compression processes of ASTC, ETC1S, and UASTC, a new KTX object is created to replace the original one, which causes the flags of the original KTX object to be lost.

This could also be handled separately in each of the three compression functions, but I think placing it here works just as well.

tools/ktx/command_create.cpp
jiangzhhhh: image
I feel like manually adding line breaks for formatting messed up the help output's layout.

MarkCallow: I see. Then insert a space before "before" otherwise the output has "componentbefore".

tools/ktx/command_create.cpp
MarkCallow: Please fix it in the compress functions. They can be called from `encode` as well as from `create`. A user could pass a premultiplied alpha .ktx2 file to `encode`.

jiangzhhhh: ok

jiangzhhhh:
I have resolved it.

Mark Callow ***@***.***> 于2025年8月23日周六 09:08写道：

tools/imageio/image.hpp
```diff
@@ -1314,6 +1315,32 @@ class ImageT : public Image {
...
+
+        uint32_t pixelCount = getPixelCount();
+        for (uint32_t i = 0; i < pixelCount; ++i) {
+            Color& c = pixels[i];
+            float alpha = c[3] * Color::rcpOne();
+            for (uint32_t comp = 0; comp < 3; comp++) {
+                float linearColor = tf.decode(c[comp] * Color::rcpOne());
+                linearColor *= alpha;
+                float srgbColor = tf.encode(linearColor);
+                float outValue = static_cast<componentType>(srgbColor * static_cast<float>(Color::one()) + 0.5f);
```
MarkCallow: What is the multiply by (Color::one() + 0.5 for)? And why the + 0.5?

jiangzhhhh:
I originally thought about referring the code of resample to resolve the
compilation errors on MSVC, but it still reports errors.
Later, I will switch to Windows to fix it again.



Mark Callow ***@***.***> 于2025年8月25日周一 20:34写道：

jiangzhhhh:
ok, I will take some time to finish it. but it doesn't matter if miss the
release.

Mark Callow ***@***.***> 于2025年9月5日周五 16:23写道：

MarkCallow:
I am on the cusp of creating v4.4.1. Please merge main today then, if CI passes, I can merge this PR and include it in the release.

PR #1051, opened 2025-08-08
Set color space in input image prior to resample.

### LINKED ISSUE #1050: Resampling, when called from ktx create, not decoding sRGB images
The cause is that _ktx create_ never sets the color space information into the input `image` but the resampler uses the `transferFunction` value to determine decoding as shown here.

https://github.com/KhronosGroup/KTX-Software/blob/d47320c24e2e20b0abf6f1c9a146fc84fe16011c/tools/imageio/image.hpp#L1150-L1154

In _ktx create_ after these lines

https://github.com/KhronosGroup/KTX-Software/blob/d47320c24e2e20b0abf6f1c9a146fc84fe16011c/tools/ktx/command_create.cpp#L1711-L1744

something like

```
image->setPrimaries(target.format().primaries());
image->setTransferFunction(target.format().transfer());
```
is needed.

All the tests involving resampling and sRGB will have to be regenerated.

Thanks to @jiangzhhhh for catching this.

MarkCallow:
Fixes #1050.

### DIFF
--- a/tools/ktx/command_create.cpp
+++ b/tools/ktx/command_create.cpp
@@ -1742,6 +1742,10 @@ void CommandCreate::executeCreate() {
                     image->transformColorSpace(*colorSpaceInfo.src.transferFunction, *colorSpaceInfo.dst.transferFunction);
                 }
             }
+            // Must be set before operations like resampling. Needed even when
+            // transformColorSpace is not called.
+            image->setPrimaries(target.format().primaries());
+            image->setTransferFunction(target.format().transfer());
 
             // TODO: Add auto conversion and warning? Not needed now
             // because all supported source formats provide top-left images.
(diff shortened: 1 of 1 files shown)


MarkCallow:
> @MarkCallow, we did not implement scaling, so I have no idea whether this is the right order of operations, therefore I'm not sure I can provide an informed review of this without understanding the details how scaling was added.

Thanks for looking. `--generate-mipmap` is also affected by this bug as both it and `--scale` use the resample method to resize images. Not having the color space set before `scaleImage` and `generateMipLevels` are called results in `Image::resample` always seeing `_UNSPECIFIED` and so never decoding sRGB images. The color space must be set before these functions are called. I am confident this is the correct order so I am going to merge this.

I am not too surprised the issue was missed when the `--generate-mipmap` tests were added as the differences between correct and incorrect images are very subtle.

PR #1072, opened 2025-10-29
Add `KTX_FEATURE_JS` option controlling whether `ktx_js[_read]` are built.

kring:
Adds a new CMake option called `KTX_FEATURE_JS`. It defaults to ON, but when set to OFF, `ktx_js` and `ktx_js_read` are not built, even under Emscripten.

My use-case for this is compiling [Cesium for Unity](https://github.com/CesiumGS/cesium-unity) for the web. Compiling Unity applications to WebAssembly requires using Unity's version of Emscripten, which is quite old (v3.1.39), and is not able to compile `ktx_js` successfully. This isn't really worth fixing because the JavaScript bindings aren't used in this context anyway. KTX functionality is only called from Emscripten'd native code, never from regular JS code.

So with this low-impact PR, it's easy to disable this unnecessary part of the KTX software suite in order to get a successful build.

The slightly bigger picture here is to then expose this via a vcpkg feature flag.

This PR builds on #1071, because it looks like that PR is going to be merged quickly and I didn't want to create any unnecessary conflicts.

PR #1073, opened 2025-10-30
Add KTX_FEATURE_JS option controlling whether `ktx_js[_read]` are built.

kring:
(Reopening #1072, now targeting `main`)

Adds a new CMake option called `KTX_FEATURE_JS`. It defaults to ON, but when set to OFF, `ktx_js` and `ktx_js_read` are not built, even under Emscripten.

My use-case for this is compiling [Cesium for Unity](https://github.com/CesiumGS/cesium-unity) for the web. Compiling Unity applications to WebAssembly requires using Unity's version of Emscripten, which is quite old (v3.1.39), and is not able to compile `ktx_js` successfully. This isn't really worth fixing because the JavaScript bindings aren't used in this context anyway. KTX functionality is only called from Emscripten'd native code, never from regular JS code.

So with this low-impact PR, it's easy to disable this unnecessary part of the KTX software suite in order to get a successful build.

The slightly bigger picture here is to then expose this via a vcpkg feature flag.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,6 +44,7 @@ option( KTX_FEATURE_JNI "Create Java bindings for libktx." OFF )
 option( KTX_FEATURE_PY "Create Python source distribution." OFF )
 option( KTX_FEATURE_TESTS "Create unit tests." ON )
 option( KTX_FEATURE_TOOLS_CTS "Enable KTX CLI Tools CTS tests (requires CTS submodule)." OFF )
+option( KTX_FEATURE_JS "Enable JavaScript bindings in Emscripten builds." ON )
 
 if(KTX_FEATURE_TOOLS_CTS AND NOT KTX_FEATURE_TOOLS)
     message(WARNING "KTX_FEATURE_TOOLS is not set -> disabling KTX_FEATURE_TOOLS_CTS.")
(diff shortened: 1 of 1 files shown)


kring:
> I am happy to add this for your use case.

Thank you!

> I suggest you look at changing Unity's build to use the Emscripten Docker image though.

Unfortunately this is something that the Unity game engine / company has to do. Rumor has it they're working on it. Cesium for Unity (the thing my team and I are responsible for) is happy to use newer Emscripten versions.

> I can't promise that the compiling the native interfaces to wasm won't at some point need a more recent version of Emscripten.

Understood, though fortunately that's much less likely to be a near-term issue than the JS interface layer, where things change pretty quickly.

> As for the idea of exposing this via vcpkg feature flag my question is does vcpkg handle dependencies for JavaScript apps?

Generally, no. It's a C/C++ package manager. But then C/C++ code can of course be used to developer JavaScript applications, right? So I guess the scenario where someone would want to turn ON the vcpkg feature to expose KTX-Software's JS interfaces is if it's part of a larger application, but they want the KTX-Software interfaces to be directly accessible from JS as well. Unlikely? Perhaps. But the JS interfaces have been available in the vcpkg package so far, so I'd be hesitant to remove them entirely. I just want to provide a way to turn them OFF when they're unnecessary or cause problems (such as in this old Emscripten version we're stuck with).

> Also, if you just want to build the library, you can now do so with lib/CMakeLists.txt. We have CI checks to ensure it works, and keeps working, as a standalone project.

That's a great improvement! In the vcpkg context, though, I think that utilizing it would require separating KTX-Software into multiple packages. One for the lib, and one or more others for tools, JS interfaces, etc. It's a bigger change, but could be worth considering...

PR #1075, opened 2025-10-30
Upstream changes from vcpkg

kring:
This PR also incorporates the changes from #1073, so please merge that first.

The [ktx vcpkg port](https://github.com/microsoft/vcpkg/tree/master/ports/ktx) incorporates six patches to the KTX-Software build system. This PR incorporates them all into this repository, aiming to minimize the impact on anyone using this repo directly. If these changes are acceptable to the maintainers, having them "upstreamed" will make it much easier to incorporate new versions of ktx into vcpkg.

I'm not the original author of these patches, but I'll explain my understanding the purpose of the changes here and what I've done to make sure they don't "break anything."

### `KTX_FEATURE_EMBEDDED_ZSTD` and `KTX_FEATURE_EMBEDDED_TOOLS_DEPENDENCIES`

KTX-Software, via basisu, has its own copy of the zstd library. Also, for the command-line tools, it incorporates its own copies of `cxxopts` and `fmt`. In the context of vcpkg (or any package manager), we want to use the package manager's versions of these libraries instead of the embedded ones. These new options allow us to do that.

By default, both are set to `ON`, so previous behavior is unchanged.

When `KTX_FEATURE_EMBEDDED_ZSTD` is set to `OFF`, `zstd.c` is _not_ compiled into any libraries. Instead, `find_package(zstd CONFIG REQUIRED)` is used to find zstd, and `target_link_libraries(... zstd::libzstd)` is used to depend on it.

When `KTX_FEATURE_EMBEDDED_TOOLS_DEPENDENCIES` is set to `OFF`, cxxopts and fmt are not included via `add_subdirectory`. Instead, similar to the above, they're found via `find_package`.

I made these separate flags because zstd is a dependency of the ktx library itself, whereas the other two are only relevant for the CLI tools. But I'd be happy to make them just one flag (`KTX_FEATURE_EMBEDDED_DEPENDENCIES`?) or three (separate flags for each of zstd, cxxopts, and fmt) if you prefer.

### Removed `/Gz` compiler option from Windows debug builds

See this comment for justification:
https://github.com/microsoft/vcpkg/pull/30804/files#r1165360173

### Tweaks to mkversion invocation

The way `mkversion` was previously run was causing errors like this when invoked from vcpkg on Windows:

> getopt: option requires an argument -- v

Cleaning up some quotes and using `--` instead of `-c` fixes it up.

### `LIBKTX_FEATURE_APPLE_FRAMEWORK`

libktx builds as a Framework on Apple iOS and similar. This is undesirable in the vcpkg context, where we prefer a simple library and headers. So this new option lets us turn off the frameworking. See here for a bit more information:
https://github.com/microsoft/vcpkg/pull/36837

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -54,6 +54,7 @@ option( LIBKTX_FEATURE_KTX2 "Enable KTX 2 support." ON )
 option( LIBKTX_FEATURE_VK_UPLOAD "Enable Vulkan texture upload." ON )
 option( LIBKTX_FEATURE_GL_UPLOAD "Enable OpenGL texture upload." ON )
 option( LIBKTX_FEATURE_ETC_UNPACK "ETC decoding support." ON )
+option( LIBKTX_FEATURE_APPLE_FRAMEWORK "Build as a Framework on APPLE_LOCKED_OS targets." ON )
 
 CMAKE_DEPENDENT_OPTION( BASISU_SUPPORT_SSE
     "Compile with SSE support so applications can choose to use it."
@@ -329,7 +330,7 @@ macro(common_libktx_settings target enable_write library_type)
         XCODE_ATTRIBUTE_ENABLE_HARDENED_RUNTIME "YES"
     )
 
-    if(APPLE_LOCKED_OS)
+    if(APPLE_LOCKED_OS AND LIBKTX_FEATURE_APPLE_FRAMEWORK)
         set_target_properties(${target} PROPERTIES
             FRAMEWORK TRUE
         )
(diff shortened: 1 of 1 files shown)


kring:
The vcpkg version of basisu has an extensive patch to their CMakeLists.txt already. And in fact it changed in the past three days with the update to v1.60. Perhaps similar to what you had to do?
https://github.com/microsoft/vcpkg/tree/master/ports/basisu
https://github.com/microsoft/vcpkg/blob/master/ports/basisu/export-cmake-config.diff

kring:
No we're not directly collaborating @MarkCallow, so thanks for the heads up.

@dg0yt I previously opened this PR to upstream into KTX-Software a lot of the patches that are currently applied in vcpkg. As you can see above, Mark asked me to reapply these changes on top of some further work he was merging first. I plan to get back to that soon. Once these changes are rebased and merged, the vcpkg port becomes drastically simpler. Here's what it looks like (currently an overlay port in our open source product):
https://github.com/CesiumGS/cesium-native/tree/main/extern/vcpkg/ports/ktx

I'm very happy for you to pick any of this up and run with it as you see fit. Or, I'll continue what I was doing before, picking up wherever you leave off, as soon as I'm able to get back to it.

MarkCallow:
> IMO all scripting should be done in CMake.

One of the purposes of mkversion is to create a version number that reflects any local code changes that have been made, possibly post CMake config. The reason for the recent change to `cmake/version .cmake` was to restore this functionality. Is it possible to run CMake scripts at build time as opposed to config time?

PR #1086, opened 2025-11-19
Avoid using user-controlled paths as regex patterns (fix #1084)

### LINKED ISSUE #1084: #1071 broke builds in directories with characters that contain special characters
https://github.com/KhronosGroup/KTX-Software/commit/6765255bed62cada1bcba80ab7b7c5c0bf437f8f#diff-37f64d12f59f0b2664549a459f48b8d2745579d42c76fced4a4833632b6ce237R163

This line checks if the original working directory is a subdirectory of the repo root by doing a regex comparison to see if one absolute path starts with the other. However, it doesn't attempt to regex escape either path, so the check can fail even when the condition it's attempting to check for is true.

As a concrete example, try cloning the repo to a directory called `c++`. When building the project, you will see an error message stating that `Working directory is not in KTX-Software repo.`, despite the working directory being in the repo.

AnyOldName3:
This might not work on MacOS, but is fine on Linux and Windows.

Otherwise, #1084 happens, and building the library after cloning it to a path like `C:\C++Libraries\KTX-Software` doesn't work.

If `realpath` isn't available on MacOS, then I guess `$(pwd)` could be run through a `sed` script that regex-escaped it, but there are still some edge cases that that wouldn't handle well. e.g. with symlinks.

Fixes #1084.

### DIFF
--- a/scripts/mkversion
+++ b/scripts/mkversion
@@ -160,13 +160,14 @@ if [ $write_all -eq 1 ]; then
     fi
   done
 else
-  if [[ $workdir =~ $(pwd)(/(.*$)?)? ]]; then
+  relative_path=$(realpath --relative-base=$(pwd) -- $workdir)
+  if [[ $relative_path =~ ^[^/] ]]; then
     # Trailing part of workdir gives object location in repo.
-    if [ -n "${BASH_REMATCH[2]}" ]; then
+    if [[ $relative_path != . ]]; then
       if [ "$OBJ" == "." ]; then
-        OBJ=${BASH_REMATCH[2]}
+        OBJ=$relative_path
       else
-        OBJ=${BASH_REMATCH[2]}/$OBJ
+        OBJ=$relative_path/$OBJ
       fi
     # else  We are in the repo root
     fi


scripts/mkversion
```diff
@@ -160,13 +160,14 @@ if [ $write_all -eq 1 ]; then
     fi
   done
 else
-  if [[ $workdir =~ $(pwd)(/(.*$)?)? ]]; then
+  relative_path=$(realpath --relative-base=$(pwd) -- $workdir)
```
MarkCallow, line 163: macos's `realpath` does not take any options except `-q` to silence warnings. This is the reason for the CI build failure.

Please find an alternative way, at least for macOS.

MarkCallow:
> Otherwise, https://github.com/KhronosGroup/KTX-Software/issues/1084 happens, and building the library after cloning it to a path like C:\C++Libraries\KTX-Software doesn't work.

Do you see paths with '\' path separators in git bash or bash running under WSL on Windows?

Wouldn't it be easier to update the RE to allow either type of path separator?

AnyOldName3:
Somehow I'd missed the relevant paragraph in the Bash manual about quoting meaning something different within a `[[` test. That solution's much simpler.

PR #1096, opened 2025-12-18
Dependency fetching and linking rework

Daniil-SV:
I'm currently working on KTX support in [OpenImageIO](https://github.com/Daniil-SV/OpenImageIO/tree/ktx) (OIIO further), and I've encountered some issues along the way.
Builds on some of my MacOS-based CIs were failing due to errors related to missing zstd headers. For some reason, the compiler simply couldn't find them.
I don't think the problem lies with OIIO, as dependencies are compiled in a separate process during cmake configuration stage.
After a little investigation, I concluded that dependencies are working a bit messy here. I suspect that in my case, the compiler was failing because some zstd headers were used directly from basisu while others contained the repository itself.
So I decided to try to "fix"/"rework" it.

## Implementation
I decided to use [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html), because it is a modern approach to dealing with dependencies, and it gives quite a lot of opportunities for developers. Under the hood, it uses a git clone and all dependencies are collected in one centralized folder, so there shouldn't be such problems with dependency files. In my opinion, this also solves the dependency maintainability issue, you just need to change the git tag inside cmake file to update dependency version. Also, FetchContent calls find_package under the hood, which allows you to use system packages to speed up the build or use custom packages, which is I think will be very useful, at least for OIIO.

## Testing
I tested this in [OIIO CI](https://github.com/Daniil-SV/OpenImageIO/actions/runs/20314602807) and [KTX-Software CI](https://github.com/Daniil-SV/KTX-Software/actions/runs/20339753287/job/58435544869). The build works fine in both cases, but some ktxdiff tests fail. I checked each failing test and concluded that the updated zstd version (more details in the notes) changed the compressed data, causing the test to fail when comparing them, even though the actual pixel data is essentially the same. I updated the test files locally and all tests pass successfully on my Windows machine, so if necessary, I can send additional PR to the CTS repository.

## Notes
- I also have changes for Basisu, but I'm a bit confused by the fact that it uses a fork with custom changes. Currently, the fetch comes from my own fork [here](https://github.com/Daniil-SV/basis_universal), based on the original repository, where I made dependency changes and changes based on KhronosGroup fork. But I'm not sure where I should submit a pull request: to the official repository or should i replicate my changes to the KhronosGroup fork.
- I updated ZSTD from 1.4.9 to 1.5.7. When using FetchContent, it adds and configures the project as if it were a add_subdirectory, but 1.4.9 is too old, so using it causes CMake version conflicts.
- I assume that FetchContent will replace the current implementation with subrepo for those dependencies for which I wrote a cmake file, but if this is not accepted, then I am ready to rewrite it at least as an option, just tell me.
- The current version of Basisu has some warnings, so I disabled global werror for a successful build (most likely temporarily)


kring:
These changes seem reasonable to me @MarkCallow. I personally prefer to use something like vcpkg or conan rather than FetchContent, but the structure here should make that easier, too. In fact, I may hold off on recreating my changes in #1075 so that they can be built on top of this PR after it's merged.

My two cents regarding warnings and `KTX_WERROR`:

1. The KTX build should only enable higher warnings levels and warnings-as-errors in its own code, not in third-party code (whether embedded or not).
2. The best way to enable warnings-as-errors is with:

```
    set_target_properties(
        ${targetName}
        PROPERTIES
          COMPILE_WARNING_AS_ERROR YES
    )
```

This requires CMake 3.24+ (it will be harmless in older versions, but warnings won't trigger errors), but the really nice thing is it makes it easy to disable warnings-as-errors in a standard way. When running the cmake configure, just add `--compile-no-warning-as-error` and warnings won't be errors anymore. This is nice when testing on a new platform or compiler version and you don't want to fix the warnings right away.

PR #1100, opened 2025-12-23
HDR Support for libktx and tools

### LINKED ISSUE #1109: Option to build tools could be more visible
In order to build tools such as `toktx`, I need to set `KTX_FEATURE_TOOLS`, like this:
```
cmake . -B build -DKTX_FEATURE_TOOLS=ON
```
I think it would be good to make this more visible to the user, maybe mention it in documentation, and have `option( KTX_FEATURE_...` like other feature options.

ViNeek:
The pull request adds initial support for HDR data in libktx and tools

In particular,

1. The CLI ktx tool is updated so that:

- the `ktx info` command supports **UASTC HDR** payload formats as defined in the KTX Specification.
- the `ktx validate` command implements new validation clauses related to the **UASTC HDR** payload formats
- the `ktx encode` command now accepts **uastc-hdr-4x4** and **uastc-hdr-6x6i** as codec strings. 
- These codecs require the input file format to be either **R16G16B16_SFLOAT** or **R16G16B16A16_SFLOAT**.
- the `ktx create` command when the **encode** parameter is used, accepts **uastc-hdr-4x4** and **uastc-hdr-6x6i** for the **codec**  parameter
- if the input is raw, these codecs require the format parameter to be either  **R16G16B16_SFLOAT** or **R16G16B16A16_SFLOAT**; 
- if the input is not raw, the input files must be EXR with half-float data type
- the ktx transcode supports the two new codecs for input files: 
- **uastc-hdr-4x4** inputs must accept three targets: **rgba16f**,  **astc-hdr-4x4**, and **bc6hu**
- **uastc-hdr-6x6i** inputs must accept three targets: **rgba16f**, **astc-hdr-6x6**, and **bc6hu**
- if an input file is neither of these two formats, the targets are rejected,
- the `ktx extract` command supports the two new codecs for input files 
- **uastc-hdr-4x4** inputs may be extracted as raw, or transcoded as in `ktx transcode`
- **uastc-hdr-6x6i** inputs require transcoding as in `ktx transcode` when used with this command 
- the `ktx compare` command, when used with **UASTC HDR 4x4** or **UASTC HDR 6x6** data and pixel comparison is requested, the image is transcoded to uncompressed half-float values prior to comparison

2. The existing JavaScript bindings is updated to support the new functionality, specifically transcoding of UASTC HDR payloads

Incidentally fixes #1109.

### DIFF
--- a/interface/js_binding/ktx_wrapper.cpp
+++ b/interface/js_binding/ktx_wrapper.cpp
@@ -1542,7 +1542,17 @@ EMSCRIPTEN_BINDINGS(ktx)
       .property("uastcRDOMaxSmoothBlockErrorScale", &ktxBasisParams::uastcRDOMaxSmoothBlockErrorScale)
       .property("uastcRDOMaxSmoothBlockStdDev", &ktxBasisParams::uastcRDOMaxSmoothBlockStdDev)
       .property("uastcRDODontFavorSimplerModes", &ktxBasisParams::uastcRDODontFavorSimplerModes)
-      .property("uastcRDONoMultithreading", &ktxBasisParams::uastcRDONoMultithreading);
+      .property("uastcRDONoMultithreading", &ktxBasisParams::uastcRDONoMultithreading)
+
+      /* UASTC HDR params */
+
+      .property("uastcHDRQuality", &ktxBasisParams::uastcHDRQuality)
+      .property("uastcHDRUberMode", &ktxBasisParams::uastcHDRUberMode)
+      .property("uastcHDRUltraQuant", &ktxBasisParams::uastcHDRUltraQuant)
+      .property("uastcHDRFavorAstc", &ktxBasisParams::uastcHDRFavorAstc)
+      .property("rec2020", &ktxBasisParams::rec2020)
+      .property("uastcHDRLambda", &ktxBasisParams::uastcHDRLambda)
+      .property("uastcHDRLevel", &ktxBasisParams::uastcHDRLevel)
     ;
 
     constant("ANIMDATA_KEY", std::string(KTX_ANIMDATA_KEY));


external/dfdutils/KHR/khr_df.h
```diff
@@ -512,6 +518,16 @@ typedef enum _khr_df_model_channels_e {
     KHR_DF_CHANNEL_UASTC_RRR   = 4U,
     KHR_DF_CHANNEL_UASTC_RRRG  = 5U,
     KHR_DF_CHANNEL_UASTC_RG    = 6U,
+    /* MODEL UASTC_4X4_LDR (alias) */
+    KHR_DF_CHANNEL_UASTC_4X4_LDR_RGB  = 0U,
```
MarkCallow: Same comment I made over in dfdutils. Don't include the block dimensions otherwise we'll see an explosion of aliases. Just use LDR and HDR.

interface/js_binding/ktx_wrapper.cpp
```diff
@@ -161,6 +169,14 @@ namespace ktx
                 return KHR_DF_PRIMARIES_UNSPECIFIED;
         }
 
+        bool isHDR() const
+        {
+            return (getColorModel() == KHR_DF_MODEL_BC6H
+                 || getColorModel() == KHR_DF_MODEL_UASTC_4X4_HDR
+                 || getColorModel() == KHR_DF_MODEL_UASTC_6x6_HDR
+                 );
+        }
```
MarkCallow: I think this needs to return true also for KHR_DF_MODEL_ASTC && format is an ASTC HDR format.

Don't we need a native version of this function too?

interface/js_binding/ktx_wrapper.cpp
```diff
@@ -1279,6 +1310,7 @@ EMSCRIPTEN_BINDINGS(ktx)
         .property("dataSize", &ktx::texture::getDataSize)
         .property("baseWidth", &ktx::texture::baseWidth)
         .property("baseHeight", &ktx::texture::baseHeight)
+        .property("colorModel", &ktx::texture::getColorModel)
 #if KTX_FEATURE_WRITE
         .property("oetf", &ktx::texture::getOETF, &ktx::texture::setOETF)
```
MarkCallow, line 1344: Not related to HDR but, given the overhead of PRs and releases, please add

```
.property("transferFunction", &ktx::texture::getTransferFunction, &ktx::texture::setTransferFunction) and the functions in ktx::texture. OETF was renamed some time ago in the native code, because we are not always dealing with an OETF, though the old names remain also for compatibility.

File: lib/include/KHR/khr_df.h [hunk 4]
```diff
@@ -1,4 +1,4 @@
-/* The Khronos Data Format Specification (version 1.4.0) */
+/* The Khronos Data Format Specification (version 1.4.1) */
```
MarkCallow: A note for when you update KTX-Software to use dfdutils as a subproject. I want to avoid the duplication of khr_df.h. For the purposes of building, it can be avoided by making the dfdutils directory a public include directory of the dfdutils library target. It also needs to be exported from the `ktx` target (libktx) and ultimately it needs to be installed along with ktx.h.

lib/include/ktx.h
```diff
@@ -1357,6 +1357,23 @@ ktxTexture2_CompressAstc(ktxTexture2* This, ktx_uint32_t quality);
...
+/**
+ * @~English
+ * @brief Options specifiying basis codec.
+ */
+typedef enum ktx_basis_codec_e {
+    KTX_BASIS_CODEC_NONE  = 0,
+        /*!< NONE. */
+    KTX_BASIS_CODEC_ETC1S   = 1,
+        /*!< BasisLZ. */
+    KTX_BASIS_CODEC_UASTC   = 2,
```
MarkCallow: Make this UASTC_LDR as that is what BasisU is now calling it.

lib/include/ktx.h
```diff
@@ -1381,8 +1398,8 @@ typedef struct ktxBasisParams {
         /*!< Size of this struct. Used so library can tell which version
              of struct is being passed.
          */
```
MarkCallow, line 1430: This breaks backward compatibility. Any suggestions for a way that does not require existing applications to be updated? Did you discuss breaking compatibility with @lexaknyazev.

lib/include/ktx.h
```diff
@@ -1518,6 +1535,28 @@ typedef struct ktxBasisParams {
...
+    ktx_bool_t uastcHDRUltraQuant;
+        /*!< UASTC HDR 4x4: Try to find better quantized CEM 7/11 endpoint values (slower)
+         */
+    ktx_bool_t uastcHDRFavorAstc;
+        /*!< UASTC HDR 4x4: By default the UASTC HDR 4x4 encoder tries to strike a balance or even slightly favor BC6H quality. If this option is specified, ASTC HDR 4x4 quality is favored instead.
+         */
+    ktx_bool_t rec2020;
+        /*!< UASTC HDR 6x6i specific option: The input image's gamut is Rec. 2020 vs. the default Rec. 709 - for accurate colorspace error calculations.
+         */
+    float uastcLambda;
```
MarkCallow: Why no `HDR` here? Is this how it is in basis_universal? If so, perhaps @richgel999 is planning to use this for all UASTC encoders. It is unfortunate to also have the differently named `uastcRDOQualityScalar` which is lamda for UASTC LDR though I see that has a different range of values.

lib/src/basis_encode.cpp
```diff
@@ -135,6 +135,66 @@ swizzle_to_rgba(uint8_t* rgbadst, uint8_t* rgbasrc, uint32_t src_len,
     }
 }
 
+typedef void (*PFNBUCOPYCB_HDR)(float* rgbadst, const basist::half_float* rgbasrc, uint32_t src_len,
+                                ktx_size_t image_size, swizzle_e swizzle[4]);
+
+// All callbacks expect source images to have no row padding and expect
+// component size to be 32 bits.
+static void
+copy_rgbaF_to_rgbaF(float* rgbadst, const basist::half_float* rgbasrc, uint32_t,
```
MarkCallow, line 146: Would using templates avoid any code duplication between LDR and HDR in these copy callbacks?

lib/src/basis_encode.cpp
```diff
@@ -371,6 +431,90 @@ ktxTexture2_rewriteDfd4Uastc(ktxTexture2* This,
     return KTX_SUCCESS;
 }
 
+static KTX_error_code
+ktxTexture2_rewriteDfd4UastcHDR4x4(ktxTexture2* This, alpha_content_e alphaContent, bool isLuminance,
+                             swizzle_e swizzle[4]) {
```
MarkCallow: Please align swizzle_e with ktxTexture above.

lib/src/basis_encode.cpp
```diff
@@ -371,6 +431,90 @@ ktxTexture2_rewriteDfd4Uastc(ktxTexture2* This,
...
+    KHR_DFDSETSVAL(nbdb, 0, SAMPLEUPPER, 0x3F800000U);
+
+    This->pDfd = ndfd;
+    free(cdfd);
+    return KTX_SUCCESS;
+}
+
+static KTX_error_code
+ktxTexture2_rewriteDfd4UastcHDR6x6i(ktxTexture2* This, alpha_content_e alphaContent, bool isLuminance,
+                             swizzle_e swizzle[4]) {
```
MarkCallow: Please align swizzle_e with ktxTexture above.

lib/src/basis_encode.cpp
```diff
@@ -371,6 +431,90 @@ ktxTexture2_rewriteDfd4Uastc(ktxTexture2* This,
     return KTX_SUCCESS;
 }
 
+static KTX_error_code
```
MarkCallow, line 442: Consider a template or a function with additional arguments. There is a lot of duplicated code between the this and ktxTexture2_rewriteDfd4UastcHDR6x6i.

lib/src/basis_encode.cpp
```diff
@@ -474,6 +618,25 @@ ktxTexture2_CompressBasisEx(ktxTexture2* This, ktxBasisParams* params)
     cparams.m_create_ktx2_file = false; // To avoid rewriting this code, continue with .basis.
```
MarkCallow, line 632: I think it is in the contract to replace the code here that builds a ktxTexture from .basis data in favour of having the BasisU encoder create the KTX file in memory. You can then use ktxTexture2_CreateFromMemory to make the new ktxTexture2 object from what the encoder produces.

Are you planning to do this later?

lib/src/basis_encode.cpp
MarkCallow: Should be 1 as 6x6i is supercompressed.

lib/src/basis_transcode.cpp
```diff
@@ -65,6 +65,24 @@ ktxTexture2_transcodeUastc(ktxTexture2* This,
                            ktx_transcode_fmt_e outputFormat,
                            ktx_transcode_flags transcodeFlags);
 
+static transcoder_texture_format
```
MarkCallow, line 68: Why did you not make the new enum values the same as those in the BasisU encoder so no mapping function is needed?

lib/src/texture2.c
```diff
@@ -818,7 +818,7 @@ ktxTexture2_constructFromStreamAndHeader(ktxTexture2* This, ktxStream* pStream,
     }
     This->isCompressed = (This->_protected->_formatSize.flags & KTX_FORMAT_SIZE_COMPRESSED_BIT);
 
-    if (This->supercompressionScheme == KTX_SS_BASIS_LZ && pBDFD->model != KHR_DF_MODEL_ETC1S) {
+    if (This->supercompressionScheme == KTX_SS_BASIS_LZ && pBDFD->model != KHR_DF_MODEL_ETC1S && pBDFD->model != KHR_DF_MODEL_UASTC_6x6_HDR) {
```
MarkCallow: This is clearly wrong. Presumably it is like this because the encoder is setting the wrong scheme id.

Must be fixed.

lib/src/texture2.c
```diff
@@ -2039,6 +2039,10 @@ ktxTexture2_NeedsTranscoding(ktxTexture2* This)
         return true;
     else if (KHR_DFDVAL(This->pDfd + 1, MODEL) == KHR_DF_MODEL_UASTC)
         return true;
+    else if (KHR_DFDVAL(This->pDfd + 1, MODEL) == KHR_DF_MODEL_UASTC_4X4_HDR)
```
MarkCallow: As I mentioned in an e-mail we need something like a `ktxTexture2_IsTranscodable()` function as UASTC 4x4 HDR can be used as is on devices that support ASTC. I think `ktxTexture2_NeedsTranscoding` should only be true for formats that cannot be used directly. ktxTexture2_IsTranscodable()` would be true for the superset of `ktxTexture2_NeedsTranscoding` and any formats like UASTC 4x4 HDR.

tests/webgl/libktx-read-webgl/libktx-read-test.js
```diff
@@ -353,21 +359,42 @@ function uploadTextureToGl(gl, ktexture) {
...
-      formatString = 'ETC';
-      format = transcode_fmt.ETC;
+    if (ktexture.isHDR) {
+      const model = ktexture.colorModel;
+
+      //const model = ktexture.get
+      if (astcSupported && HDRsupported) {
+        if (ktx.khr_df_model.KHR_DF_MODEL_UASTC_4X4_HDR == model) {
+          formatString = 'ASTC_HDR_4x4';
+          format = transcode_fmt.ASTC_HDR_4x4_RGBA;
```
MarkCallow, line 369: Transcoding in this case is a lot of work for nothing. The transcoder itself will likely not do anything but the overhead of passing data to the transcoder and receiving it back is likely high plus the work done in `ktxTexture2_TranscodeBasis`.

Need to set the right example so do not transcode. Just use it.

tools/ktx/command_create.cpp
```diff
@@ -171,7 +171,7 @@ struct OptionsCreate {
                     " the \'Generate Mipmap\' options to tune the resampler.",
                     cxxopts::value<float>(), "<float>")
                 (kEncode, "Encode the created KTX file. Case insensitive."
-                    "\nPossible options are: basis-lz | uastc", cxxopts::value<std::string>(), "<codec>")
+                    "\nPossible options are: basis-lz | uastc | uastc-hdr-4x4 | uastc-hdr-6x6i", cxxopts::value<std::string>(), "<codec>")
```
MarkCallow: For orthogonality add `uastc-ldr-4x4` equivalent to `uastc`.

tools/ktx/command_encode.cpp
```diff
@@ -154,7 +154,7 @@ void CommandEncode::OptionsEncode::init(cxxopts::Options& opts) {
         (kCodec, "Target codec."
                   " With each encoding option the encoder specific options become valid,"
                   " otherwise they are ignored. Case-insensitive."
-                  "\nPossible options are: basis-lz | uastc", cxxopts::value<std::string>(), "<target>");
+                  "\nPossible options are: basis-lz | uastc | uastc-hdr-4x4 | uastc-hdr-6x6i", cxxopts::value<std::string>(), "<target>");
```
MarkCallow: Add uastc-ldr-4x4 equivalent to uastc here too.

tools/ktx/command_transcode.cpp
```diff
@@ -56,8 +56,8 @@ Transcode a KTX2 file.
             selected based on the number of channels in the input texture.
             Block compressed transcode targets can only be saved in raw format.
             Case-insensitive. Possible options are:
-            etc-rgb | etc-rgba | eac-r11 | eac-rg11 | bc1 | bc3 | bc4 | bc5 | bc7 | astc |
-            r8 | rg8 | rgb8 | rgba8.
+            etc-rgb | etc-rgba | eac-r11 | eac-rg11 | bc1 | bc3 | bc4 | bc5 | bc6hu | bc7 | astc | astc-hdr-4x4 | astc-hdr-6x6 |
```
MarkCallow, line 59: I don't think you can transcode uastc-hdr-4x4 to uastc-hdr 6x6 or vice-versa. Even if you can, there is no point to a 6x6 to 4x4 transcode.  I think  the only option should be astc-hdr. Block dimension will be chosen based on the input.

tools/ktx/command_transcode.cpp
```diff
@@ -122,8 +122,8 @@ void CommandTranscode::OptionsTranscode::init(cxxopts::Options& opts) {
                    " Block compressed transcode targets can only be saved in raw format."
                    " Case-insensitive."
                    "\nPossible options are:"
-                   " etc-rgb | etc-rgba | eac-r11 | eac-rg11 | bc1 | bc3 | bc4 | bc5 | bc7 | astc |"
-                   " r8 | rg8 | rgb8 | rgba8."
+                   " etc-rgb | etc-rgba | eac-r11 | eac-rg11 | bc1 | bc3 | bc4 | bc5 | bc6hu | bc7 |"
+                   " astc | astc-hdr-4x4 | astc-hdr-6x6 | r8 | rg8 | rgb8 | rgba8 | rgba16f."
```
MarkCallow, line 126: Same comment I made on line 59.

tools/ktx/encode_utils_basis.h
```diff
@@ -279,13 +302,22 @@ struct OptionsEncodeBasis : public ktxBasisParams {
                 "Default is 18.0. Larger values expand the range of blocks considered smooth.",
                 cxxopts::value<float>(), "<deviation>")
             (kUastcRdoF, "Do not favor simpler UASTC modes in RDO mode.")
-            (kUastcRdoM, "Disable RDO multithreading (slightly higher compression, deterministic).");
+            (kUastcRdoM, "Disable RDO multithreading (slightly higher compression, deterministic).")
```
MarkCallow, line 358: I think this is incorrect. "Slightly slower" is more likely correct. I can't see disabling multithreading for RDO as having any effect on the ultimate level of compression. I realize this wording is not new and is the same as basisu_tool. Need to check with @richgel999.

tools/ktx/encode_utils_basis.h
```diff
@@ -179,6 +181,12 @@ struct OptionsEncodeBasis : public ktxBasisParams {
     inline static const char* kUastcRdoS = "uastc-rdo-s";
     inline static const char* kUastcRdoF = "uastc-rdo-f";
     inline static const char* kUastcRdoM = "uastc-rdo-m";
+    inline static const char* kUastcHdrUberMode = "uastc-hdr-uber-mode";
+    inline static const char* kUastcHdrUltraQuant = "uastc-hdr-ultra-quant";
+    inline static const char* kUastcHdrFavorAstc = "uastc-hdr-favor-astc";
+    inline static const char* kRec2020 = "rec-2020";
+    inline static const char* kLambda = "uastc-lambda";
```
MarkCallow: Need to check with @richgel999 if he plans to use this for the upcoming new LDR formats. If not, I feel this should be "uastc-hdr-lambda".

tools/ktx/encode_utils_basis.h
```diff
@@ -349,6 +381,28 @@ struct OptionsEncodeBasis : public ktxBasisParams {
                 "Invalid use of argument --{} when UASTC RDO post-processing was not enabled.", name);
     }
 
+    void validateUASTCOrUASTCHDR4x4Arg(Reporter& report, const char* name) {
+        if (codec != BasisCodec::UASTC && codec != BasisCodec::UASTC_HDR_4x4)
+            report.fatal(
+                rc::INVALID_ARGUMENTS,
+                "Invalid use of argument --{} that only applies when the used codec is UASTC or UASTC HDR 4x4.",
```
MarkCallow: "...ment, --{}, that only ..."

tools/ktx/transcode_utils.cpp
```diff
@@ -27,7 +27,10 @@ TranscodeSwizzleInfo determineTranscodeSwizzle(const KTXTexture2& texture, Repor
 
     if (texture->supercompressionScheme == KTX_SS_BASIS_LZ) {
         result.defaultNumComponents = 0;
-        if (sample0 == KHR_DF_CHANNEL_ETC1S_RGB && sample1 == KHR_DF_CHANNEL_ETC1S_AAA) {
+        if (khr_df_model_e(KHR_DFDVAL(bdfd, MODEL)) == KHR_DF_MODEL_UASTC_6x6_HDR) {
```
MarkCallow: Another place that needs fixing when the SS scheme id is fixed.

tools/ktx/validate.cpp
```diff
@@ -1003,6 +1003,11 @@ void ValidationContext::validateDFDBasic(uint32_t blockIndex, const uint32_t* df
...
                     error(DFD::SampleCountMismatch, blockIndex, samples.size(), toString(VkFormat(header.vkFormat)), expectedSamples->size());
 
+                // UASTC 4x4 requires lower to be 0.
+                if (khr_df_model_e(block.model) == khr_df_model_e::KHR_DF_MODEL_UASTC_4X4_HDR)
+                    for (auto& expectedSample : *expectedSamples) 
+                        expectedSample.lower = 0;
+
                 for (std::size_t i = 0; i < std::min(samples.size(), expectedSamples->size()); ++i) {
                     const auto& parsed = samples[i];
                     const auto& expected = (*expectedSamples)[i];
```
MarkCallow, line 1022: Isn't there more the validator should be checking?

CMakeLists.txt
```diff
@@ -45,6 +45,7 @@ option( KTX_FEATURE_PY "Create Python source distribution." OFF )
 option( KTX_FEATURE_TESTS "Create unit tests." ON )
 option( KTX_FEATURE_TOOLS_CTS "Enable KTX CLI Tools CTS tests (requires CTS submodule)." OFF )
 option( KTX_FEATURE_JS "Enable JavaScript bindings in Emscripten builds." ON )
+option( KTX_FEATURE_TOOLS "Enable KTX CLI Tools." OFF)
```
MarkCallow, line 48: I am astonished I didn't notice this was missing. Thanks for fixing it.

interface/java_binding/src/main/cpp/libktx-jni.cpp
```diff
@@ -52,7 +52,7 @@ jfieldID KtxAstcParams_normalMap_field; // "Z"
 jfieldID KtxAstcParams_perceptual_field; // "Z"
 jfieldID KtxAstcParams_inputSwizzle_field; // "[C"
 
-jfieldID KtxBasisParams_uastc_field; // "Z"
```
javagl, line 55: The `"Z"` here is only a comment (to indicate the field type without having to look at the Java side), but should be changed to `"I"` for clarity

interface/java_binding/src/main/cpp/libktx-jni.cpp
```diff
@@ -151,7 +151,7 @@ JNIEXPORT jint JNICALL JNI_OnLoad(JavaVM *jvm, void *)
 
     // Obtain the fieldIDs of the KtxBasisParams class
     if (!initClass(env, cls, "org/khronos/ktx/KtxBasisParams")) return JNI_ERR;
-    if (!initField(env, cls, KtxBasisParams_uastc_field, "uastc", "Z")) return JNI_ERR;
```
javagl, line 154: The `"Z"` indicates that the type of that field is a `boolean`, but it is actually an `int`. So the `"Z"` has to be changed to `"I"`.

external/dfdutils/KHR/khr_df.h
```diff
@@ -310,7 +310,11 @@ typedef enum _khr_df_model_e {
     /* PowerVR Texture Compression */
     KHR_DF_MODEL_PVRTC         = 164U,
     KHR_DF_MODEL_PVRTC2        = 165U,
+    /* UASTC for BASIS supercompression */
     KHR_DF_MODEL_UASTC         = 166U,
+    KHR_DF_MODEL_UASTC_4X4_LDR = 166U,
```
MarkCallow: We should use the same word order as basis_universal: UASTC_LDR_4x4, UASTC_HDR_4x4 and UASTC_HDR_6x6.

external/dfdutils/KHR/khr_df.h
MarkCallow: Since the newly released version of basis_universal only introduces one new format, `XUASTC` the explosion I feared won't happen. So we should follow the existing model where the channel Ids use the color model name which means the block size should be included as now.

interface/java_binding/src/main/java/org/khronos/ktx/KtxTexture2.java
```diff
@@ -123,7 +123,7 @@ protected KtxTexture2(long instance) {
 	 * <br>
 	 * The images are either encoded to ETC1S block-compressed format and supercompressed
 	 * with Basis LZ or they are encoded to UASTC block-compressed format. UASTC format
-	 * is selected by setting the {@link KtxBasisParams#setUastc(boolean)} to <code>true</code>.
+	 * is selected by setting the {@link KtxBasisParams#setCodecFlag(int)} to <code>2</code>.
```
MarkCallow: Is this doc. accurate? I.e does the Java wrapper still only support ETC1S and UASTC LDR? If so, it should be updated to expose the new codecs. Also this text implies if you do nothing, you will get ETC1S/BasisLZ. Is that correct? The BasisLZ encoder has the value 1 in `codecFlag` according to the comment in KtxBasisParams.java.

Do users really have to hard code the value`2`? There should be defines or constants named for all encoders.

interface/java_binding/src/test/java/org/khronos/ktx/test/KtxTexture2Test.java
```diff
@@ -485,7 +485,7 @@ public void testInputSwizzleBasisEx() throws IOException {
         // the former G channel becomes the B channel
         // the former A channel remains the A channel
         KtxBasisParams inputParams = new KtxBasisParams();
-        inputParams.setUastc(false);
+        inputParams.setCodecFlag(1);
```
MarkCallow: Need named constants or defines, whatever the Java way is, for the codec values.

interface/js_binding/ktx_wrapper.cpp
```diff
@@ -1438,7 +1487,7 @@ EMSCRIPTEN_BINDINGS(ktx)
     class_<ktxBasisParams>("basisParams")
       .constructor<>()
       .property("structSize", &ktxBasisParams::structSize)
-      .property("uastc", &ktxBasisParams::uastc)
+      .property("codecFlag", &ktxBasisParams::codecFlag)
```
MarkCallow: I don't see the codecFlag enum anywhere here. It must be defined so users can set the this property easily.

lib/include/KHR/khr_df.h
MarkCallow: @ViNeek you mentioned in an e-mail that you had fixed the duplication, yet this copy of the file still exists as does the orginal in `external/dfdutils`. What am I missing?

interface/python_binding/pyktx/ktx_texture2.c
```diff
@@ -106,7 +106,7 @@ KTX_error_code PY_ktxTexture2_CompressBasisEx(ktxTexture2 *texture,
 {
     ktxBasisParams params = {
         .structSize = sizeof(ktxBasisParams),
-        .uastc = uastc,
+        .codecFlag = (uastc) ? KTX_BASIS_CODEC_UASTC_LDR : KTX_BASIS_CODEC_ETC1S,
```
MarkCallow: The new encoders should be exposed here too but maybe it's not in the scope of the current project.

lib/src/basis_sgd.h
```diff
@@ -78,6 +78,14 @@ enum alpha_content_e {
     eGreen
 };
 
+// 1, or 2 slices per image (i.e. layer, face & slice).
```
MarkCallow: There is only 1 slice per image for UASTCHDR6x6i since it only supports RGB. Please fix the comment.

lib/CMakeLists.txt
```diff
@@ -505,6 +497,7 @@ if(LIBKTX_VERSION_FULL)
     )
     common_libktx_settings(ktx 1 ${LIB_TYPE})
     add_lib_dependencies(ktx basisu_encoder)
+    target_link_libraries(ktx PUBLIC dfdutils)
```
MarkCallow: Move this setting to common_libktx_settings?

lib/CMakeLists.txt
```diff
@@ -514,6 +507,7 @@ if(LIBKTX_VERSION_READ_ONLY)
     )
     common_libktx_settings(ktx_read 0 ${LIB_TYPE})
     add_lib_dependencies(ktx_read basisu_encoder)
+    target_link_libraries(ktx_read PUBLIC dfdutils)
```
MarkCallow: And delete this one.

lib/CMakeLists.txt
```diff
@@ -666,13 +660,13 @@ endif()
 # Install
 
 if(LIBKTX_VERSION_FULL)
-    list(APPEND LIBKTX_INSTALL_TARGETS ktx)
+    list(APPEND LIBKTX_INSTALL_TARGETS ktx dfdutils)
```
MarkCallow, line 665: dfdutils should be linked into a shared library therefore there should be no need to install it. Is this the way of getting khr_df.h installed? If so, how does it avoid also installing the library? Or has a decision been made to provide the dfdutils library to users of libktx?

lib/CMakeLists.txt
```diff
@@ -666,13 +660,13 @@ endif()
...
 endif()
 # ktx_read has never been included in the install packages and no
 # issues have been filed. Only install it if building static libs
 # or full version is not being built.
 if(LIBKTX_VERSION_READ_ONLY AND NOT (BUILD_SHARED_LIBRARIES OR LIBKTX_VERSION_FULL))
-    list(APPEND LIBKTX_INSTALL_TARGETS ktx_read)
+    list(APPEND LIBKTX_INSTALL_TARGETS ktx_read dfdutils)
```
MarkCallow, line 671: Subject to the answer to my last question in the previous comment, dfdutils for static libktx should be added using the `add_lib_dependencies` macro which is in this file.

tools/ktx/command_create.cpp
```diff
@@ -1369,9 +1383,21 @@ void CommandCreate::processOptions(cxxopts::Options& opts, cxxopts::ParseResult&
...
         }
+    } else if (options.codec == BasisCodec::UASTC_HDR_4x4 || options.codec == BasisCodec::UASTC_HDR_6x6i) {
+        switch (options.vkFormat) {
+        case VK_FORMAT_R16G16B16_SFLOAT:
+        case VK_FORMAT_R16G16B16A16_SFLOAT:
+            // Allowed formats
+            break;
+        default:
+            fatal_usage(
+                "Only RGB16 or RGBA16 SFLOAT can be encoded, "
```
MarkCallow: Please make a toString function for the BasisCodec class and use it here to append "to \<codec name\>" after "encoded".

Change the message in line 1365 to use this function too.

lib/include/ktx.h
```diff
@@ -1357,6 +1369,24 @@ ktxTexture2_CompressAstc(ktxTexture2* This, ktx_uint32_t quality);
...
+/**
+ * @~English
+ * @brief Options specifiying basis codec.
+ */
+typedef enum ktx_basis_codec_e {
+    KTX_BASIS_CODEC_NONE  = 0,
+        /*!< NONE. */
+    KTX_BASIS_CODEC_ETC1S   = 1,
+        /*!< BasisLZ. */
+    KTX_BASIS_CODEC_UASTC_LDR   = 2,
```
MarkCallow: Change to KTX_BASIS_CODEC_UASTC_LDR_4x4 for consistency with naming in basis encoder.

tools/ktx/encode_utils_basis.h
```diff
@@ -18,6 +18,8 @@ enum class BasisCodec {
     NONE = 0,
     BasisLZ,
     UASTC,
```
MarkCallow: Add UASTC_LDR_4x4 for consistency. Make UASTC an alias of this.

tools/ktx/command_encode.cpp
```diff
@@ -227,21 +239,39 @@ void CommandEncode::executeEncode() {
...
+        case VK_FORMAT_R8G8_SRGB:
+        case VK_FORMAT_R8G8B8_UNORM:
+        case VK_FORMAT_R8G8B8_SRGB:
+        case VK_FORMAT_R8G8B8A8_UNORM:
+        case VK_FORMAT_R8G8B8A8_SRGB:
+            // Allowed formats
+            break;
+        default:
+            fatal_usage(
+                "Only R8, RG8, RGB8, or RGBA8 UNORM and SRGB formats can be encoded, "
```
MarkCallow: Use the BasisCodec::toString method, requested in a comment in command_create.cpp, here to add the encoding type being attempted.

tools/ktx/command_encode.cpp
```diff
@@ -227,21 +239,39 @@ void CommandEncode::executeEncode() {
...
+    } else if (options.codec == BasisCodec::UASTC_HDR_4x4 ||
+               options.codec == BasisCodec::UASTC_HDR_6x6i) {
+        switch (texture->vkFormat) {
+        case VK_FORMAT_R16G16B16_SFLOAT:
+        case VK_FORMAT_R16G16B16A16_SFLOAT:
+            // Allowed formats
+            break;
+        default:
+            fatal_usage(
+                "Only RGB16 or RGBA16 SFLOAT can be encoded, "
```
MarkCallow: Use BasisCodec::toString here too.

tools/ktx/validate.cpp
```diff
@@ -1618,80 +1626,128 @@ void ValidationContext::validateSGD() {
     const auto buffer = std::make_unique<uint8_t[]>(sgdByteLength);
     read(sgdByteOffset, buffer.get(), sgdByteLength, "the SGD");
 
-    if (header.supercompressionScheme != KTX_SS_BASIS_LZ)
+    if ((header.supercompressionScheme != KTX_SS_BASIS_LZ) && (header.supercompressionScheme != KTX_SS_UASTC_HDR_6X6_INTERMEDIATE))
         return;
 
     // Validate BASIS_LZ SGD
```
MarkCallow: Comment is outdated as you are now validating two SGD schemes.

tools/ktx/validate.cpp
```diff
@@ -1618,80 +1626,128 @@ void ValidationContext::validateSGD() {
...
-        error(SGD::BLZEExtendedByteLengthNotZero, bgh.extendedByteLength);
-
-    // Validate ImageDesc
-    if (sgdByteLength < sizeof(ktxBasisLzGlobalHeader) + sizeof(ktxBasisLzEtc1sImageDesc) * imageCount)
-        return;
+    switch (header.supercompressionScheme) {
+        case KTX_SS_UASTC_HDR_6X6_INTERMEDIATE: {
+            uint32_t imageCount = 0;
+            // uint32_t layersFaces = numLayers * header.faceCount;
+            for (uint32_t level = 0; level < numLevels; ++level)
```
MarkCallow: Don't duplicate this code. Either move the imageCount declaration and its setting outside the switch or use a lambda.

Actually now I've reviewed further, I see that imageCount is not being used in KTX_SS_UASTC_HDR_6X6_INTERMEDIATE validation. There is no validation of the length of the SGD. Please fix.

interface/java_binding/src/test/java/org/khronos/ktx/test/KtxTexture2Test.java
javagl: Consistent with the patterns that are used elsewhere, this should be a 

```java
/*
 * Copyright (c) 2026, Khronos Group and Contributors
 * SPDX-License-Identifier: Apache-2.0
 */

package org.khronos.ktx;

/**
 * Options specifying basis codec.<br>
 * <br>
 * These constants can be passed to the {@link KtxBasisParams#setCodecFlag(int)}
 * function.
 */
public class KtxBasisCodec {
    /**
     * None
     */
    public static final int NONE = 0;

    /**
     * BasisLZ
     */
    public static final int ETC1S = 1;

    /**
     * UASTC
     */
    public static final int UASTC_LDR = 2;

    /**
     * UASTC_HDR_4X4
     */
    public static final int UASTC_HDR_4X4 = 3;

    /**
     * UASTC_HDR_6X6i
     */
    public static final int UASTC_HDR_6X6_INTERMEDIATE = 4;

    /**
     * Returns a string representation of the given basis codec
     *
     * @param n The basis codec
     * @return A string representation of the given basis codec
     */
    public static String stringFor(int n) {
	switch (n) {
	case NONE:
	    return "NONE";
	case ETC1S:
	    return "ETC1S";
	case UASTC_LDR:
	    return "UASTC_LDR";
	case UASTC_HDR_4X4:
	    return "UASTC_HDR_4X4";
	case UASTC_HDR_6X6_INTERMEDIATE:
	    return "UASTC_HDR_6X6_INTERMEDIATE";
	}
	return "[Unknown KtxBasisCodec]";
    }

    /**
     * Private constructor to prevent instantiation
     */
    private KtxBasisCodec() {
	// Prevent instantiation
    }

}
```
Then the call here would be
`inputParams.setCodecFlag(KtxBasisCodec.ETC1S);`

javagl: Edited: I had copy-pasted the 
`specifiying` from the native header, but now changed it to
`specifying` ...

lib/include/ktx.h
```diff
@@ -1357,6 +1369,24 @@ ktxTexture2_CompressAstc(ktxTexture2* This, ktx_uint32_t quality);
 KTX_API KTX_error_code KTX_APIENTRY
 ktxTexture2_DecodeAstc(ktxTexture2* This);
 
+/**
+ * @~English
+ * @brief Options specifiying basis codec.
```
MarkCallow, line 1391: Typo. s/ifiy/ify/

ViNeek: FIXED on 77893b30916334be10dfb9717621957b45497513

lib/include/ktx.h
agkar: FIXED on ad8096276c396bb742d82fe28532e6c6d7533af8

lib/include/ktx.h
agkar: Fixed on 5195768161ed70f92b1f5700f89ec6709627b7e6

tools/ktx/validate.cpp
ViNeek: Fixed on 52a2b430d52ce2a715e68e8fcc8afe4f259b98a5

tools/ktx/validate.cpp
ViNeek: Fixed on 52a2b430d52ce2a715e68e8fcc8afe4f259b98a5

tools/ktx/validate.cpp
ViNeek: Fixed on 52a2b430d52ce2a715e68e8fcc8afe4f259b98a5

lib/src/basis_encode.cpp
agkar: Fixed on d94f1fa143e55f398a5cbab0ac48f514f4dd1fb9

lib/src/texture2.c
agkar: Fixed on f739c5f0a59b281cb4426a6f8f24e7895c9f0635

tools/ktx/encode_utils_basis.h
agkar: Fixed on 5195768161ed70f92b1f5700f89ec6709627b7e6

tools/ktx/encode_utils_basis.h
agkar: Fixed on 3e578fcfd0b96bfa2b024bd0c02268045d0fb2f6

tools/ktx/command_create.cpp
agkar: Fixed on 96035fa3c9f164a2c0de539a42a9d622353f008e

interface/js_binding/ktx_wrapper.cpp
ViNeek: Added on 7ed9345e77325852f2c23fbec3e677ac450ded75

tools/ktx/command_encode.cpp
agkar: Fixed on b8809789fb250747bcae54b9f7fa85c1fe28429e

tools/ktx/command_encode.cpp
agkar: Fixed on b8809789fb250747bcae54b9f7fa85c1fe28429e

lib/src/basis_transcode.cpp
ViNeek: Enum values have started clashing so direct mapping is not possible anymore. E.g, KTX_TTF_BC1_OR_3 = 23 in KTX and 	cTFASTC_HDR_4x4_RGBA = 23 in basisu.

tools/ktx/transcode_utils.cpp
ViNeek: Fxed as of 2bfface794bbf2e6c954b1c0c7db8234212bca7d

interface/js_binding/ktx_wrapper.cpp
ViNeek: Fixed in 8d7482d824afd53f506da69fe925927b375ab887

lib/include/KHR/khr_df.h
ViNeek: Fixed in 6b7a0f4e4b8538f9b62ed8c9ac128f30b517806e. The file deletion was not pushed in the initial commits.

lib/src/basis_transcode.cpp
MarkCallow: Ahh! I see. I stupidly picked values for `KTX_TTF_BC1_OR_3`  and `KTX_TTF_ETC` too close to the existing ones and Rich has used them.

Since it looks like we will be upping the major version due to the `bool uastc` to code enum change in `ktx.h` we could just change these values. However transcode is probably more widely used than encode so doing this would break more users. We should probably leave it the way you have done it.

MarkCallow:
We need to be consistent about naming UASTC_LDR_4x4. The rules should be

- Follow the basisu_encoder naming, i.e, UASTC_LDR_4x4.
- Only use bare "UASTC" as an alias for the above for backward compatibility.

Some places still just have UASTC. Some have just UASTC_LDR. I think the latter are my fault due to a previous incorrect comment. Sorry about that.

lib/include/ktx.h
MarkCallow: Apologies. basisu_encoder and other Binomial code refers to this as codec as UASTC LDR 4x4. We should use the same terminology so please change again.

agkar: Fixed on cbd134f912d7e412f3da72962cdbced5a1d1eb29

lib/src/basis_encode.cpp
agkar: Fixed on b26289cf648419723226a9895a32831032b17ad7

tools/ktx/command_create.cpp
agkar: Fixed on f2dad7a121a0f716c413c5183261a018a83e0aff

tools/ktx/command_encode.cpp
agkar: Fixed on f2dad7a121a0f716c413c5183261a018a83e0aff

lib/include/ktx.h
agkar: Fixed on cbd134f912d7e412f3da72962cdbced5a1d1eb29

tools/ktx/encode_utils_basis.h
agkar: Fixed on f2dad7a121a0f716c413c5183261a018a83e0aff

lib/include/ktx.h
ViNeek: This was discussed with @lexaknyazev and we concluded as acceptable expecting future changes

tools/ktx/command_transcode.cpp
ViNeek: Invalid combinations are handled internally and reported to the user. In general, as we have discussed, it is best to avoid any form of automatic disambiguation of block sizes for this release. It's the same discussion that we had with transcoding for the JS bindings.

tools/ktx/command_transcode.cpp
ViNeek: Please see answer for line 59

external/dfdutils/KHR/khr_df.h
ViNeek: Fixed in 33e5919392f5d218cef621b56e82a40f075530ce and onwrads

lib/src/texture2.c
```diff
@@ -2042,6 +2042,26 @@ ktxTexture2_GetPrimaries_e(ktxTexture2* This)
...
+ * @memberof ktxTexture2
+ * @~English
+ * @brief Query if the images require transcoding or can be used directly format.
+ *
+ * @param[in]     This     pointer to the ktxTexture2 object of interest.
+ */
+ktx_bool_t
+ktxTexture2_IsTranscodable(ktxTexture2* This)
 {
     if (KHR_DFDVAL(This->pDfd + 1, MODEL) == KHR_DF_MODEL_ETC1S)
```
MarkCallow: Just call ktxTexture2_NeedsTranscoding rather than duplicating the code. The only addition here should be UASTC_HDR_4X4.

lib/src/basis_encode.cpp
MarkCallow: Decision on 2026.2.9 conf. call that the HDR and SDR code is sufficiently different that templates would not help.

lib/include/ktx.h
```diff
@@ -1442,7 +1442,7 @@ typedef struct ktxBasisParams {
 
     /* ETC1S params */
 
-    ktx_uint32_t compressionLevel;
+    ktx_uint32_t etc1sCompressionLevel;
         /*!< Encoding speed vs. quality tradeoff. Range is [0,6]. Higher values
```
MarkCallow: Please update this documentation to correspond with that in the version of basisu_encoder now being used:

```
        /*!< ETC1S compression effort level. Range is [0,6]. Higher values are much
             slower, but give slightly higher quality. Higher levels are intended
             for video. This parameter controls numerous internal encoding speed vs.
             compression efficiency/performance tradeoffs. Note this is NOT the same
             as the ETC1S quality level, and most users shouldn't change this.
             There is no default. Callers must explicitly set this value. Callers
             can use KTX\_ETC1S\_DEFAULT\_COMPRESSION\_LEVEL as a default value.
             Currently this is 2.
```

ViNeek: Fixed with b6d517dd203677db4b5c6bc874f120b7476051a7

lib/src/texture2.c
ViNeek: Fixed with c5ac97c59b233016b7d46a1871165c9e4f03afaf

tools/ktx/command_encode.cpp
MarkCallow: The linked commit does not add a BasisCodec::toString method and the error message still does not say which format you are attempting to encode to.

lib/include/ktx.h
MarkCallow: Fixed in [b6d517d](https://github.com/KhronosGroup/KTX-Software/pull/1100/commits/b6d517dd203677db4b5c6bc874f120b7476051a7)

lib/src/texture2.c
MarkCallow: Fixed in [c5ac97c](https://github.com/KhronosGroup/KTX-Software/pull/1100/commits/c5ac97c59b233016b7d46a1871165c9e4f03afaf).

lib/src/basis_encode.cpp
```diff
@@ -1111,6 +1111,7 @@ ktxTexture2_CompressBasisEx(ktxTexture2* This, ktxBasisParams* params)
                         kimages[image].rgbSliceByteOffset = slice->m_file_ofs
                                                        - level_file_offsets[level];
                         kimages[image].rgbSliceByteLength = slice->m_file_size;
+                        kimages[image].rgbSliceType = 0x0000ABCD;
```
MarkCallow: I think this should be set from the header of the encoded bitstream not hardcoded.

tools/ktx/validate.cpp
```diff
@@ -1664,6 +1664,20 @@ void ValidationContext::validateSGD() {
 
                             if (image.rgbSliceByteOffset + image.rgbSliceByteLength > levelIndices[level].byteLength)
                                 error(SGD::UH6X6IEInvalidRGBSlice, level, layer, face, zSlice, image.rgbSliceByteOffset, image.rgbSliceByteLength, levelIndices[level].byteLength);
+
+                            if (image.rgbSliceType != 0x0000ABCD)
```
MarkCallow: Re this and L1674. Must allow both possible compression type values. We validate against the spec. not what this implementation can do.

tools/ktx/validate.cpp
MarkCallow: Made moot by [490a2ff](https://github.com/KhronosGroup/KTX-Software/pull/1100/commits/490a2ff8469bd4b6163540c26ab78b1877649d0c).

tools/ktx/validate.cpp
ViNeek: Improved in dfc9e96742dccaf0ad89af33c339c1e48d60994e

lib/src/basis_encode.cpp
ViNeek: Improved in dfc9e96742dccaf0ad89af33c339c1e48d60994e

tools/ktx/validate.cpp
```diff
@@ -1665,14 +1665,26 @@ void ValidationContext::validateSGD() {
...
 
-                            if (image.rgbSliceType != 0x0000ABCD)
-                                error(SGD::UH6X6IEInvalidRGBSliceType, level, layer, face, zSlice, image.rgbSliceType);
+                            const auto sgd_profile = static_cast<unsigned char>(image.rgbSliceType >> 8);
+                            const auto sgd_version = static_cast<unsigned char>(image.rgbSliceType);
+
+                            if (sgd_profile != 0xAB)
+                                error(SGD::UH6X6IEInvalidRGBSliceTypeProfile, level, layer, face, zSlice, sgd_profile);
+
+                            if ((sgd_version != 0xCD) && (sgd_version != 0xCE))
```
MarkCallow, line 1674: Where did the idea to split rgbSliceType into profile and version come from? Does the Binomial documentation use those terms somewhere that I missed?

My concern is more about the error messages than the code. Will people understand them? The KTX spec only talks about rgbSliceType.

interface/python_binding/pyktx/ktx_texture2.c
agkar: Yes, not in the scope of the current project.

tools/ktx/command_encode.cpp
agkar: Fixed on b7def46b43ee6959eb68b56f4f0452e8127e5ed3

tools/ktx/validate.cpp
ViNeek: Yes the idea came from the Binomial spec. Since the KTX spec explictly mentions "The bitstream is as documented in the [format specification](https://github.com/BinomialLLC/basis_universal/wiki/UASTC-HDR-6x6-Intermediate-File-Format-(Basis-GPU-Photo-6x6)) on the basis_universal Wiki" I thought that the best handling for validation is through the actual Binomial spec of the underlying bitstream. Should we revert back to simply requiring a slice type of `0xABCD` or `0xABCE`? Is there something more that we can convey to the user other than these values?

tools/ktx/command_encode.cpp
agkar: Fixed on b7def46b43ee6959eb68b56f4f0452e8127e5ed3

interface/java_binding/src/test/java/org/khronos/ktx/test/KtxTexture2Test.java
ViNeek: Outside of the scope of this project

lib/src/basis_encode.cpp
agkar: The channelID is KHR_DF_CHANNEL_UASTC_HDR_4X4_RGB or KHR_DF_CHANNEL_UASTC_HDR_6X6_RGB for HDR textures. swizzle is not used

lib/src/basis_sgd.h
ViNeek: Fixed in caea6eaeb68b1b834d3f6aa92c58cd295bdaad68

tests/webgl/libktx-read-webgl/libktx-read-test.js
ViNeek: Updated in 9e04b19a133be3d467c89841a14bb4666a8de5cc

lib/CMakeLists.txt
ViNeek: Moved in d52d33308333315ddf55168d68be81551f3d568a

lib/CMakeLists.txt
ViNeek: Moved in d52d33308333315ddf55168d68be81551f3d568a

tools/ktx/validate.cpp
MarkCallow: Thanks. I had not refreshed the open tab in my browser so had not seen the recent changes. Given what I see now, using profile and type are fine. I wonder if I should update the KTX specification description of rgbSliceVersion.

lib/src/basis_transcode.cpp
MarkCallow, line 179: Need to check that all rgbSliceVersions in the UASTC 6x6i global data are 0xABCD as our current Binomial version does not support 0xABCE. I don't know if the Basis library has a query for supported "stream semantic versions". I asked Rich but did not get a reply. It is preferable to use a query, if one exists.

lib/src/basis_encode.cpp
MarkCallow: I am talking about text alignment in the source code. If swizzle is not used I suggest also removing `swizzle[4]` here. Aren't you getting a warning (error) about the unused parameter?

interface/java_binding/src/test/java/org/khronos/ktx/test/KtxTexture2Test.java
MarkCallow: @javagl please provide a PR to add this to the Java interface after I have merged this PR to main.

lib/src/basis_encode.cpp
agkar: Fixed on fefca867fbf6938dd8103841786fe5bec1a74145 . We are not getting a warning because the parameters are declared as unused

lib/src/basis_transcode.cpp
ViNeek: Please review a9ac560115ded528ffd52ef0055e11df1c6f5c6a for this. I have made sure that we are validating against 0xABCD for now but I have left the code required for the future 2.0 merge in comments.

interface/java_binding/src/test/java/org/khronos/ktx/test/KtxTexture2Test.java
javagl: @MarkCallow I took a "TODO" for that. Should be a quick fix. 

(I'm receiving notifications for this PR, and will try to keep track of when it was merged. But there have been quite a few comments/updates, so I'm not tracking everything closely. I might easily have overlooked that `@mention`. In doubt, you can always ping me via mail)

lib/src/basis_transcode.cpp
MarkCallow: The change to `basis_transcode.cpp` is fine but please revert the changes to the validator. It must validate according to the spec.

ViNeek: Reverted in 296150b1a8878ca5fb13afe4c845dd296173fa12

MarkCallow:
I just reviewed the test updates again. Now I see that there are no tests of `ktx validate` on the new formats. This must be tested. Please add suitable tests.

MarkCallow:
I am not liking having `NXN` in `KHR_DF_MODEL` and `KTX_SS_UASTC_HDR` and NxN everywhere else including the Vulkan and OpenGL format names and `KTX_PACK_ASTC_BLOCK_DIMENSION`. Please change the `KTX_SS_UASTC_HDR` names. I can probably tolerate the KHR names being different but would ideally like to  have those changed too.

MarkCallow:
`khr_df.h` is not installed by, at least the macOS, installer.

MarkCallow:
Re `khr_df.h`, an install like this one in dfdutils.h

```
install(DIRECTORY
        ${CMAKE_CURRENT_SOURCE_DIR}/KHR/
        DESTINATION include/KHR
)
```

needs a line, `COMPONENT dev`, to add it to the installer. However I'm not sure the `dev` component will be known to `dfdutils/CMakeLists.txt`. You might have to add this to `ktx/CMakeLists.txt`. There are old install attempts in the latter file that can probably be deleted. See LL 206-210 and LL 722-735.

I think `dfdutils/CMakeLists.txt` should install only `KHR/khr_df.h` when EMBED is set, otherwise it should install it, `dfdutils.h` and the library. I can do that myself later, if you don't want to do it.

interface/java_binding/src/test/java/org/khronos/ktx/test/KtxTexture2Test.java
javagl: This is now tracked/addressed in https://github.com/KhronosGroup/KTX-Software/pull/1147

Issue #1101, opened 2025-12-26
Tools do not compile with g++14


ltjax:
I'm getting these errors with the `Reporter`s

```
In file included from /home/ltjax/.conan2/p/b/ktx0e591fa15a9b9/b/src/tools/ktx/command.cpp:6:
/home/ltjax/.conan2/p/b/ktx0e591fa15a9b9/b/src/tools/ktx/command.h: In instantiation of ‘void ktx::Reporter::fatal_usage(Args&& ...) [with Args = {const char (&)[26], const std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&}]’:
/home/ltjax/.conan2/p/b/ktx0e591fa15a9b9/b/src/tools/ktx/command.h:233:31:   required from here
/home/ltjax/.conan2/p/b/ktx0e591fa15a9b9/b/src/tools/ktx/command.h:122:19: error: ‘args#0’ is not a constant expression
  122 |         fmt::print(std::cerr, std::forward<Args>(args)...);
      |         ~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

```

See also: https://github.com/conan-io/conan-center-index/pull/28400#issuecomment-3693432592

I think the first parameter needs to be explicitly spelled out as `fmt::format_string` for compile-time checking, as described by the {fmt} author here: https://stackoverflow.com/a/69596901/518626 - or compile time checks be disabled via fmt::runtime or vformat

MarkCallow:
It is set here:

https://github.com/KhronosGroup/KTX-Software/blob/0eaebb015c29a8945c8df4a35f9bf2cb5c925a7f/tools/ktx/CMakeLists.txt#L82-L86

After a little research, it looks like `target_compile_features` sets a minimum required but if CMAKE_CXX_STANDARD sets something higher that will be used. Check if Conan is setting CMAKE_CXX_STANDARD. Perhaps I should use

```
set_property(TARGET ktxtools PROPERTY CXX_STANDARD 17)
```
 instead. This will not be overridden by CMAKE_CXX_STANDARD.

There seem to be a lot of differences between c++17 and c++20, such as the one you encountered, that can break compilation so allowing an override to happen is not a good idea.

ltjax:
That makes sense, and also explains why my build worked with v4.3.2, because it still used the `CXX_STANDARD` way of setting the language version.

[...]

PR #1102, opened 2025-12-30
Fix build problems w/ fmt and C++23

ltjax:
PR for Issue #1101 
Pretty superficial fixes, e.g. formatOptionalFileOffset could probably be refactored to work without fmt::runtime, but this was less invasive.

Unfortunately, it doesn't compile with C++ 20 and 23 for me anymore after I rebased onto main, but due to different problems, i.e.:
```
/home/ltjax/code/KTX-Software/tools/ktx/command_convert.cpp:208:65: error: no matching function for call to ‘ktx::OutputStreamEx::OutputStreamEx(std::u8string, ktx::CommandConvert&)’
  208 |     OutputStreamEx outputStream(outputFilepath.u8string(), *this);
      |                                                                 ^
/home/ltjax/code/KTX-Software/tools/ktx/command_convert.cpp:82:5: note: candidate: ‘ktx::OutputStreamEx::OutputStreamEx(const std::string&, ktx::Reporter&)’
   82 |     OutputStreamEx(const std::string& filepath, Reporter& report)
      |     ^~~~~~~~~~~~~~
/home/ltjax/code/KTX-Software/tools/ktx/command_convert.cpp:82:39: note:   no known conversion for argument 1 from ‘std::u8string’ {aka ‘std::__cxx11::basic_string<char8_t>’} to ‘const std::string&’ {aka ‘const std::__cxx11::basic_string<char>&’}
   82 |     OutputStreamEx(const std::string& filepath, Reporter& report)
      |                    ~~~~~~~~~~~~~~~~~~~^~~~~~~~
```
I don't know how to proceed there. You seem to have put some thought into utf8 handling and might better know how to fix that one.

### DIFF
--- a/tools/ktx/command_compare.cpp
+++ b/tools/ktx/command_compare.cpp
@@ -648,11 +648,11 @@ public:
         const auto space = outputFormat != OutputFormat::json_mini ?  " " : "";
         const auto nl = outputFormat != OutputFormat::json_mini ?  "\n" : "";
 
-        auto formatOptionalFileOffset = [](const std::optional<std::size_t>& fileOffset, std::size_t imageOffset, bool json) {
+        auto formatOptionalFileOffset = [](const std::optional<std::size_t>& fileOffset, std::size_t imageOffset, bool json) -> std::string {
             if (fileOffset.has_value())
                 return fmt::format(fmt::runtime(json ? "{}" : "0x{:x}"), *fileOffset + imageOffset);
             else
-                return fmt::format(fmt::runtime(json ? "null" : "N/A"));
+                return json ? "null" : "N/A";
         };
 
         auto formatPacked = [=](const ImageSpan::TexelBlockPtr<>& texelBlock, bool json) {


tools/ktx/command_compare.cpp
```diff
@@ -178,7 +178,7 @@ struct DiffHexFixedWidth : public DiffBase<T> {
 
     virtual std::string value(std::size_t index, OutputFormat format) const override {
         if (format == OutputFormat::text) {
-            return fmt::format(fmt::format("0x{{:0{}x}}", sizeof(T) << 1), DiffBase<T>::rawValue(index));
+            return fmt::format(fmt::runtime(fmt::format("0x{{:0{}x}}", sizeof(T) << 1)), DiffBase<T>::rawValue(index));
```
MarkCallow: What is the fmt::runtime for?

ltjax: It just turns off the compile-time checking of the format string for the outer format call. I basically added it whenever the format string was runtime-generated.

Actually, I think in this case, you can probably get away with just a single format call and get rid of `fmt::runtime`. AFAIK, fmt can deal with placeholders in placeholders just fine.

MarkCallow: I'd prefer if we can get rid of it.

ltjax: Ok, done.

ltjax:
Ok, I rebased and removed that `fmt::runtime`, but it still does not compile with C++20 and C++23, albeit with different errors:

For `CMAKE_CXX_STANDARD=23` I'm getting:
```
/home/ltjax/code/KTX-Software/external/basis_universal/encoder/cppspmd_sse.h:496:24: error: cannot bind non-const lvalue reference of type ‘cppspmd_sse41::spmd_kernel::vfloat&’ to an rvalue of type ‘cppspmd_sse41::spmd_kernel::vfloat’
  496 |                 return dst;
      |                        ^~~
/home/ltjax/code/KTX-Software/external/basis_universal/encoder/cppspmd_sse.h: In member function ‘cppspmd_sse41::spmd_kernel::vfloat& cppspmd_sse41::spmd_kernel::store_all(vfloat&&, const vfloat&)’:
/home/ltjax/code/KTX-Software/external/basis_universal/encoder/cppspmd_sse.h:508:24: error: cannot bind non-const lvalue reference of type ‘cppspmd_sse41::spmd_kernel::vfloat&’ to an rvalue of type ‘cppspmd_sse41::spmd_kernel::vfloat’
  508 |                 return dst;
      |    
```
Not sure what the intention is with those, since the type does not really seem movable? But it's an external anyways, so I'm not sure whether you want that fixed.

For `CMAKE_CXX_STANDARD=20`, the first error I'm getting is:
```
/home/ltjax/code/KTX-Software/utils/platform_utils.h:125:21: error: could not convert ‘std::__cxx11::basic_string<char8_t>((& s)->std::__cxx11::basic_string<char>::begin(), (& s)->std::__cxx11::basic_string<char>::end(), std::allocator<char8_t>())’ from ‘basic_string<char8_t>’ to ‘basic_string<char>’
  125 |         return std::u8string(s.begin(), s.end());
      |                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                     |
      |                     basic_string<char8_t>
```
I think the return type should be `std::u8string`, but even after fixing that there are still a lot of utf8 related errors, e.g. the additional `OutputStream` (without Ex) constructor has `OutputStream::`, but it defined in the class. And `InputStream` and `fmtInFile` in command_compare still cannot deal with std::u8string.

MarkCallow:
> I think the return type should be std::u8string

Sorry for the stupid copy and paste error. The return type should indeed be std::u8string. Please fix it in this PR.

> e.g. the additional OutputStream (without Ex) constructor has OutputStream::, but it defined in the class.

Sorry about that too. I started implementation in command.cpp but when I realized it would simply convert the input and call the other constructor I moved it to the class definition.

I do not understand how everything built in our CI. All of the libktx tests use c++20 and two of them include `platform_utils.h`.Most peculiar.

I have fixed the InputStream and fmtInFile issues. Please see the c++fixes branch for the fixes. I am leaving tomorrow for a 10-day trip. I'm not sure I will have time to merge this to main before I leave. Should I not, please feel free to include the fixes directly in this PR.

I won't run the workflows until the fixes are in this PR.

I could not thoroughly test building the ktx tool with the above changes because of all the fmt related issues whose fixes are in this PR.

As for the C++23 issue with `cppspmd_sse.h`, please report it to https://github.com/BinomialLLC/basis_universal/.

ltjax:
Great! It works for C++17 and C++20 for me now, C++23 if I fix or disable (via `BASISU_SSE=OFF`) the SSE stuff

tools/ktx/command_compare.cpp
MarkCallow: I still see fmt::runtime calls in several of the changed files.

tools/ktx/command_compare.cpp
```diff
@@ -650,9 +650,9 @@ class PrintDiff {
 
         auto formatOptionalFileOffset = [](const std::optional<std::size_t>& fileOffset, std::size_t imageOffset, bool json) {
             if (fileOffset.has_value())
-                return fmt::format(json ? "{}" : "0x{:x}", *fileOffset + imageOffset);
+                return fmt::format(fmt::runtime(json ? "{}" : "0x{:x}"), *fileOffset + imageOffset);
             else
-                return fmt::format(json ? "null" : "N/A");
+                return fmt::format(fmt::runtime(json ? "null" : "N/A"));
```
MarkCallow: Is this and the fmt::runtime in line 653 still needed?

tools/ktx/validate.cpp
```diff
@@ -119,25 +119,25 @@ struct ValidationContext {
...
-        callback(ValidationReport{issue.type, issue.id, std::string{issue.message}, fmt::format(issue.detailsFmt, std::forward<Args>(args)...)});
+        callback(ValidationReport{issue.type, issue.id, std::string{issue.message}, fmt::format(fmt::runtime(issue.detailsFmt), std::forward<Args>(args)...)});
     }
 
     template <typename... Args>
     void fatal(const IssueFatal& issue, Args&&... args) {
         ++numError;
         returnCode = +rc::INVALID_FILE;
-        const auto report = ValidationReport{issue.type, issue.id, std::string{issue.message}, fmt::format(issue.detailsFmt, std::forward<Args>(args)...)};
+        const auto report = ValidationReport{issue.type, issue.id, std::string{issue.message}, fmt::format(fmt::runtime(issue.detailsFmt), std::forward<Args>(args)...)};
```
MarkCallow, line 140: Are the 4 fmt::runtime calls in this file still needed too?

tools/ktx/command_compare.cpp
ltjax: The first can be refactored to use two different `fmt::format` calls instead, the second doesn't really need formatting (but probably needs to explicitly be converted into `std::string` then, one way or another). Should I do that?

tools/ktx/validate.cpp
ltjax: Yes, at the moment. Getting rid of them would need a totally different approach, i.e. tighter coupling between what is currently the `Issue` type-hierarchy and the `Args...`

MarkCallow: Leave this one as is then.

tools/ktx/command_compare.cpp
MarkCallow: Thanks for the explanation. Leave 653 as is but please change 655 to convert to a std::string.

ltjax: Ok, done using an explicit lambda return type. Hope that works for you.

Issue #1122, opened 2026-02-17
Implement KTXmapRange


MarkCallow:
`KTXmapRange` is not implemented in the HDR additions. It needs to be displayed by `ktx info`, validated by `ktx validate` and settable in `ktx create`. This new metadata item was added to the KTX spec. as part of the HDR format additions.

_Originally posted by @MarkCallow in https://github.com/KhronosGroup/KTX-Software/issues/1100#issuecomment-3911004522_

Issue #1120, opened 2026-02-17
Use `NxN` consistently for format size in tokens


MarkCallow:
I am not liking having `NXN` in `KHR_DF_MODEL` and `KTX_SS_UASTC_HDR` and NxN everywhere else including the Vulkan and OpenGL format names and `KTX_PACK_ASTC_BLOCK_DIMENSION`. We should change the `KTX_SS_UASTC_HDR` names. I can probably tolerate the KHR names being different but would ideally like to  have those changed too. Discuss the latter with @fluppeteer.

_Originally posted by @MarkCallow in https://github.com/KhronosGroup/KTX-Software/issues/1100#issuecomment-3906760073_

Issue #1119, opened 2026-02-17
Add an early out to `ktxTexture2_TranscodeBasis` when `outputFormat` is ASTC_HDR_4x4 and the texture format is UASTC_HDR_4X4


MarkCallow:
Add a check to `ktxTexture2_TranscodeBasis` so that it returns immediately if `outputFormat` is `KTX_TTF_ASTC_HDR_4x4_RGBA` and the texture's `colorModel` is `KHR_DF_MODEL_UASTC_HDR_4X4`.

_Originally posted by @MarkCallow in https://github.com/KhronosGroup/KTX-Software/issues/1100#issuecomment-3906735646_

MarkCallow:
> So if someone requests transcoding to `KTX_TTF_ASTC_HDR_4x4_RGBA`, should the texture be updated to reflect the new format (in headers, formats e.t.c.) or should we simply return immediately without touching either the data or the metadata.

The before and after format is identical with the exception of the colorModel, channel id (and, IIRC, the sign qualifier on the channel) so I was thinking return immediately but it is more semantically correct to update the colorMode, channel id and sign qualifier. The reason I want an early return is because of cubemaps, arrays and 3d textures. For example I have a cubemap with 72 images.

PR #1124, opened 2026-02-18
Improved HDR support

ViNeek:
Addressing issues #1122, #1119 and #1120

### DIFF
--- a/lib/src/basis_transcode.cpp
+++ b/lib/src/basis_transcode.cpp
@@ -36,6 +36,7 @@
   #pragma GCC diagnostic ignored "-Wunused-value"
 #endif
 #include "transcoder/basisu_transcoder.h"
+#include "transcoder/basisu_astc_helpers.h"
 #if defined(__GCC__) && !defined(__clang__)
   #pragma GCC diagnostic pop
 #endif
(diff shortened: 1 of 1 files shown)


tools/ktx/command_create.cpp
```diff
@@ -1913,6 +1913,16 @@ void CommandCreate::executeCreate() {
                 options.swizzle->c_str());
     }
 
+    // Add KTXmapRange metadata
```
MarkCallow, line 1916: There needs to be a CLI option to set the range both here and in ktx encode. In here it must be possible to set it for any supported formats, not just UASTC_HDR formats. I think it is not necessary to set it unless the option has been specified.

The spec says

"This metadata entry may be used with:
any floating point format. In the case of a combined depth-stencil format, e.g. D32_SFLOAT_S8_UINT it must be applied only to the depth component.

any linear *_UNORM or *_SNORM format.

It is incompatible with and must not be used with:
non-linear formats such as those with sRGB encoding,

true integer or boolean formats.

If KTXmapRange appears in a KTX file containing these formats, the file is invalid."

MarkCallow (commented):
Thank you for the quick work on these issues. I have just the one comment.

I will review the test updates next.

tools/ktx/command_create.cpp
MarkCallow: It occurs to me that an explicit option will be hard to use. You have to find the minimum and maximum values in the input file to get the correct values. We should automate this. I have to think about an easy way to achieve this. If the minimum and maximum values are separated by 16 float bits or less then it can be fully automatic. If the input range is greater than representable by 16 bits then some input from the user is needed.

Suggestions welcome.

PR #1129, opened 2026-02-20
Fix #1121

### LINKED ISSUE #1121: `khr_df.h` is not included in the install packages
`khr_df.h` is not installed by, at least the macOS, installer. It is not included in the install packages. This stems from the change to use the dfdutils library instead of explicitly including the source files.

_Originally posted by @MarkCallow in https://github.com/KhronosGroup/KTX-Software/issues/1100#issuecomment-3907398115_

ViNeek:
Make sure that `KHR/khr_df.h` is properly installed and installed with the `package` target on Mac OS X. #1121

### DIFF
--- a/tools/imageio/exr.imageio/exrinput.cc
+++ b/tools/imageio/exr.imageio/exrinput.cc
@@ -139,8 +139,9 @@ void ExrInput::open(ImageSpec& newspec) {
     const uint32_t width = header.data_window[2] - header.data_window[0] + 1;
     const uint32_t height = header.data_window[3] - header.data_window[1] + 1;
 
-    // Use "chromaticities" attribute, if present, to determine color primaries
-    khr_df_primaries_e colorPrimaries = KHR_DF_PRIMARIES_UNSPECIFIED;
+    // Use "chromaticities" attribute, if present, to determine color primaries.
+    // Per the EXR spec. in the absence of chromaticities, use bt.709.
+    khr_df_primaries_e colorPrimaries = KHR_DF_PRIMARIES_BT709;
     for (int i = 0; i < header.num_custom_attributes; ++i) {
         auto attributeName = std::string_view(header.custom_attributes[i].name);
         if (attributeName == "chromaticities") {


lib/CMakeLists.txt
```diff
@@ -722,11 +722,15 @@ endif()
...
 # preserve the include source folder hierarchy.
 # See https://gitlab.kitware.com/cmake/cmake/-/issues/16739.
-if (APPLE_LOCKED_OS)
+if (APPLE)
     set_source_files_properties(
-          include/KHR/khr_df.h
+          ${CMAKE_SOURCE_DIR}/external/dfdutils/KHR/khr_df.h
           PROPERTIES MACOSX_PACKAGE_LOCATION Headers/KHR
     )
+    install(FILES "${CMAKE_SOURCE_DIR}/external/dfdutils/KHR/khr_df.h"
```
MarkCallow, line 730: I do not think this is sufficient. We must support the Linux and Windows installers too. The install for those is commented out below.

Why did you add a special case for APPLE? I think the commented out install at line 736 is probably sufficient for all 3 OSes. Maybe change to install(DIRECTORY like you have in dfdutils.

If you do the install from cmake, rather than building a package, won't both this CMakeLists.txt and that in dfdutils attempt to install this file on the local machine?

lib/CMakeLists.txt
```diff
@@ -722,11 +722,15 @@ endif()
 # Use of this to install KHR/khr_df.h is due to CMake's failure to
 # preserve the include source folder hierarchy.
 # See https://gitlab.kitware.com/cmake/cmake/-/issues/16739.
-if (APPLE_LOCKED_OS)
+if (APPLE)
```
MarkCallow, line 725: APPLE_LOCKED_OS is because we only make an app bundle for those hence the location given here. On macOS we install to /usr/local. Revert this change.

MarkCallow:
Do not spend any more time on this PR. I will be creating a replacement very soon. I did not want to waste your time by making you do a lot of iterations while I figured out some things with CMake. I also saw a lot of other cmake cleanup I wanted to do.

Thank you for this anyway.  It provide some inspiration and the change to ktx-uninstall was particularly helpful. I would have missed it otherwise.

PR #1132, opened 2026-02-21
Add support cmake multiconfig generators

ygrik-rit:


MarkCallow:
Why is this change necessary? Multi-config generators are already supported by a change to the setting of CMAKE_RUNTIME_OUTPUT_DIRECTORY https://github.com/KhronosGroup/KTX-Software/blob/e0325cc60831263031c8f8d7401b03cd720ed6c0/external/basis_universal/CMakeLists.txt#L19-L22
The note about overwriting the output applies to the unmodified CMakeLists.txt when used with single config generators.

Upstream has no interest in taking this fix.

Also why do you want to build basisu tool within the KTX-Software project?

ygrik-rit:
> change to the setting of CMAKE_RUNTIME_OUTPUT_DIRECTORY

so in my case it anyway create subdir (RelWithDebInfo, Debug, etc), maybe in newer version cmake (4+) this behavior was changed. 
> Upstream has no interest in taking this fix.

Anyway this change does not change the behavior for single generators, just adds support for multi-config generators. From the CMake point of view, this a more accurate approach to specifying the path to the target
> Also why do you want to build basisu tool within the KTX-Software project?

I want use command line tool for convert images to .ktx2. I found only basisu executable in this project, which looks like what I need.

MarkCallow:
> > change to the setting of CMAKE_RUNTIME_OUTPUT_DIRECTORY
> 
> so in my case it anyway create subdir (RelWithDebInfo, Debug, etc), maybe in newer version cmake (4+) this behavior was changed.

We have been using the GenEx method for years in this project's root CMakeLists.txt. This feature is not new in CMake 4+. I do not know why you still see subdirs. I always use a multi-generator config, Xcode is my development environment, and after the change to a GenEx I no longer see added subdirs. Perhaps the previous setting of CMAKE_RUNTIME_OUTPUT_DIRECTORY is stored in your cache.

> 
> > Upstream has no interest in taking this fix.
> 
> Anyway this change does not change the behavior for single generators, just adds support for multi-config generators. From the CMake point of view, this a more accurate approach to specifying the path to the target

But it does so in a way that will break any scripts or tests that expect to find basisu in the bin directory of `basis_universal` because it will be in bin/Debug or bin/Release.

> 
> > Also why do you want to build basisu tool within the KTX-Software project?
> 
> I want use command line tool for convert images to .ktx2. I found only basisu executable in this project, which looks like what I need.

You had to really burrow down into the source code to find the basisu. It is not exposed in the root cmake project because we use only the encoder library.

This project has its own suite of tools for working with .ktx2 files. You'll find these in the ktxtools target (tools/ktx in the source code). They are all grouped under the umbrella `ktx` command. `ktx create` is the tool for creating .ktx2 files from images in other formats. This project's README.md lists all the tools and has a link to the documentation for them.

PR #1139, opened 2026-02-25
Fix issue with HDR transcode and extract

### LINKED ISSUE #1134: Can't extract UASTC_HDR_4x4 files, with potential fix
If I create a UASTC HDR 4x4 file:

```
>ktx create --format R16G16B16_SFLOAT --encode uastc-hdr-4x4 desk.exr 3.ktx2
```

Then try to extract it:

```
>ktx extract 3.ktx2 x.exr
ktx extract fatal: Requested format conversion from VK_FORMAT_ASTC_4x4_SFLOAT_BLOCK is not supported.
```

It fails. It also fails on files generated by Basis Universal v2.1 (which will be released sometime this week):

```
>ktx extract C:\dev\bu_2_13\basis_universal\bin\uastc_hdr_4x4\Desk.ktx2 3.exr
ktx extract fatal: Requested format conversion from VK_FORMAT_ASTC_4x4_SFLOAT_BLOCK is not supported.
```

However, the fix seems simple. In lib/src/texture2.c:

Image

With this fix KTX-Software extracts UASTC HDR 4x4 files generated via ktx.exe or basisu.exe v2.1.

ViNeek:
The PR should fix the issues with `ktx extract` for both UASTC 4x4 and UASTC 6x6i (see issue #1134). The code has also been patched to properly extract mip map levels. Thank you @richgel999 for providing the code snippet.

Extraction from regular ASTC data `ktx extract --all astc_hdr_6x6.ktx2` won't work. Direct encoding/decoding of VK_FORMAT_ASTC_6x6_SFLOAT_BLOCK and VK_FORMAT_ASTC_4x4_SFLOAT_BLOCK (or any other SFLOAT block size) is not supported by the current version of the software. Same goes with `ktx create`:

A command like `ktx create --format VK_FORMAT_ASTC_4x4_SFLOAT_BLOCK input.exr output.ktx2` does not support the creation of an `VK_FORMAT_ASTC_4x4_SFLOAT_BLOCK`-encoded ktx2 file.

Fixes #1134.

### DIFF
--- a/lib/src/basis_transcode.cpp
+++ b/lib/src/basis_transcode.cpp
@@ -933,7 +933,8 @@ transcodeUastcHDR6x6_intermediate(ktxTexture2* This, alpha_content_e alphaConten
 
     // Pointer and length of the image description seek table within the global supercompressed data.
     // This array of structs contain offsets and length fields relative to each mipmap level's data.
-    const ktxUASTCHDR6x6IntermediateImageDesc* imageDescs = reinterpret_cast<ktxUASTCHDR6x6IntermediateImageDesc*>(This->_private->_supercompressionGlobalData);
+    const ktxUASTCHDR6x6IntermediateImageDesc* imageDescs =
+        reinterpret_cast<ktxUASTCHDR6x6IntermediateImageDesc*>(This->_private->_supercompressionGlobalData);
     const uint64_t totalImageDescs = This->_private->_sgdByteLength / sizeof(ktxUASTCHDR6x6IntermediateImageDesc);
 
     for (ktx_int32_t level = This->numLevels - 1; level >= 0; level--) {


lib/src/basis_transcode.cpp
```diff
@@ -931,16 +931,20 @@ transcodeUastcHDR6x6_intermediate(ktxTexture2* This, alpha_content_e alphaConten
     std::vector<basisu_transcoder_state> xcoderStates;
     xcoderStates.resize(This->isVideo ? This->numFaces : 1);
 
+    // Pointer and length of the image description seek table within the global supercompressed data.
+    // This array of structs contain offsets and length fields relative to each mipmap level's data.
+    const ktxUASTCHDR6x6IntermediateImageDesc* imageDescs = reinterpret_cast<ktxUASTCHDR6x6IntermediateImageDesc*>(This->_private->_supercompressionGlobalData);
```
MarkCallow: Please break this long line after the '='. I know it's minor but I hate long  lines. They slow down my reading.

MarkCallow:
I have merged the new tests. Please update the reference here.

I attempted to do a manual FF merge to keep the same commit id but your fork does not have the latest changes so it was not possible.

lib/src/basis_transcode.cpp
ViNeek: Fixed with 4ce46a8d53d16d406fe964dbc41d54c1d432387d

MarkCallow:
Thanks @ViNeek.

PR #1157, opened 2026-04-14
Remove `--rec2020` flag and add range mapping functionality to `ktx create` and `ktx extract`

### LINKED ISSUE #1127: Remove --rec2020 option; set encoder param automatically from the DFD values
if the input .exr file has bt.2020 chromaticities. The loader already matches the chromaticities attribute, if present, to a KHR_DF_PRIMARIES value.

### LINKED ISSUE #1142: Provide help for users to set KTXmapRange metadata
A command line argument to explicitly provide the scale and offset values will be difficult to use as another tool would have to be used first to determine the min and max data values in the input image and then scale and offset have to be calculated. Better is an option to say "set it" and let _ktx _create_ determine min and max and calculate the values. If the range exceeds that supported by the output format then user input will have to be requested.

ViNeek:
This PR fixes #1142 and fixes #1127.

For #1127, it removes the flag from `ktx create` and `ktx encode` and makes sure that the profile is automatically inferred from the loaded EXR file.

For #1142 It adds 3 knew flags to `ktx create`, namely, `map-range-auto`, `map-range-offset` and `map-range-scale` to be used for mapping of floating point values in HDR formats. 
 
`ktx extract` is also updated to handle inverse mapping of values when the `KTXmapRange` key/value pair is present in a KTX2 file.

Currenlty, `offset` and `scale` are not used for the alpha channel.
Currently, range mapping only works with floating point data types and the new HDR formats.


tools/imageio/image.hpp
```diff
@@ -1335,6 +1384,352 @@ class ImageT : public Image {
     virtual ImageT& copyToRGB(Image& dst, std::string_view swizzle) override { return copyTo((ImageT<componentType, 3>&)dst, swizzle); }
     virtual ImageT& copyToRGBA(Image& dst, std::string_view swizzle) override { return copyTo((ImageT<componentType, 4>&)dst, swizzle); }
 
+
+    virtual RangeMapping mapRangeInfo(VkFormat format, ktx_basis_codec codec) override {
```
MarkCallow: This breaks modularity by introducing VkFormat and ktx_basis_codec to the low-level Image utility library. I'd prefer if this was moved to one of the existing, or a new, utility file in `tools/ktx`. Also can it use the DFD information to determine the range rather than a switch on vkFormat? If you must use a switch, I think you need to add `[[fallthrough]];` on all the cases that are falling through.

tools/ktx/command_create.cpp
```diff
@@ -1891,15 +1927,25 @@ void CommandCreate::executeCreate() {
...
         struct KTXmapRange {
             float scale;
             float offset;
         };
-        KTXmapRange data = {1, 0};
+        KTXmapRange data = {scale, offset};
         ktxHashList_AddKVPair(&texture->kvDataHead, KTX_MAP_RANGE_KEY, sizeof(KTXmapRange), &data);
     }
 
+    // Set Rec 2020
```
MarkCallow: I think this is best set in `ktxTexture2_CompressBasisEx` in libktx from the primaries in the input ktxTexture2. Otherwise everyone has to worry about setting it. If the`rec2020` option is set in the incoming Basis params then that must override the primaries of the ktxTexture2. The documentation in `lib/include/ktx,h` must be updated to state that is to be used only when it is desired to override the primaries of the input texture.

MarkCallow (changes requested):
Thank you for this. Just two, hopefully straightforward, requests

MarkCallow:
For the Web build problem you need to change line 155 of `glloader.c` to

```
static MAYBE_UNUSED GLboolean supportsSwizzle = GL_TRUE;
```

It seems the latest version of Clang is issuing set but unused warnings for static global variables.

tools/ktx/command_create.cpp
MarkCallow: On reflection, I think we should remove the rec2020 option from `ktxBasisParams`. We've never had the equivalent for sRGB, the related Binomial compressor params option has always been set from the DFD transfer function, and nobody has complained.

MarkCallow (commented):
Code looks good now. Thanks. Just need to fix the build and test errors.

I was expecting this new review to remove the "requested changes" marker. It did not. I guess I need to mark the PR as approved but I don't feel I can do that until the errors are fixed.

MarkCallow:
You have probably realized this already but just in case ...

The PR does not update the CTS reference so it is pointing to the same test set as `main`. You should update it to point at the changed test set you submitted in https://github.com/KhronosGroup/KTX-Software-CTS/pull/67 before investigating the test failures.

tools/ktx/command_create.cpp
agkar: Fixed on 1cf0fba7082bc850cb8871faf96d3667ce10d82a.

MarkCallow (approved):

MarkCallow:
I have merged https://github.com/KhronosGroup/KTX-Software-CTS/pull/67. I was unable to do so without the commit id changing so please update the CTS reference here.

ViNeek:
daef714028258d7b457af9b17667b8312df56bbe should fix this

MarkCallow:
Thank you for this effort @ViNeek and @agkar.

PR #1158, opened 2026-04-20
Fix handling of DFD linear qualifier

### LINKED ISSUE #1156: Problems with handling of DFD linear qualifier
Two problems were identified as part of reviewing https://github.com/KhronosGroup/KTX-Specification/issues/231.

- [x] `ktx create` does not set this qualifier when setting a non-linear transfer function on a UNORM format texture with an alpha channel.
- [x] `ktx validate` incorrectly reports that the qualifier must not be set for a UNORM texture even if the transfer function is non-linear.

MarkCallow:
Fix `ktx create` to set the linear qualifier according to the transfer function and channel id, instead of always setting it to the default value for the output VkFormat.

Fix the validator to allow values that differ from the default provided they comply with the KTX specification.

Fixes #1156.


cwfitzgerald:
Thanks for all the help with this!

PR #1170, opened 2026-05-14
Update basisu to v2.1.0

MarkCallow:


solbjorn:
Pls make sure you always handle `m_ktx2_and_basis_srgb_transfer_function` correctly as BasisU 2.0 changed the default value (unless the branch you're based this PR on handles it differently) (just a reminder from my comment under the previous PR).

Also, why not the HEAD commit of BasisU, but `HEAD~46`? I've seen several fixes there in the last couple weeks which might be important (can't say for sure).

MarkCallow:
> Pls make sure you always handle `m_ktx2_and_basis_srgb_transfer_function` correctly as BasisU 2.0 changed the default value (unless the branch you're based this PR on handles it differently) (just a reminder from my comment under the previous PR).

Looking into this. Thanks for pointing it out again.

> 
> Also, why not the HEAD commit of BasisU, but `HEAD~46`? I've seen several fixes there in the last couple weeks which might be important (can't say for sure).

Because we want the latest release for more stability especially as we are about to make a release.

MarkCallow:
> Pls make sure you always handle m_ktx2_and_basis_srgb_transfer_function correctly as BasisU 2.0 changed the default value (unless the branch you're based this PR on handles it differently) (just a reminder from my comment under the previous PR).

I committed a change to set it to false for non-sRGB but it does not make any difference to the results because basisu_encoder only uses it to determine what transfer function metadata to write to the output file and _libktx_ writes the .ktx2 file itself. The `perceptual` param controls data handling in the encoder and we were already setting that false.

solbjorn:
@MarkCallow, I see, thanks! Glad it's finally merged.

PR #1171, opened 2026-05-14
fix: macOS universal build

atteneder:
## Summary

Since ~4.4.0 I was not able to build a universal `libktx` or `libktx_read` macOS binary anymore. Even though those will fade out with Apple deprecating Intel CPU support, I still want to support that for the time being.

Here's what this PR does to fix that:

1. Multiple `astc-encoder*` targets may get added as dependency (in most cases `astcenc-sse4.1-static` and `astcenc-neon-static`).
2. SSE 4.1 is enabled for Basis Universal `x86_64` only (so not for `arm64`)

### Test

On a Mac navigate to the root of this repository clone and run:

```shell
cmake . -B build-mac-test -G Xcode -D CMAKE_OSX_ARCHITECTURES="arm64;x86_64"
cmake --build build-mac-test
```

Verify SSE instructions:

```shell
LIB=build-mac-test/Debug/libktx_read.4.0.0.dylib

# Disassemble the dylib slice and look for x86 SSE instructions that
# can only exist if SSE was actually compiled in. pshufb / pmaddubsw
# are SSSE3+ and appear in basisu's SSE kernels but never in arm64 code.
otool -tV -arch x86_64 "$LIB" | grep -cE '\b(pshufb|pmaddubsw|pmulhrsw)\b'
# output: 16
otool -tV -arch arm64  "$LIB" | grep -cE '\b(pshufb|pmaddubsw|pmulhrsw)\b'
# output: 0 (makes sense :) )
```

### Behavioral change

Due to a (presumably CMake) limitation the build only succeeds if `CMAKE_OSX_ARCHITECTURES` is set explicitly to `arm64;x86_64`. Using `$(ARCHS_STANDARD)` leads to linking errors.

Since before it did not compile at all, I consider this an acceptable improvement.

To reduce confusion I added an explicit configuration error that instructs users to use `arm64;x86_64` explicitly. I prefer that option over silently overriding the value, which would eventually lead to future problems should Apple decide to change that standard arch set in the future.

### Not Addressed

I verified building universal iOS simulator binaries works, but have not checked whether the x86_64 arch has SSE enabled. I presume not, but do not consider this a blocker.

## History

This picks up the previous attempt from #1043 , which was not accepted since the BasisU was built without SSE optimization in that one.


atteneder:
I noticed legacy code that sets compiler option `-msse4.1` (lib/CMakeLists.txt:495). It's likely safe to remove:

It's effectively redundant now, and the per-arch block accidentally proves it.

  The target_compile_options at line 495 fires when `BASISU_SSE` is `TRUE` and the compiler is Clang/AppleClang/GNU. After
  the change at line 541 forces `BASISU_SSE` `FALSE` for any universal build, the only configuration where it still fires
  is single-arch `x86_64` with Clang/GCC (Linux, Windows-clang, or a hypothetical Intel-only macOS build). For
  arm64-only, universal macOS, and MSVC builds it's a no-op.

  The empirical evidence that it isn't actually needed even in that remaining case comes from the per-arch override
  block at lines 745–761:

  - For the x86_64 slice of ktx in a universal build, we deliberately inject BASISU_SUPPORT_SSE=1 via -Xarch_x86_64 but
   do not inject -msse4.1 on ktx (only on basisu_encoder).
  - That x86_64 slice compiles basis_encode.cpp cleanly. So basisu's PUBLIC headers, when consumed by ktx, do not parse
   any SSE 4.1 intrinsics that would require -msse4.1 — they just reference an extern flag (g_cpu_supports_sse41). The
  actual SSE 4.1 instructions live inside basisu_kernels_sse.cpp, where basisu_encoder's own -msse4.1 (from
  external/basis_universal/CMakeLists.txt) is what matters.

  So the same logic applies to the single-arch x86_64 case: ktx doesn't need -msse4.1 either, only basisu_encoder does,
   and basisu already arranges that on its own target.

  Tradeoff: it's defensive code that's been there a long time, so removing it has a small chance of breaking some
  configuration nobody currently exercises (e.g. Linux GCC x86_64 build with a future basisu version that does inline
  SSE 4.1 intrinsics in a header). If you'd like to delete it for clarity I can; otherwise leaving it as harmless dead
  weight on the cases where it still fires is also defensible.

atteneder:
I noticed that currently it's not possible to force SSE off via `BASISU_SSE` option anymore.

To solve it I suggest to introduce option `LIBKTX_FEATURE_BASISU_SSE`.

After that change the last message about the legacy compiler option may become obsolete.

edit: Can confirm that the `-msse41` compiler mentioned above is indeed still subject to be removed.

MarkCallow:
@atteneder, I'll be away for the next 5 days. I will review when I return. I am puzzled by your statement that it is not possible to force SSE off via BASISU_SSE. I can see no deliberate reason for that. Perhaps there is an error in basis_universal's CMakeLists.txt. Make sure the CMake cache is cleared before running your tests. Maybe the value is coming from there.

atteneder:
> @atteneder, I'll be away for the next 5 days. I will review when I return.

Much appreciated! No rush.

> @atteneder, I am puzzled by your statement that it is not possible to force SSE off via BASISU_SSE. I can see no deliberate reason for that. Perhaps there is an error in basis_universal's CMakeLists.txt. Make sure the CMake cache is cleared before running your tests. Maybe the value is coming from there.

Apologies for the confusing statements. Of course `BASISU_SSE` still works as intended by the authors of BasisU. However, this PR changes how it's used. In order to work around BasisU's limiting compile-time switch `BASISU_SSE` is disabled but its effects (compile option `-msse4.1` and define `BASISU_SUPPORT_SSE`) are later injected for arch `x86_64` only like so:

```cmake
if(APPLE_MAC_OS AND LIBKTX_FEATURE_BASISU_SSE)
    ####################################################
    # Per-architecture SSE 4.1 in universal macOS builds.
    ####################################################
    # basisu_encoder cannot enable SSE for the whole project in a universal build
    # (the -msse4.1 flag would break the arm64 slice). With BASISU_SSE OFF, basisu
    # emits BASISU_SUPPORT_SSE=0 PUBLIC and no -msse4.1; we then use clang's
    # -Xarch_<arch> driver option to inject -msse4.1 and override
    # BASISU_SUPPORT_SSE to 1 only for the x86_64 sub-invocation. -Xarch_x86_64
    # applies to the immediately-following argument only, so each flag needs its
    # own prefix. Works with any generator (Xcode, Ninja, Make) on Apple.
    if(APPLE_MAC_OS_UNIVERSAL AND APPLE_MAC_OS_ARCH_x86_64)
        # SHELL: prevents CMake from deduplicating the repeated -Xarch_x86_64
        # tokens, which would otherwise leave only the first flag arch-qualified.
        target_compile_options(basisu_encoder PRIVATE
            "SHELL:-Xarch_x86_64 -msse4.1"
            "SHELL:-Xarch_x86_64 -UBASISU_SUPPORT_SSE"
            "SHELL:-Xarch_x86_64 -DBASISU_SUPPORT_SSE=1"
        )
        foreach(t ktx ktx_read)
            if(TARGET ${t})
                target_compile_options(${t} PRIVATE
                    "SHELL:-Xarch_x86_64 -UBASISU_SUPPORT_SSE"
                    "SHELL:-Xarch_x86_64 -DBASISU_SUPPORT_SSE=1"
                )
            endif()
        endforeach()
    endif()
endif()
```
[Source](https://github.com/KhronosGroup/KTX-Software/pull/1171/changes#diff-4803677bd280f4dded52c428e3c8ac6c618508dcabcbe880b68be85897c2ad41R736-R763)

This ensures SSE is enabled for `x86_64` but won't break `arm64` builds.

I needed to introduce `LIBKTX_FEATURE_BASISU_SSE` so that users are able to opt-out of SSE again.

I opened the PR as soon as the macOS universal builds were functional. There's still issues with some (cross-compiling) platform builds that I'm trying to address now.

### What to do with `set_target_processor_type`?

It appears that `set_target_processor_type` (in `cputypetest.cmake`) has historically been used to work around the BasisU SSE issue. It added a certain amount of complexity to the setup and still handles certain cases of cross-compilation poorly (embedded Linux `aarch64` as an example).

I'd consider getting rid of it, if this PR proves to handle the BasisU SSE issue reliably.

That being said, this would need extensive testing of all platform compilation targets to ensure that there's no regression.

I can offer limited support in getting the CI to succeed and test my own CI setups in addition.

atteneder:
Update: I've tested macOS Universal with Xcode and Ninja generators, but Unix Makefiles did not succeed.

edit: I personally think that's not a blocker, but you be the judge.

cmake/cputypetest.cmake
```diff
@@ -109,51 +109,39 @@ function(set_target_processor_type out)
...
-            # set to $(ARCHS_STANDARD) an Xcode build setting whose value in
-            # recent Xcode versions is "x86_64 arm64" to create universal binaries.
-            # This is passed to Xcode which calls the compiler twice passing the
-            # same set of target_definitions each time so in this case we have
-            # to choose BASISU_SUPPORT_SSE=OFF or the arm64 compile will fail.
+            # For macOS the caller may set CMAKE_OSX_ARCHITECTURES to a single
+            # arch (e.g. "x86_64" or "arm64") or to a list ("arm64;x86_64") to
+            # request a universal build. The literal "$(ARCHS_STANDARD)" is
+            # rejected by the root CMakeLists.txt because CMake stores it
+            # unexpanded and breaks $<TARGET_OBJECTS:...> paths.
```
MarkCallow: Please add a pointer to the detailed explanation of the problem in the root CMakeLists.txt.

Too bad we can't query the ARCHS_STANDARD build setting value and expand it ourselves.

cmake/cputypetest.cmake
```diff
@@ -109,51 +109,39 @@ function(set_target_processor_type out)
...
+            # representative arch — the universal-aware code paths in
+            # lib/CMakeLists.txt key off `universal_build` and the per-arch
+            # `arch_arm64`/`arch_x86_64` flags instead. In particular,
+            # BASISU_SSE is forced off for universal builds and the x86_64
+            # slice gets -msse4.1 re-enabled per-arch via -Xarch_x86_64.
             if(APPLE AND NOT "${CMAKE_SYSTEM_NAME}" STREQUAL "Darwin")
                 # Building for iOS, iPadOS, etc. Since we don't care what
                 # type of ARM processor, arbitrarily set armv8.
                 # It should be arm64 but there is a check in tests/CMakeLists.txt
                 # that is dropping loadtests for Apple Silicon arm64.
```
MarkCallow: Probably should fix this and tests/CMakeLists.txt. I don't recall why it wants to drop loadtests for Apple Silicon arm64.

BUILDING.md
```diff
@@ -248,7 +248,7 @@ cmake --build build/mac
 ```
 ##### Apple Silicon and Universal Binaries
 
-Macs are either based on Intel or the newer Apple Silicon architecture. By default CMake configures to build for your host's platform, whichever it is. If you want to cross compile universal binaries (that support both platforms), add the parameter `-DCMAKE_OSX_ARCHITECTURES="\$(ARCHS_STANDARD)"` to cmake.
+Macs are either based on Intel or the newer Apple Silicon architecture. By default CMake configures to build for your host's platform, whichever it is. If you want to cross compile universal binaries (that support both platforms), add the parameter `-DCMAKE_OSX_ARCHITECTURES="arm64;x86_64"` to cmake.
 
 > **Known limitations:**
 > - Intel Macs have support for SSE, but if you're building universal binaries,
```
MarkCallow: This limitation is now outdated. Delete.

lib/CMakeLists.txt
```diff
@@ -24,7 +26,7 @@ project(libktx
 if(APPLE)
     if(CMAKE_SYSTEM_NAME STREQUAL "iOS" OR CMAKE_SYSTEM_NAME STREQUAL "tvOS" OR CMAKE_SYSTEM_NAME STREQUAL "visionOS")
         set( APPLE_LOCKED_OS ON )
-    else()
+    elseif(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
```
MarkCallow: Why add "Darwin"? This is inside an `if(APPLE)`. Are you expecting Apple to change the name or introduce another OS?

lib/CMakeLists.txt
```diff
@@ -46,6 +48,21 @@ include(${KTX_ROOT_DIR}/cmake/cputypetest.cmake)
 
 include(GNUInstallDirs) # Must be after project.
 
+
+# Detect macOS multi-architecture universal build.
+set(APPLE_MAC_OS_UNIVERSAL OFF)
+if(APPLE_MAC_OS)
+    # Check CMAKE_OSX_ARCHITECTURES for multiple architectures.
+    list_contains(CMAKE_OSX_ARCHITECTURES "$(ARCHS_STANDARD)" APPLE_MAC_OS_ARCH_STANDARD)
```
MarkCallow: I thought we can't use $(ARCHS_STANDARD).

lib/CMakeLists.txt
```diff
@@ -598,27 +623,10 @@ endif()
...
-    list(FIND CMAKE_OSX_ARCHITECTURES "$(ARCHS_STANDARD)" archs_standard)
-    list(LENGTH CMAKE_OSX_ARCHITECTURES architecture_count)
-    if(NOT ${archs_standard} EQUAL -1 OR architecture_count GREATER 1)
-        set(universal_build ON)
-    endif()
+if(APPLE_MAC_OS_UNIVERSAL)
     # Set ordinary variable to override astc-encoder option's ON default
     # and hide the option.
-    set(ASTCENC_UNIVERSAL_BUILD ${universal_build})
+    set(ASTCENC_APPLE_MAC_OS_UNIVERSAL ON)
```
MarkCallow: This is no longer an astc-encoder option, is it? So, at the least, the comment is inaccurate. But why create this new variable?

lib/CMakeLists.txt
```diff
@@ -631,7 +639,16 @@ endif()
 # choose ASTCENC_ISA_AVX2. If ASTCENC_ISA_AVX2 fails to compile user must
 # choose another x86_64 option.
 
-if(NOT ${universal_build})
+if(APPLE_MAC_OS_UNIVERSAL)
+    if(APPLE_MAC_OS_ARCH_x86_64h)
+        list(APPEND ASTCENC_LIB_TARGETS astcenc-avx2-static)
+    elseif(APPLE_MAC_OS_ARCH_x86_64 OR APPLE_MAC_OS_ARCH_STANDARD)
```
MarkCallow: By using APPLE_MAC_OS_ARCH_STANDARD aren't you enshrining the problem you mentioned somewhere else of failing to handle a new architecture added to $(ARCH_STANDARD)?

lib/CMakeLists.txt
```diff
@@ -684,29 +702,69 @@ set(ASTCENC_CLI OFF) # Only build as library not the CLI astcencoder
...
+    # emits BASISU_SUPPORT_SSE=0 PUBLIC and no -msse4.1; we then use clang's
+    # -Xarch_<arch> driver option to inject -msse4.1 and override
+    # BASISU_SUPPORT_SSE to 1 only for the x86_64 sub-invocation. -Xarch_x86_64
+    # applies to the immediately-following argument only, so each flag needs its
+    # own prefix. Works with any generator (Xcode, Ninja, Make) on Apple.
+    if(APPLE_MAC_OS_UNIVERSAL AND APPLE_MAC_OS_ARCH_x86_64)
+        # SHELL: prevents CMake from deduplicating the repeated -Xarch_x86_64
+        # tokens, which would otherwise leave only the first flag arch-qualified.
+        target_compile_options(basisu_encoder PRIVATE
+            "SHELL:-Xarch_x86_64 -msse4.1"
```
MarkCallow, line 793: How/where do these get set for a build that is just for x86_64 or x86_64h?

lib/CMakeLists.txt
```diff
@@ -684,29 +702,69 @@ set(ASTCENC_CLI OFF) # Only build as library not the CLI astcencoder
...
+            "SHELL:-Xarch_x86_64 -UBASISU_SUPPORT_SSE"
+            "SHELL:-Xarch_x86_64 -DBASISU_SUPPORT_SSE=1"
+        )
+        foreach(t ktx ktx_read)
```
MarkCallow, line 797: Is this really necessary for ktx_read? The only occurrences of BASISU_SUPPORT_SSE that I see are in ktxTexture2_CompressBasisEx which should not be in the read only version of the library.

MarkCallow (commented):
Looks okay but I have a lot of questions. Please answer them.

MarkCallow:
I think you are correct about the `-msse4.1` setting at line 495. I think it is historical from when we directly included the basisu_encoder sources in libktx. Please remove it.

I originally introduced set_target_processor_type to determine whether the BASISU_SSE option should be shown or not, to remove a possible point of confusion for users. It wasn't connected to universal builds. I still want to avoid showing this option when it is of no use. I had hoped it would be temporary until basisu_encoder was changed to use compiler pre-defined macros and run-time queries to determine if SSE could be used and what variant. It doesn't look as if that is going to happen.

They key here is to not show the BASISU_SSE option, now LIBKTX_FEATURE_BASISU_SSE, if the target processor does not support it.

cmake/cputypetest.cmake
atteneder: Will do.

And agreed, that would be nice.

That got me thinking and...what if we substitute the literal with hard-coded values and add a custom pre-build command to check if it (still) matches up:

```cmake
if(ADD_XCODE_MACOS_ARCHS_STANDARD_CHECK)
        add_custom_command(TARGET ${target} PRE_BUILD
            COMMAND /bin/sh -c [[
                expected="arm64 x86_64"
                actual="$ARCHS_STANDARD"
                if [ "$actual" != "$expected" ]; then
                    echo "ARCHS_STANDARD mismatch: got [$actual], expected [$expected]" >&2
                    exit 1
                else
                    echo "ARCHS_STANDARD check passed: [$actual]"
                fi
            ]]
            VERBATIM
        )
    endif()
```

I tested it and it worked (with both `ONLY_ACTIVE_ARCH=NO` and `YES`), so I'll add it.

MarkCallow: Will this work on Windows? If not, you can use `cmake -P` with cmake code to do the same check.

lib/CMakeLists.txt
atteneder: This used to fail if you're cross-compiling on macOS targeting Android for example.

Explicitly filtering for "Darwin" is safer.

I noticed it introduced a bug, since CMAKE_SYSTEM_NAME is not necessarily set at all, so I updated the condition to assume macOS as target when it's not set.

lib/CMakeLists.txt
atteneder: Valid point. I guess linking the ASTC encoders in `ktx_read` didn't make sense before either and I just carried over the mistake.
Will check if it works.

cmake/cputypetest.cmake
atteneder: I'd only add this on the conditions macOS target and Xcode generator, so it wouldn't affect other platforms.

How would that `cmake -P` solution look like? Maybe it's more elegant than this.

lib/CMakeLists.txt
atteneder: WOMM!

atteneder: It didn't link successfully for Android since `astc_codec.cpp` references those libs, even with `KTX_FEATURE_WRITE` disabled.

cmake/cputypetest.cmake
MarkCallow: You write the script in the CMake language and launch it with `cmake -P`. Or on Linux and macOS you can add `#! /usr/bin/env cmake -P` at the top and run it like any other script or program. For an example, see https://github.com/KhronosGroup/KTX-Software/blob/93e59998c9938577d33a6f31519a775fa3e4929b/cmake/launch.cmake#L1-L25

MarkCallow: Hmm? I don't think that `cmake -P` has an equivalent to the shells' `-c`.

lib/CMakeLists.txt
MarkCallow: Ahh! Yes. We would probably need to build decoder only and full versions of the the ASTC libraries to avoid setting`-msse4.1` for _ktx_read_. I decided it wasn't worth the extra build time as it is a static library; the linker should extract from it what the using dylib or application needs. As imminent work is likely to remove the ARM astc-encoder in favour of the the one now in basis-universal, I do not want to spend time on improvements.

What is "WOMM"?

atteneder: > What is "WOMM"?
Acronym for "works on my machine".

Thanks for the feedback thus far. I'm off duty today and will try to continue tomorrow.

cmake/cputypetest.cmake
atteneder: I think I understand. But since this is conditioned to builds targeting macOS only I think it's safe like that.

lib/CMakeLists.txt
atteneder: I believe this was a "find a replace" mistake. Will undo.

lib/CMakeLists.txt
atteneder: Good observation. I'll make sure not to put `APPLE_MAC_OS_ARCH_STANDARD` directly in conditions anywhere.

lib/CMakeLists.txt
atteneder: Implicitly (or as originally intended) by `external/basis_universal/CMakeLists.txt`, since it that case `BASISU_SSE` is not overwritten to `FALSE`.

The idea is that for macOS universal only we disable `BASISU_SSE`, but manually inject the required definitions and compiler with the prefix `-Xarch_x86_64`, which limits it to the architectures capable of SSE only.

cmake/cputypetest.cmake
MarkCallow: Good point. `sh` is safe here.

lib/CMakeLists.txt
atteneder: Okay, that sounds good.

I've reverted the attempt to not have `ktx_read` depend on ASTC encoder.

atteneder:
> I think you are correct about the `-msse4.1` setting at line 495. I think it is historical from when we directly included the basisu_encoder sources in libktx. Please remove it.

Will do.

> I originally introduced set_target_processor_type to determine whether the BASISU_SSE option should be shown or not, to remove a possible point of confusion for users. It wasn't connected to universal builds. I still want to avoid showing this option when it is of no use. I had hoped it would be temporary until basisu_encoder was changed to use compiler pre-defined macros and run-time queries to determine if SSE could be used and what variant. It doesn't look as if that is going to happen.
> 
> They key here is to not show the BASISU_SSE option, now LIBKTX_FEATURE_BASISU_SSE, if the target processor does not support it.

Thanks for explaining.

As far as I can see after my changes it's still used for:

1. Failing 32-bit builds (presumably excluding unsupported windows, linux x86)
2. Driving visibility of `LIBKTX_FEATURE_BASISU_SSE`
3. Disable BasisU SSE for non x86_64
4. Driving `ASTCENC_ISA_*` for non-macOS-Universal builds.

I guess it's worth keeping, at the very least for point 4 and 3. I just saw some level of redundancy in arch detection across `CMakeLists.txt` and `lib/CMakeLists.txt` and thought it's worth considering addressing it.

lib/CMakeLists.txt
```diff
@@ -509,13 +540,41 @@ macro(add_lib_dependencies target lib)
...
 set(BASISU_TOOL FALSE)
 set(BASISU_EXAMPLES FALSE)
-if(NOT ${CPU_ARCHITECTURE} STREQUAL "x86_64")
+# Force BASISU_SSE off when basisu's compile would otherwise apply -msse4.1
+# unconditionally to a slice that can't take it. That is true for
+# non-x86_64 single-arch builds (obvious) and for any multi-arch macOS
+# universal build (the arm64 slice would fail). For universal builds the
+# x86_64 slice still gets SSE re-enabled per-arch via -Xarch_x86_64 in the
+# block further down this file, so we lose no functionality here.
+if(APPLE_MAC_OS_UNIVERSAL OR NOT "${CPU_ARCHITECTURE}" STREQUAL "x86_64")
```
MarkCallow: I think the second part of this if is redundant because of line 131. The dependent option LIBKTX_FEATURE_BASISU_SSE will be OFF (FALSE) if architecture is x86_64.

MarkCallow:
> 1. Failing 32-bit builds (presumably excluding unsupported windows, linux x86)

This came after `set_processor_architecture_type` was first introduced.

MarkCallow:
Please note that I will not merge this until after v5.0.0 is released which I hope will be RSN.

cmake/cputypetest.cmake
```diff
@@ -109,51 +109,37 @@ function(set_target_processor_type out)
...
-            # When building for macOS (and possibly for iOS, not tested) the
-            # user can specify CMAKE_OSX_ARCHITECTURES if they want to compile
-            # for something different than the current processor. This could be
-            # set to $(ARCHS_STANDARD) an Xcode build setting whose value in
-            # recent Xcode versions is "x86_64 arm64" to create universal binaries.
-            # This is passed to Xcode which calls the compiler twice passing the
-            # same set of target_definitions each time so in this case we have
-            # to choose BASISU_SUPPORT_SSE=OFF or the arm64 compile will fail.
+            # For macOS the caller may set CMAKE_OSX_ARCHITECTURES to a single
+            # arch (e.g. "x86_64" or "arm64") or to a list ("arm64;x86_64")/to
```
MarkCallow: s#or to a list ("arm64;x86_64")/to#, to a list ("arm64;x86_64") or to#

lib/CMakeLists.txt
atteneder: True. I'll remove that and move the comment within the if condition, where it makes more sense.

cmake/cputypetest.cmake
atteneder: good catch!

atteneder:
> Please note that I will not merge this until after v5.0.0 is released which I hope will be RSN.

Understood. Good luck for the release and thanks for all the great work!

walcht:
I don't know if this is the same issue or not, but Intel-based MacOS CIs for OpenImageIO fail for me (see here: https://github.com/walcht/OpenImageIO/actions/runs/28403033763/job/84158793776). Context: adding KTX2 to OpenImageIO and the CIs for Intel-based MacOS's are failing. Thank you @atteneder since this PR will probably address that.

walcht:
(In case this is relevant) I tried to test this PR to see if the failing CI on Intel-MacOS runners are fixed (see my comment above) and encountered some ASTCENC linkage issues (can be seen exactly here: https://github.com/walcht/OpenImageIO/actions/runs/28750777343/job/85249361051#step:14:667).

E.g. (from log link above):

```text
Undefined symbols for architecture x86_64:
  "astcenc_config_init(astcenc_profile, unsigned int, unsigned int, unsigned int, float, unsigned int, astcenc_config*)", referenced from:
      _ktxTexture2_DecodeAstc in libktx.a[x86_64][35](astc_codec.cpp.o)
      _ktxTexture2_CompressAstcEx in libktx.a[x86_64][35](astc_codec.cpp.o)
```

I haven't set any CMake flags/variables (used default config set by this PR's CMakeLists.txt).

walcht:
My comment above about the linkage issue is unrelated to this PR. I solved it and the issue is detailed in here: https://github.com/KhronosGroup/KTX-Software/issues/1210 (it's about the default ISA that is picked when none is provided).

Issue #1173, opened 2026-05-19
ktx create: crash with invalid pointer message when given the format R16G16B16_SFLOAT


walcht:
To reproduce:

1. Get the GoldenGate.exr RGB (**tiled OpenEXR image**) image from official OpenEXR test files: https://openexr.com/en/latest/test_images/Tiles/GoldenGate.html
1. Then run:
   ```bash
   ktx create --format R16G16B16_SFLOAT --assign-tf linear --assign-primaries bt709 GoldenGate.exr 2d_rgb16_linear.ktx2
   ```
   Or encode to UASTC (both commands output same error):
   ```bash
   ktx create --encode uastc-hdr-4x4 --format R16G16B16_SFLOAT --assign-tf linear --assign-primaries bt709 GoldenGate.exr 2d_uastc_hdr_4x4.ktx2
   ```
1. You might get something like this (I always get this on my system):
   ```text
   munmap_chunk(): invalid pointer
   Aborted (core dumped)
   ```
1. You can also get another non-tiled OpenEXR image (with or without an alpha channel) and this will work just fine (e.g., https://openexr.com/en/latest/test_images/ScanLines/CandleGlass.html):
   ```bash
   ktx create --format R16G16B16_SFLOAT --assign-tf linear --assign-primaries bt709 CandleGlass.exr 2d_rgb16_linear.ktx2
   ```

**Edit: I have failed to notice that this is a tiled image. Still, ktx create should gracefully exit if it doesn't support tiled OpenEXR images.**

<s>Even if this format is not supported, error'ing via an invalid pointer crash should definitely not happen (at least an early exit error of "this format is not supported" is significantly better).</s>

The crash is caused by a call to free the texture's data `free(This->pData)` which means that This->pData is overwritten somewhere with an address to a data that isn't allocated by malloc or the like functions. Haven't looked into details yet...

Tested on latest commit on branch main:

- commit hash: `1dacc4bc51b744070b14e3e10afa662ed328e57c`
- system: Ubuntu 24.04.4 LTS (x86)

MarkCallow:
_ktx create_ does support tiled images. This was an error in handling tiles at the edges of the image which will have data not in the image, unless `width % tile_size_x = height % tile_size_y = 0`.

[...]

PR #1167, opened 2026-05-07
Add BCn encoders/decoders with RDO support

walcht:
As discussed in #1159: `ktxTexture2_CompressBCn` and `ktxTexture2_DecodeBCn` are introduced in this PR to allow libktx users/consumers to encode/decode BCn textures from/to raw decompressed formats.

https://github.com/richgel999/bc7enc_rdo does not support BC6HU/BC6HS encoding/decoding and also no BC2 (this format is essentially dead since BC3 replaces it).

Please feel free to give feedback, edit, and nitpick as much as possible.

Some context: I am adding KTX2 support to OIIO (PR: https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/5185) and having libktx encode/decode BCn formats significantly simplifies things (also ETC encoding/decoding which I can also open a PR for - if approved).

I haven't updated the KTX-Software-CTS with the added BCn test files.

Once this is finalized, this will fix #587.

Current TODOs:

- [x] add CTS test files from a primary platform. As far as I understand, I shouldn't create/transcode KTX2 test files on non-primary platfroms (non arm64). I did do the testing on my local machine (see https://github.com/walcht/KTX-Software-CTS). => added CTS tests + golden files (still not on a primary platform).
- [x] add RDO post processing step
  - [x] expose RDO parameters with detailed description (some parameters are still missing - I haven't understood them yet)
  - [x] BC1 mode RDO
  - [x] BC2 mode RDO
  - [x] BC3 mode RDO
  - [x] BC4 mode RDO
  - [x] BC5 mode RDO
  - [x] <s>BC6H mode RDO (HDR). I don't know if this is possible. If not, then I will not adjust the ert::reduce_entropy function for the moment.</s> postponed (too much work/overload to include in this PR).
  - [x] BC7 mode RDO
  - [x] ultra smooth mode (see: https://richg42.blogspot.com/2021/02/updated-bc7encrdo-with-improved-smooth.html). => trivial to implement, just have to remove dependency on image_u8
- [x] add multithreading for encoder part (up to 9X decrease in time for my 6-core system)
- [x] add multithreading for RDO (bc7enc_rdo makes use of OpenMP so we need to do our own MT same as in encoder - this is trivial) => **5.50X to 7.00X time decrease compared to single-threaded mode**.
- [x] add BCn encoding support for `ktx encode` command
- [x] add BCn decoding to PNG support for `ktx extract` command
- [x] add BCn encoding support for `ktx create` command
[...]

PR #1176, opened 2026-05-25
Fix ASTC Decode_astc_8x8_unorm_array_7 testcase segfault

walcht:
As partially discussed here: https://github.com/KhronosGroup/KTX-Software/issues/1173.

I skimmed through some tests and by grepping for "KTX_SUCCESS" all checks almost always use EXPECT_XX which should, probably, be changed to ASSERT_XX (because why proceed to a segfault/other failure when a core Ktx function returned an exit error code?).


MarkCallow:
After I merge this, you will no longer be a "first-time contributor" so pushes thereafter to PR #1167 will start a CI run. Is #1167 ready for that?

MarkCallow:
> I skimmed through some tests and by grepping for "KTX_SUCCESS" all checks almost always use EXPECT_XX which should, probably, be changed to ASSERT_XX (because why proceed to a segfault/other failure when a core Ktx function returned an exit error code?).

Only the KTX_SUCCESS checks on functions that will return a `nullptr` on failure, e.g. `ktxTexture?_Create*`, need to be changed. For those cases where the following nullptr check outputs a message, the message must be moved to the KTX_SUCCESS check.

Please make these changes, if you have time.

walcht:
So I looked at other KTX_SUCCESS checks, and they all seem to make sense. Some are:

```c++
EXPECT_EQ(result, KTX_SUCCESS);
ASSERT_TRUE(texture != NULL) << "ktxTexture?_Create failed: " << ktxErrorString(result);
```

While others:

```c++
ASSERT_TRUE(result == KTX_SUCCESS);
ASSERT_TRUE(texture != NULL) << "ktxTexture_CreateFromMemory failed: " << ktxErrorString(result);
```

Some others do not assert on `texture != nullptr` but have reasonable if statements afterwards:
```c++
if (texture) { /* other stuff/tests */ }
```

So I just changed the single check that is causing the issue without touching anything else.

> Is https://github.com/KhronosGroup/KTX-Software/pull/1167 ready for that?

Finally, yes.

MarkCallow:
> ```c++
> ASSERT_TRUE(result == KTX_SUCCESS);
> ASSERT_TRUE(texture != NULL) << "ktxTexture_CreateFromMemory failed: " << ktxErrorString(result);
> ```

IIUC, in this case the test will stop if result != KTX_SUCCESS which means the error string for `result` will not be printed. Please change these to `EXPECT_EQ`.

MarkCallow:
Please merge the latest main which fixes the cloudflare-related Windows CI failure.

MarkCallow:
Please merge latest main here and in PR #1167.

MarkCallow:
I'll come back in an hour and if you've done it, I'll run the builds.

PR #1190, opened 2026-06-19
Record ASTC options in ktx encode

### LINKED ISSUE #1188: ASTC encoder not recording ASTC options in KTXwriterScParams
In `ktx encode`, ASTC encoder is not recording its options in KTXwriterScParams.

See missing `options.astcOptions` in: https://github.com/KhronosGroup/KTX-Software/blob/b0ad39f54e4432b425fce2f0a1a7bd9f2d577325/tools/ktx/command_encode.cpp#L332

Since this will require re-generation of CTS files, I'm not comfortable regenerating these on my machine.
It would also be nice if this only gets solved after merging the BCn PR (since it will cause conflicts with CTS files).

walcht:
Fixes #1188.

### DIFF
--- a/tools/ktx/command_encode.cpp
+++ b/tools/ktx/command_encode.cpp
@@ -329,7 +329,7 @@ void CommandEncode::executeEncode() {
     }
 
     // Add KTXwriterScParams metadata
-    const auto writerScParams = fmt::format("{}{}{}", options.codecOptions, options.commonOptions, options.compressOptions);
+    const auto writerScParams = fmt::format("{}{}{}{}", options.astcOptions, options.codecOptions, options.commonOptions, options.compressOptions);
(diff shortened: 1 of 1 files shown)


MarkCallow:
I have regenerated the golden files for this test and made PR https://github.com/KhronosGroup/KTX-Software-CTS/pull/76. Please update the CTS ref. here to https://github.com/KhronosGroup/KTX-Software-CTS/pull/76/changes/06ae7b6f47230d9797af3b73bc1ae45d59afc14f. The CI should then pass.

I would like to merge this before the BCn encoder. Your other two PRs in KTX-Software-CTS, @walcht change a different set of test files so there will be no conflicts. I'd like to include this in the upcoming v5.0.0 while I will not be including the BCn encoder until after that is released.

PR #1197, opened 2026-06-30
Export some missing functions

tuket:
This PR exports some missing functions in ktx.h.

You could call those functions though the macros that do the vtable dispatch. However, I have found that those macros are not suitable for creating C# bindings.\
Yes, you can call find those function pointers manually from C# and perform the calls: I have done that and it works in Windows. The problem comes when you compile for other platforms such as iOS and Wasm with AOT (ahead of time compilation). When compiling with AOT calling function pointers from C# will crash.


MarkCallow:
Thank you for this.

PR #1203, opened 2026-07-05
Fix memory leak in ktxHashList_DeleteKVPair/Entry functions

### LINKED ISSUE #1202: Memory leak in ktxHashList_DeleteKVPair and ktxHashList_DeleteEntry
Not a false positive. Memory is definitely leaked when using `ktxHashList_DeleteKVPair` or `ktxHashList_DeleteEntry`.

I noticed that when you run any trivial ktx writer (e.g., `ktx create`) (make sure to do a Debug build) through Valgrind it detects a memory leak:

```bash
cd tests/
valgrind --leak-check=yes ktx create --format R8G8B8A8_SRGB --assign-tf srgb resources/input/png/blue16.png foo.ktx2
```

I ignored it since it didn't hinder my development at all and doing proper address sanitation requires re-building libc++ and/or glibc with certain flags (which I didn't do). However, since I'm adding ktx2 to OpenImageIO (OIIO), its CIs have detected the exact same leak (see: https://github.com/walcht/OpenImageIO/actions/runs/28750777343/job/85249361061)

Why this is happening?

In both `ktxHashList_DeleteKVPair` and `ktxHashList_DeleteEntry`, the `ktxKVListEntry` is not freed (in `ktxHashList_Destruct` it is properly freed).

Patch that fixes this:

```patch
diff --git a/lib/src/hashlist.c b/lib/src/hashlist.c
index e34206ce..126aeb67 100644
--- a/lib/src/hashlist.c
+++ b/lib/src/hashlist.c
@@ -250,8 +250,10 @@ ktxHashList_DeleteKVPair(ktxHashList* pHead, const char* key)
         ktxKVListEntry* kv;
 
         HASH_FIND_STR( *pHead, key, kv );  /* kv: pointer to target entry. */
-        if (kv != NULL)
+        if (kv != NULL) {
             HASH_DEL(*pHead, kv);
+            free(kv);
+        }
         return KTX_SUCCESS;
     } else
         return KTX_INVALID_VALUE;
@@ -275,6 +277,7 @@ ktxHashList_DeleteEntry(ktxHashList* pHead, ktxHashListEntry* pEntry)
 {
     if (pHead && pEntry) {
         HASH_DEL(*pHead, pEntry);
+        free(pEntry);
         return KTX_SUCCESS;
     } else
         return KTX_INVALID_VALUE;
```

walcht:
Fixes #1202.

### DIFF
--- a/lib/src/hashlist.c
+++ b/lib/src/hashlist.c
@@ -250,8 +250,10 @@ ktxHashList_DeleteKVPair(ktxHashList* pHead, const char* key)
         ktxKVListEntry* kv;
 
         HASH_FIND_STR( *pHead, key, kv );  /* kv: pointer to target entry. */
-        if (kv != NULL)
+        if (kv != NULL) {
             HASH_DEL(*pHead, kv);
+            free(kv);
+        }
         return KTX_SUCCESS;
     } else
         return KTX_INVALID_VALUE;
@@ -275,6 +277,7 @@ ktxHashList_DeleteEntry(ktxHashList* pHead, ktxHashListEntry* pEntry)
 {
     if (pHead && pEntry) {
         HASH_DEL(*pHead, pEntry);
+        free(pEntry);
         return KTX_SUCCESS;
     } else
         return KTX_INVALID_VALUE;


MarkCallow:
Thank you for this. Great find.

PR #1205, opened 2026-07-06
Fix memory issues reported by asan

### LINKED ISSUE #1204: Memory issues (leaks, overflow access) after running AddressSanitizer alongside tests
Context: I decided to run AdressSanitizer when running tests to figure if my BCn PR introduced any leaks (see below on CMakeLists.txt patch to replicate how I did it).

I found a dozen or so of issues (mostly memory leaks in test files but there are a couple in libktx which I think are the most important to access). <s>I will just include the patches here.</s> => will wait on whether to do a single PR or multiple ones.

To replicate, apply this patch to topmost CMakeLists:

```patch
diff --git a/CMakeLists.txt b/CMakeLists.txt
index f85d63fa..3c6bf9c8 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,6 +47,11 @@ option( KTX_FEATURE_TOOLS_CTS "Enable KTX CLI Tools CTS tests (requires CTS subm
 option( KTX_FEATURE_JS "Enable JavaScript bindings in Emscripten builds." ON )
 option( KTX_FEATURE_TOOLS "Enable KTX CLI Tools." OFF)
 
+# Enabling sanitizers will cause significantly slow downs (noticeably, tests become significantly slower)
+set(SANITIZE "" CACHE STRING "Build code using sanitizer (address, thread, undefined)")
+
 if(KTX_FEATURE_TOOLS_CTS AND NOT KTX_FEATURE_TOOLS)
     message(WARNING "KTX_FEATURE_TOOLS is not set -> disabling KTX_FEATURE_TOOLS_CTS.")
     set(KTX_FEATURE_TOOLS_CTS "OFF")
@@ -190,6 +195,22 @@ elseif(${CMAKE_CXX_COMPILER_ID} STREQUAL "GNU"
     else()
         add_link_options( $<IF:$<CONFIG:Debug>,-g,-O3> )
     endif()
+    # Optional sanitizer (only Debug builds)
+    if (SANITIZE AND CMAKE_BUILD_TYPE STREQUAL "Debug")
+        message (STATUS "Compiling for sanitizer=${SANITIZE}")
+        string (REPLACE "," ";" SANITIZE_FEATURE_LIST ${SANITIZE})
+        foreach (feature ${SANITIZE_FEATURE_LIST})
+            message (STATUS "  sanitize feature: ${feature}")
+            add_compile_options (-fsanitize=${feature})
+            add_link_options (-fsanitize=${feature})
+        endforeach()
+        if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+            # finds writes past size of std::vector (won't be detected without this)
+            add_compile_definitions (_GLIBCXX_SANITIZE_VECTOR=1)
+        endif ()
+        # This has to be added to all (i.e., not using the target_compile_options variant)
+        add_compile_options (-fno-omit-frame-pointer -fPIC)
+    endif()
 else()
     message(FATAL_ERROR "${CMAKE_CXX_COMPILER_ID} not yet supported.")
 endif()
```

Then run CMake with the following flag: `cmake <usual-options> -DCMAKE_BUILD_TYPE=Debug -DSANITIZE="address"`

I'm almost done with verifying all of these and the absolute majority are not false positives.

List of files where I applied patches to address theses issues:
```
	modified:   external/dfdutils/KHR/khr_df.h => nothing wrong here, just added a new Macro. Should probably be done somewhere else.
	modified:   lib/src/basis_encode.cpp => alloc-dealloc-mismatch
	modified:   lib/src/texture.c => missing cleanup on `KTX_INVALID_VALUE`
	modified:   lib/src/texture2.c => dfd overflow access with some non-valid KTX2 files (which should never occur regardless of the input)
	modified:   tests/ktxdiff/ktxdiff_main.cpp => obvious mistake (see below)
	modified:   tests/texturetests/texturetests.cc => a lot of missing resource cleanups in here
	modified:   tests/transcodetests/transcodetests.cc
	modified:   tests/unittests/unittests.cc
	modified:   tests/unittests/wthelper.h
	modified:   tools/ktx/utility.h => move assignment bad implementation
```

PR is ready and I just wanted to ask:

1. Include all of these in single PR or separate PR or PR for tests, PR for libktx, etc.?
1. Whether it would make sense to add a CI where tests are run through an AddressSanitizer? (same as how OpenImageOI does) => this will definitely result in better security/memory issues checks but comes at the inconvenience of tests running at least 10x slower.

<s>Below is the list of diffs to fix these memory issues</s> => turns out there are quite a bit => see PR:
E.g.:
1. ktxdiff destructor does not properly free up the allocated KTX2 texture:

```patch
diff --git a/tests/ktxdiff/ktxdiff_main.cpp b/tests/ktxdiff/ktxdiff_main.cpp
index bc81718e..8d3bf540 100644
--- a/tests/ktxdiff/ktxdiff_main.cpp
+++ b/tests/ktxdiff/ktxdiff_main.cpp
@@ -84,7 +84,7 @@ public:
         loadMetadata();
     }
     ~Texture() {
-        std::free(handle);
+        if (handle) ktxTexture2_Destroy(handle);
     }
     void loadFile();
     void loadKTX();
```

```patch
diff --git a/tools/ktx/utility.h b/tools/ktx/utility.h
index 75b1ab22..b9f525d8 100644
--- a/tools/ktx/utility.h
+++ b/tools/ktx/utility.h
@@ -352,7 +352,17 @@ public:
         other.handle_ = nullptr;
     }
 
-    KTXTexture2& operator=(KTXTexture2&& other) & {
+    KTXTexture2& operator=(KTXTexture2&& other) noexcept {
+        // If self assignement => don't do anything
+        if (&other == this) {
+          return *this;
+        }
+        // Release our ktxTexture resource
+        if (handle_ != nullptr) {
+            ktxTexture_Destroy(ktxTexture(handle_));
+            handle_ = nullptr;
+        }
+        // Take ownership of other's ktxTexture resource
         handle_ = other.handle_;
         other.handle_ = nullptr;
         return *this;
```

```patch
diff --git a/lib/src/basis_encode.cpp b/lib/src/basis_encode.cpp
index 76eac83b..949a4944 100644
--- a/lib/src/basis_encode.cpp
+++ b/lib/src/basis_encode.cpp
@@ -1157,7 +1157,8 @@ ktxTexture2_CompressBasisEx(ktxTexture2* This, ktxBasisParams* params)
                  + image_desc_size * num_images
                  + bfh.m_endpoint_cb_file_size + bfh.m_selector_cb_file_size
                  + bfh.m_tables_file_size;
-        bgd = new ktx_uint8_t[bgd_size];
+        // Don't use new[] as this will cause alloc-dealloc-mismatch when This texture is freed
+        bgd = (ktx_uint8_t*)malloc(bgd_size);
         ktxBasisLzGlobalHeader& bgdh = *reinterpret_cast<ktxBasisLzGlobalHeader*>(bgd);
         bgdh.endpointCount = (uint16_t)bfh.m_total_endpoints;
         bgdh.endpointsByteLength = bfh.m_endpoint_cb_file_size;
```

```patch
diff --git a/lib/src/texture.c b/lib/src/texture.c
index 14abc418..247459a1 100644
--- a/lib/src/texture.c
+++ b/lib/src/texture.c
@@ -105,21 +105,21 @@ ktxTexture_construct(ktxTexture* This,
      * 1D, 2D, 3D, cube, and array variants of these.
      */
     if (createInfo->numDimensions < 1 || createInfo->numDimensions > 3)
-        return KTX_INVALID_VALUE;
+        goto error_invalid_value;
 
     if (createInfo->baseWidth == 0 || createInfo->baseHeight == 0
         || createInfo->baseDepth == 0)
-        return KTX_INVALID_VALUE;
+        goto error_invalid_value;
 
     switch (createInfo->numDimensions) {
       case 1:
         if (createInfo->baseHeight > 1 || createInfo->baseDepth > 1)
-            return KTX_INVALID_OPERATION;
+            goto error_invalid_operation;
         break;
 
       case 2:
         if (createInfo->baseDepth > 1)
-            return KTX_INVALID_OPERATION;
+            goto error_invalid_operation;
         break;
 
       case 3:
@@ -128,7 +128,7 @@ ktxTexture_construct(ktxTexture* This,
          */
         if (createInfo->isArray || createInfo->numFaces != 1
             || createInfo->numLayers != 1)
-            return KTX_INVALID_OPERATION;
+            goto error_invalid_operation;
         break;
     }
     This->numDimensions = createInfo->numDimensions;
@@ -137,29 +137,29 @@ ktxTexture_construct(ktxTexture* This,
     This->baseHeight = createInfo->baseHeight;
 
     if (createInfo->numLayers == 0)
-        return KTX_INVALID_VALUE;
+        goto error_invalid_value;
     This->numLayers = createInfo->numLayers;
     This->isArray = createInfo->isArray;
 
     if (createInfo->numFaces == 6) {
         if (This->numDimensions != 2) {
             /* cube map needs 2D faces */
-            return KTX_INVALID_OPERATION;
+            goto error_invalid_operation;
         }
         if (createInfo->baseWidth != createInfo->baseHeight) {
             /* cube maps require square images */
-            return KTX_INVALID_OPERATION;
+            goto error_invalid_operation;
         }
         This->isCubemap = KTX_TRUE;
     } else if (createInfo->numFaces != 1) {
         /* numFaces must be either 1 or 6 */
-        return KTX_INVALID_VALUE;
+        goto error_invalid_value;
     }
     This->numFaces = createInfo->numFaces;
 
     /* Check number of mipmap levels */
     if (createInfo->numLevels == 0)
-        return KTX_INVALID_VALUE;
+        goto error_invalid_value;
     This->numLevels = createInfo->numLevels;
     This->generateMipmaps = createInfo->generateMipmaps;
 
@@ -169,12 +169,20 @@ ktxTexture_construct(ktxTexture* This,
         if (max_dim < ((GLuint)1 << (This->numLevels - 1)))
         {
             /* Can't have more mip levels than 1 + log2(max(width, height, depth)) */
-            return KTX_INVALID_OPERATION;
+            goto error_invalid_operation;
         }
     }
 
     ktxHashList_Construct(&This->kvDataHead);
     return KTX_SUCCESS;
+
+error_invalid_value:
+    free(This->_protected);
+    return KTX_INVALID_VALUE;
+
+error_invalid_operation:
+    free(This->_protected);
+    return KTX_INVALID_OPERATION;
 }
 
 /**
```

This will be too long if I include the others ...

### LINKED ISSUE #1206: Linux CIs never run tests in Debug configuration (even though the logs state so)
While adding ASan CI to my PR https://github.com/KhronosGroup/KTX-Software/pull/1205, I noticed the following:

- In `scripts/build_linux.sh`, `-DCMAKE_BUILD_TYPE=$CONFIGURATION` is never passed to cmake arguments. This means that for single-configuration generators, cmake is invoked without  setting CMAKE_BUILD_TYPE (you can see it empty in the CI logs).

- Debug CIs with tools_tests ON are as fast as Release CIs with tools_tests ON (**Package (x86_64): Total Test time (real) = 413.70 sec** vs. **Default: Total Test time (real) = 234.89 sec**). Debug tests are significantly slower on my machine.

- See this line: https://github.com/KhronosGroup/KTX-Software/actions/runs/28780956339/job/85336156254#step:12:201
CMAKE_BUILD_TYPE is empty, so the default for the compiler toolchain is used.

After doing the fixes (in PR above), the numbers make more sense:

Package (x86_64) (Release): Total Test time (real) = 378.61 sec
Default (Debug): Total Test time (real) = 1384.77 sec

The Linux CIs still pass `--config Debug` (or Release) to the cmake build step, but that does nothing for single-configuration generators (most of the runners on Linux). Same for testing with `ctest -C Debug` => it does nothing for single-configuration generators.

Source of numbers above: https://github.com/KhronosGroup/KTX-Software/actions/runs/28780956339/job/85336156254

walcht:
Fixes #1204.

There are still a couple more issues (I think 4) that I found to be hard to fix.
I will add details about these issues if I can't solve them tonight.

@MarkCallow - Please don't merge this yet (I think I can clean it up more by using std::unique_ptr's when possible).

Hopefully adding this here states that this PR closes that issue as well...
Fixes #1206.


CMakeLists.txt
```diff
@@ -47,6 +47,9 @@ option( KTX_FEATURE_TOOLS_CTS "Enable KTX CLI Tools CTS tests (requires CTS subm
 option( KTX_FEATURE_JS "Enable JavaScript bindings in Emscripten builds." ON )
 option( KTX_FEATURE_TOOLS "Enable KTX CLI Tools." OFF)
 
+# Enabling sanitizers will cause significantly slow downs (noticeably, tests become significantly slower)
+set(SANITIZE "" CACHE STRING "Build code using sanitizer (address, thread, undefined)")
```
MarkCallow: I don't understand what the items in the parentheses means. Are those the things the sanitizer checks for?

Also the help string must say that it causes "significant slowdowns, especially in tests". There is probably no need to keep the comment once the info is in the help string.

lib/src/texture2.c
```diff
@@ -253,13 +253,20 @@ ktx_uint32_t e5b9g9r9_ufloat_comparator[e5b9g9r9_bdbwordcount] = {
 bool
 ktxFormatSize_initFromDfd(ktxFormatSize* This, ktx_uint32_t* pDfd)
 {
+    const uint32_t sizeof_dfdTotalSize = sizeof(uint32_t);
+    const uint32_t sizeof_DFDBHeader = sizeof(uint32_t) * 2;
```
MarkCallow: This is the value given by `KHR_DFD_SIZEFOR_DESCRIPTORBLOCKSIZE`. Prefer you use that macro or use `KHR_DF_WORD_DESCRIPTORBLOCKSIZE + 1` in initializing this variable. Also it should use `BDFD` not `DFDB`. It means Basic Data Format Descriptor.

lib/src/texture2.c
```diff
@@ -253,13 +253,20 @@ ktx_uint32_t e5b9g9r9_ufloat_comparator[e5b9g9r9_bdbwordcount] = {
...
+    const uint32_t sizeof_BDFD = sizeof(uint32_t) * KHR_DF_WORD_SAMPLESTART;
+    const uint32_t sizeof_BDFDSample = sizeof(uint32_t) * KHR_DF_WORD_SAMPLEWORDS;
+
     uint32_t* pBdb = pDfd + 1;
     // pDfd[0] contains totalSize in bytes, check if it has at least
     // KHR_DFD_SIZEFOR_DESCRIPTORBLOCKSIZE bytes
-    if (pDfd[0] < KHR_DFD_SIZEFOR_DESCRIPTORBLOCKSIZE || *pBdb != 0) {
+    // Added: '+ sizeof(uint32_t)' so that this aligns with printdfd.c, otherwise it doesn't make sense ...
```
MarkCallow: Omitting the `sizeof(pDfd[0])` here was indeed an error.

lib/src/texture2.c
```diff
@@ -284,11 +291,19 @@ ktxFormatSize_initFromDfd(ktxFormatSize* This, ktx_uint32_t* pDfd)
...
+    if (pDfd[0] < (sizeof_dfdTotalSize + sizeof_BDFD)) {
+        return false;  // Invalid DFD: Missing or partial basic DFD block
+    }
+
     // DFD has supported type and version. Process it.
     This->blockWidth = KHR_DFDVAL(pBdb, TEXELBLOCKDIMENSION0) + 1;
     This->blockHeight = KHR_DFDVAL(pBdb, TEXELBLOCKDIMENSION1) + 1;
     This->blockDepth = KHR_DFDVAL(pBdb, TEXELBLOCKDIMENSION2) + 1;
     if (KHR_DFDVAL(pBdb, BYTESPLANE0) == 0) {
+        if (pDfd[0] < (sizeof_dfdTotalSize + sizeof_BDFD + sizeof_BDFDSample)) {
```
MarkCallow: Move this test to line 298.

lib/src/texture2.c
```diff
@@ -311,7 +326,9 @@ ktxFormatSize_initFromDfd(ktxFormatSize* This, ktx_uint32_t* pDfd)
         }
     } else {
         // An uncompressed format.
-
+        if (pDfd[0] < (sizeof_BDFD + sizeof_BDFDSample + sizeof_dfdTotalSize)) {
```
MarkCallow: Delete this test. It is no longer needed once the test at line 304 is moved to line 298.

lib/src/texture2.c
```diff
@@ -785,6 +802,15 @@ ktxTexture2_constructFromStreamAndHeader(ktxTexture2* This, ktxStream* pStream,
         result = KTX_FILE_DATA_ERROR;
         goto cleanup;
     }
+    // TODO: shouldn't this be < sizeof(BDFD)?
+    // I added this check because:
+    //  - ASan reports "unknow-crash" on 'pBDFD->descriptorBlockSize'
+    //  - what if we allocated less than 8 bytes for BDFD struct? (i.e., < 12 bytes total)
+    if ((pHeader->dataFormatDescriptor.byteLength - sizeof(ktx_uint32_t)) < 24) {
```
MarkCallow: I suggest

```
pHeader->dataFormatDescriptor.byteLength < (KHR_DF_WORD_SAMPLESTART + 1 + KHR_DF_SAMPLEWORDS) * sizeof(ktx_uint32_t
```

This allows for at least size of dfd size value (4), basic descriptor block (24) and one sample (16).

tests/ktxdiff/ktxdiff_main.cpp
```diff
@@ -83,7 +83,7 @@ struct Texture {
         loadMetadata();
     }
     ~Texture() {
-        std::free(handle);
+        if (handle) ktxTexture2_Destroy(handle);
```
MarkCallow: Look into using `std::unique_ptr` in whatever type `handle` is.

MarkCallow (commented):
10 more files to review but I decided to send this batch of comments to get you started.

lib/src/texture2.c
MarkCallow: Thinking about this more deeply, the use for this field is to for a file reader skip over the DFD, if it is not interested in its contents. The value comes from the file and is not used to allocate the DFD struct. The size in the first 4 bytes of the DFD is used for that. Therefore (a) I do not understand what ASan means by "unknown -crash" and (b) the most important thing to check here is that `pHeader->dataFormatDescriptor.byteOffset + pHeader->dataFormatDescriptor.byteLength` does not exceed the length of the file in case the reader is reading from a file in memory. If offset or length is incorrect but within the file data, it means the DFD info will be wrong which will be caught by the many other checks on the DFD data.

tests/texturetests/texturetests.cc
```diff
@@ -253,22 +253,26 @@ class ktxTextureTestBase : public ::testing::Test {
...
         if (fv == KTX_FORMAT_VERSION_ONE) {
            kvDataLen = helper.kvDataLen;
-           kvData = helper.kvData;
+           // kvData = helper.kvData;
+           kvData = (unsigned char*)malloc(kvDataLen);
+           memcpy(kvData, helper.kvData, kvDataLen);
            errorCode = ktxTexture1_Create(&texinfo,
                                            KTX_TEXTURE_CREATE_ALLOC_STORAGE,
-                                           (ktxTexture1**)&texture);
+                                           (ktxTexture1**)&texture_tmp);
```
MarkCallow: Why have you changed the variable name to `texture_tmp`?

tests/texturetests/texturetests.cc
```diff
@@ -308,7 +312,10 @@ class ktxTextureTestBase : public ::testing::Test {
     }
 
     ~ktxTextureTestBase() {
-        delete ktxMemFile;
+        free(ktxMemFile);
+        if (texture_tmp) {
```
MarkCallow: Where is kvData being freed? Is it?

tests/unittests/unittests.cc
```diff
@@ -1037,6 +1039,9 @@ class SwizzleTestBase : public ::testing::Test {
                    EXPECT_EQ(swizzle[c], dest[i][c]) << "c = " << c << ", i = "  << i;
            }
        }
+       free(dest);
+       if (texture)
+         ktxTexture2_Destroy(texture);
```
MarkCallow: Indent by 2 more spaces.

tools/ktx/command_convert.cpp
```diff
@@ -308,8 +308,10 @@ void CommandConvert::convertKtx(InputStream& inputStream, OutputStreamEx& output
                } else {
                    warning("Dropping unrecognized KTX metadata \"{}\"", key);
                }
+               // TODO: shouldn't we save the next pointer BEFORE deleting the entry?
```
MarkCallow: Yes. I think so though, as it is unlikely a KTX v1 input file will have multiple KTXOrientation fields, the result will probably be the same as setting it to nullptr.

tools/ktx/command_compare.cpp
```diff
@@ -1787,30 +1787,43 @@ void CommandCompare::compareKVD(PrintDiff& diff, InputStreams& streams) {
     std::vector<uint8_t> keyValueStores[2] = {};
     std::map<std::string_view, ktxHashListEntry*> keys[2] = {};
 
+    // To rely on RAII to cleanup hashlist allocated in the loop below
+    std::vector<std::unique_ptr<ktxHashList, void (*)(ktxHashList*)>> hashlists;
```
MarkCallow, line 1791: I don't think there is any need to treat this vector differently than the one, and  the map, above. I.e make it a fixed size of 2. _ktx compare_ only supports comparing 2 files.

tools/ktx/command_compare.cpp
```diff
@@ -1787,30 +1787,43 @@ void CommandCompare::compareKVD(PrintDiff& diff, InputStreams& streams) {
...
         if (headers[i].keyValueData.byteLength == 0) continue;
 
         keyValueStores[i].resize(headers[i].keyValueData.byteLength);
         read(streams[i], headers[i].keyValueData.byteOffset, keyValueStores[i].data(),
             headers[i].keyValueData.byteLength, "the KVD");
 
-        ktxHashList kvDataHead = nullptr;
-        KTX_error_code result = ktxHashList_Deserialize(&kvDataHead,
+        // hashlists.emplace_back(), ktxHashList_Destroy);
+        // auto kvDataHead = hashlists[i].get();
```
MarkCallow: Please delete these comments. I find them confusing.

MarkCallow (commented):
Here are my comments on the rest of the files.

tools/ktx/command_compare.cpp
walcht: I couldn't do the usual `... hashlist[2] = {}`, tried supplying a default value but compiler complained (in arcane error messages) about it.

I did the usual `reserve(2)` then I kept the same emplace_back. I moved it up so that we don't have to call the `free`.

tests/texturetests/texturetests.cc
walcht: This one is very ugly.

There is memcpy above where we copy kvData from its `helper` to the texture.
This memcpy'ed kvData is free'd by the texture (in ktxTexture_Destroy).
The original kvData (in helper) is free'd by the the helper.

This was leaking before. This fix is ugly. I made it much cleaner now via the usage to std::unique_ptr's.

tests/texturetests/texturetests.cc
walcht: I cleaned this up (was very ugly before). Now every resource is managed via a unique_ptr if ownership is intended.

As to why I changed the name, I moved this to class member (it no longer is there, I cleaned this up) and Windows builds kept complaining about obscuring name.

CMakeLists.txt
walcht: Ah, these mean the values that can be assigned to SANITIZE. I clarified this in latest commit.

lib/src/texture2.c
walcht: Addressed in detail. See latest comment in main thread.

walcht:
Updates:
- I addressed all review commets.
- All memory issues reported by ASan are now fixed.
- Used smart pointers in texturetests instead of manual KtxTexture_Destroy/free/delete calls. Now it is guaranteed that all resources are properly de-allocated (of course, unless there are memory leaks in libktx or other calls etc.).
- I have done the same thing in unittests but I haven't pushed it yet. Do you agree that I should just use smart pointers there (mainly unique_ptrs) same as in `texturetests.cc`?
- Added `ASan GCC (x86_64)`
- A PR in `dfdutils` is also opened (a single line of code change that fixes buffer-overflow access with certain inputs)
- Fixed another leak in `writer1.c`

If CIs pass then this is now ready for merging. As usual, please do a second round of review especially in `dfd`-related changes since I am not very familiar with those (yet).

walcht:
Fixes #1206.

lib/src/texture2.c
```diff
@@ -324,11 +326,7 @@ ktxFormatSize_initFromDfd(ktxFormatSize* This, ktx_uint32_t* pDfd)
         if (KHR_DFDVAL(pBdb, MODEL) == KHR_DF_MODEL_PVRTC) {
             This->minBlocksX = This->minBlocksY = 2;
         }
-    } else {
-        // An uncompressed format.
-        if (pDfd[0] < (sizeof_BDFD + sizeof_BDFDSample + sizeof_dfdTotalSize)) {
-            return false;  // Invalid DFD: Missing or partial basic DFD sample
-        }
+    } else { // An uncompressed format.
```
MarkCallow: This comment is wrong and misleading. Please remove it. It was right when first written though still misleading (my fault). It should have been "non-supercompresed".

lib/src/texture2.c
```diff
@@ -802,15 +800,13 @@ ktxTexture2_constructFromStreamAndHeader(ktxTexture2* This, ktxStream* pStream,
...
-    // TODO: shouldn't this be < sizeof(BDFD)?
-    // I added this check because:
-    //  - ASan reports "unknow-crash" on 'pBDFD->descriptorBlockSize'
-    //  - what if we allocated less than 8 bytes for BDFD struct? (i.e., < 12 bytes total)
-    if ((pHeader->dataFormatDescriptor.byteLength - sizeof(ktx_uint32_t)) < 24) {
-        // BDFD has invalid size
-        result = KTX_FILE_DATA_ERROR;
+
+    // Before doing any access on pDfd, make use of the extensive checks in ktxFormatSize_initFromDfd
+    if (!ktxFormatSize_initFromDfd(&This->_protected->_formatSize, This->pDfd)) {
```
MarkCallow: This is pedantic but UNSUPPORTED_TEXTURE_TYPE is the wrong error type for faulty DFDs caught by the checks. Either separate those checks into another function that returns FILE_DATA_ERROR or change `ktxFormatSize_initFromDfd` to return a `ktx_error_code_e`. I can't remember why I made it a bool. Probably need to check for other callers and what they expect.

lib/src/writer1.c
```diff
@@ -958,6 +958,11 @@ ktxTexture1_WriteKTX2ToMemory(ktxTexture1* This,
     result = ktxTexture1_WriteKTX2ToStream(This, &dststr);
     if(result != KTX_SUCCESS)
     {
+        /* Make sure to free data on destruct in case of errors otherwise the
+         * caller will not be able to free the data because the stream pointer
+         * is not yet returned.
+         */
+        // dststr.closeOnDestruct = true;
```
MarkCallow: If this is not necessary, remove it and the comment above it.

scripts/build_linux.sh
```diff
@@ -128,6 +135,9 @@ for arg in "${cmake_args[@]}"; do
   esac
 done
 echo ${config_display%??}
+
+# Print cmake command to be able to verify configuration and replicate locally
+echo "running cmake command (in source directory): cmake . ${cmake_args[@]}"
```
MarkCallow, line 140: The config display provides the same information in a  more human friendly way. Why is this necessary.

CMakeLists.txt
```diff
@@ -47,8 +47,11 @@ option( KTX_FEATURE_TOOLS_CTS "Enable KTX CLI Tools CTS tests (requires CTS subm
 option( KTX_FEATURE_JS "Enable JavaScript bindings in Emscripten builds." ON )
 option( KTX_FEATURE_TOOLS "Enable KTX CLI Tools." OFF)
 
-# Enabling sanitizers will cause significantly slow downs (noticeably, tests become significantly slower)
-set(SANITIZE "" CACHE STRING "Build code using sanitizer (address, thread, undefined)")
+# SANITIZE can be set to the same set of values acceptable by '-fsanitize' by both Clang and GCC
+#   - to enable AddressSanitizer (Asan), set to "address"
+#   - to enable ThreadSanitizer (Tsan), set to "thread"
+#   - to enable UndefinedBehaviorSanitizer (UBSan), set to "undefined"
+set(SANITIZE "" CACHE STRING "Build code using sanitizer. Results in significantly slower runtime. Can be set to any value acceptable by '-fsanitize'")
```
MarkCallow: Please use a STRINGS property to make a list of acceptable strings for this variable. See line 77 in this file for an example for KTX_FEATURE_LOADTEST_APPS.

Do not show this option if the compiler is MSVC.

tests/streamtests/streamtests.cc
```diff
@@ -203,6 +203,9 @@ class StreambufStream
 
     static void destruct(ktxStream* str)
     {
+        // It could be that 'self' is no longer a valid address (e.g., ktxTexture outlives the instance that 'self' points to).
+        // This falls down to the following requirement: the destructor of any ktxTexture that references this object (e.g.,
+        // via ktxStream) SHOULD occur BEFORE the destructor of said object.
```
MarkCallow: I don't understand the comment or the problem.

1. The ktxTexture is expected to outlive the stream. It is only needed until the texture has been loaded or uploaded to a GPU. 
2. The parent here is the std::streambuf which is unrelated to the ktxTexture.

tests/texturetests/texturetests.cc
```diff
@@ -3025,12 +3071,9 @@ class ktxTexture2AstcLdrEncodeDecodeTestBase
             if (status != 0) {
                 std::cout << std::ifstream(ktxdiffOut).rdbuf();
             }
-            if (texture) {
-                ktxTexture_Destroy(ktxTexture(texture));
-                fs::remove(original);
-                fs::remove(decoded);
-                fs::remove(ktxdiffOut);
-            }
+            fs::remove(original);
```
MarkCallow: Does fs::remove quietly handle an attempt to remove a non-existent file?

MarkCallow (commented):
Here are my thoughts on the latest revision.

Why are you hesitating about posting the unittests changes to use unique_ptr? Are they even more extreme than in texturetests?

CMakeLists.txt
MarkCallow: Oh! A STRINGS property probably does not handle combinations of the strings. It is more like an option menu.

.github/workflows/linux.yml
```diff
@@ -67,6 +67,17 @@ jobs:
             doc: OFF, jni: ON, loadtests: OpenGL+Vulkan, py: ON, tools: ON, tools_cts: ON,
             sse: ON, opencl: OFF
           }
+        - name: ASan GCC (x86_64)
```
MarkCallow, line 73: I wanted to just add sanitize to one of the existing jobs but none seem suitable. The most likely candidate is "Default" but I expect adding the loadtests and python binding will cause sanitize to find a whole lot more issues. But I want to keep testing builds of those things on ubuntu-latest.

tests/texturetests/texturetests.cc
walcht: No. It throws an exception. I added `fs::exists` checks in latest commit.

.github/workflows/linux.yml
```diff
@@ -177,6 +188,9 @@ jobs:
 
       VULKAN_SDK_VER: ${{ matrix.vk_sdk_ver }}
 
+      # Only perform ASan/UBSan/TSan on linux runners because they slow down tests
```
MarkCallow: As this workflow is only for Linux and the following line is simply setting the env var from the job matrix, I find this comment confusing. Move it to the ASan GCC entry in the matrix and change it to

"We perform ASan/UBSan/TSan on only one job because they greatly slow down test runs. We use a linux runner because they consume the fewest GH Actions credits."

MarkCallow (commented):
Forgot to review linux.yml before.

lib/src/basis_encode.cpp
```diff
@@ -1298,6 +1298,8 @@ ktxTexture2_CompressBasisEx(ktxTexture2* This, ktxBasisParams* params)
         if (result != KTX_SUCCESS) goto cleanup;
 
         // Reflect this in the formatSize
+        // TODO: adding checks here causes numerous Basis tests to fail.
+        //       Why the return value isn't checked here?
```
MarkCallow: They must be affecting the content of the `formatSize` structure, maybe detecting a problem and returning early.

lib/src/texture2.c
```diff
@@ -358,7 +361,7 @@ ktxFormatSize_initFromDfd(ktxFormatSize* This, ktx_uint32_t* pDfd)
             result = interpretDFD(pDfd, &rgba[0], &rgba[1], &rgba[2], &rgba[3],
                                   &wordBytes);
             if (result >= i_UNSUPPORTED_ERROR_BIT)
-                return false;
+                return KTX_FILE_DATA_ERROR;
```
MarkCallow: This should be UNSUPPORTED_TEXTURE_TYPE. interpretDFD does not support all valid DFDs.

lib/src/texture2.c
```diff
@@ -464,8 +467,9 @@ ktxTexture2_construct(ktxTexture2* This,
 #ifdef _DEBUG
         // If this fires, an unsupported format or incorrect DFD
         // has crept into vk2dfd.
-        assert(ktxFormatSize_initFromDfd(&formatSize, This->pDfd));
+        assert(ktxFormatSize_initFromDfd(&formatSize, This->pDfd) == KTX_SUCCESS);
 #else
+        // TODO: why no check on return here? Is it guaranteed that ktxVk2dfd will always correctly construct the dfd?
```
MarkCallow: This function constructs a new ktxTexture2. The DFD is created by vk2dfd from the vkFormat specified in the createInfo parameter. As the comment under _DEBUG says the only potential cause of failure is a coding error in vk2dfd. One of the unittests does a roundtrip test of every Vulkan format vk to dfd to vk so we have a high level of confidence vk2Dfd is correct.

tests/streamtests/streamtests.cc
```diff
@@ -203,9 +203,15 @@ class StreambufStream
...
     static void destruct(ktxStream* str)
     {
-        // It could be that 'self' is no longer a valid address (e.g., ktxTexture outlives the instance that 'self' points to).
-        // This falls down to the following requirement: the destructor of any ktxTexture that references this object (e.g.,
-        // via ktxStream) SHOULD occur BEFORE the destructor of said object.
+        // It could be that 'self' is no longer a valid address (e.g., ktxTexture outlives the instance that 'self'
+        // points to). This falls down to the following requirement: the destructor of any ktxTexture that references
+        // this object (e.g., via ktxStream) SHOULD occur BEFORE the destructor of said object.
+        //
+        // More details:
```
MarkCallow: Thank for for the details. Is `ktxTexture_Destroy` not checking if the stream still exists before calling its destructor? If not, it definitely should be. If it is, then I am still not understanding what is happening here.

MarkCallow (commented):
More comments.

walcht:
> Thank for for the details. Is ktxTexture_Destroy not checking if the stream still exists before calling its destructor? If not, it definitely should be. If it is, then I am still not understanding what is happening here.

I'm not talking about the ktxStream but rather the wrapper around it in streamtests (i.e., `StreambufStream`). ktxTexture_Destroy gets supplied with a callback (the static function `destruct` in streamtests) and this function sets a bool member (_destruct) to false. It accesses said bool via the `self` pointer which points to the instance object **of StreambufStream** (again, not ktxStream but the wrapper around it). I don't think there is anything ktxTexture_Destroy can do or check for here. It was simply supplied with a callback that may perform an invalid pointer access.

I think if you run AddressSanitizer on this PR after reverting the changes in `streamtests` you will see the exact stacktrace and that might explain it better than I do here.

Again,  this one was a bit tricky to find.

> They must be affecting the content of the formatSize structure, maybe detecting a problem and returning early.

I was doing this check `if (result == KTX_SUCCESS) goto cleanup;` (rather than !=). It took me too long to notice it...

I have pushed unitttests.cc updates.

This now should be ready for merging (after you review unittests.cc).

Update: as usual, GitHub actions failure rate is impressive, re-run the CI (I'm certainly confident that all should pass).

tools/imageio/exr.imageio/exrinput.cc
```diff
@@ -11,11 +11,15 @@
...
 #include <string_view>
 #include <vector>
-// TEXR is not defined in tinyexr.h. Current GitHub tinyexr master uses
-// assert. The version in astc-encoder must be old.
-#define TEXR_ASSERT(x) assert(x)
 #define TINYEXR_IMPLEMENTATION
-#include "tinyexr.h"
+#define TEXR_ASSERT(x) assert(x)
+// Use tinyexr.h from basis_universal because it is self-contained. By default
+// tinyexr.h uses miniz but the astc-encoder has decided to use the std_zlib
```
MarkCallow: s/std/stb/

Also on line 19.

I don't know if some autocorrect caught me out here or simply failed to spot my error.

walcht:
@MarkCallow - Before merging, please notice that I updated the CTS submodule (see CTS submodule commit message for details on why).

MarkCallow:
> I'm not talking about the ktxStream but rather the wrapper around it in streamtests (i.e., `StreambufStream`). ktxTexture_Destroy gets supplied with a callback (the static function `destruct` in streamtests) and this function sets a bool member (_destruct) to false. It accesses said bool via the `self` pointer which points to the instance object **of StreambufStream** (again, not ktxStream but the wrapper around it). I don't think there is anything ktxTexture_Destroy can do or check for here. It was simply supplied with a callback that may perform an invalid pointer access.

We make StreambufStream available to developers via `utils/sbufstream.h` so I want to ensure it works without having to worry about the ordering of calls to `ktxTexture_Destroy` and deletion or destruction of StreambufStream, if possible.

Are the tests deleting the StreambufStream while the texture still exists? If so, it seems like what is missing is a way for `StreambufStream` to remove the callback when it is destroyed before the `ktxTexture*`. But why do the tests delete the `StreambufStream`? The `ktxTexture` functions dealing with loading data, destroy the `ktxStream` when they finish loading the data and the destruct functions do so when the `ktxTexture*` is being destructed, if it still exists.

walcht:
> Are the tests deleting the StreambufStream while the texture still exists?

Before my commits, yes.

> If so, it seems like what is missing is a way for StreambufStream to remove the callback when it is destroyed before the ktxTexture*

This is true for the `StreambufStream` in `utils/sbufstream.h`. This is the destructor defined in `utils/sbufstream.h`:

```C++
    virtual ~StreambufStream()
    {
        if (!_destructed)
            stream()->destruct(stream());
            /*  here the destruct callback should be removed */
    }
```

For `StreambufStream` in `tests/streamtests.cc`, doing this will defeat part of the purpose of the tests in first place (which is to test that `ktxTextre*` destroys the stream it was supplied with). In other words, for streamtests, `StreambufStream` has to outlive the `ktxStreamTest` wrappers to test if `ktxTexture_Destroy` destructs the stream.

Does `utils/sbufstream.h` have (unit)tests? I can write a simple test that will trigger the same issue and then add the callback removal to fix it.

MarkCallow:
> Does utils/sbufstream.h have (unit)tests? I can write a simple test that will trigger the same issue and then add the callback removal to fix it.

There are the streamtests, `tests/streamtests.cc`. I've just looked at that file and it is not using `utils/sbufstream`. It has its own copy. It should use the header file. Please change it to use the header, add a test that triggers the issue and a fix.

It is an error to delete a stream while it is in use in a ktxTexture. I think a destructor like ~StreambufStream cannot fail so I don't see a way to prevent the bad behaviour.

walcht:
At first, I added this `KtxTextureOutlivesStreambufStream`, removed the StreambufStream from streamtests and used the one from `utils/sbufstream` (which, by the way, had a missing implementation).

The added test then failed when ran through ASan (as expected).

<s>So I added this one liner to `destruct()`: `str->data.file = NULL;` which in turn fixed these issues (no need to worry about ordering/object lifetimes anymore).</s> => oops, this still fails (because `str` is not the `ktxStream*` that is created by `ktxTexture?_CreateFromStream`). I fixed this by setting `stream->data.file = NULL` so that further destructs are not called by libktx.

<s>I think we should introduce a private field...</s> See how I addressed this in latest commit.

tests/streamtests/streamtests.cc
```diff
@@ -360,44 +194,66 @@ bool expectSameTextureContent(const ktxTexture* tex1, const ktxTexture* tex2)
 
 // --- Tests ---
 
+/// Test any memory issues (mainly through ASan) that might occur when ktxTexture?
+/// outilives the ktxStream owner StreambufStream. This mainly tests that
```
MarkCallow: s/outilives/outlives/

tests/streamtests/streamtests.cc
```diff
@@ -360,44 +194,66 @@ bool expectSameTextureContent(const ktxTexture* tex1, const ktxTexture* tex2)
...
+    {  // ktxTexture2 outlives StreambufStream but calls destruct before StreambufStream goes out of scope
+        StreambufStream ktx2Stream{std::move(_ktx2Streambuf), std::ios::in};
+        ktxTexture2_CreateFromStream(ktx2Stream.stream(), KTX_TEXTURE_CREATE_LOAD_IMAGE_DATA_BIT, texture2.pHandle());
+    }
+}  // ktxTexture? should NOT call destruct because it will cause invalid memory access
+
+/// This is commented out because this illustrates an incorrect usage of ktxStreams.
+/// If you create an object (ktxTexture2) that still needs to access data owned by another object
+/// (in this case StreambufStream) then it is your responsibility to guarantee that the dependency
+/// at least outlives (>=) the dependant
```
MarkCallow: s/dependant/dependent/

tests/streamtests/streamtests.cc
```diff
@@ -360,44 +194,66 @@ bool expectSameTextureContent(const ktxTexture* tex1, const ktxTexture* tex2)
...
+
+// TEST_F(ktxStreamTest, KtxTextureOutlivesStreambufStreamCase1)
+// {
+//     KtxTexture<ktxTexture2> texture2;
+//     {  // ktxTexture2 outlives StreambufStream but doesn't call 
+//         StreambufStream srcStream{std::move(_ktx2Streambuf), std::ios::in};
+//         ktxTexture2_CreateFromStream(srcStream.stream(), KTX_TEXTURE_CREATE_NO_FLAGS, texture2.pHandle());
+//         // ktxStream copy created by ktxTexture?_CreateFromStream should be destroyed otherwise you get all
```
MarkCallow: I don't think `ktxTexture?_CreateFromStream` creates a copy. But indeed, after this call the stream will still be alive but so will srcStream. What is the incorrect usage?

utils/sbufstream.h
```diff
@@ -70,18 +70,16 @@ class StreambufStream
...
-            stream()->destruct(stream());
-    }
-
     inline ktxStream* stream() const
     {
         return _stream.get();
     }
 
-    std::streambuf* streambuf() const;
+    // TODO: why wasn't this defined before? This was causing linkage errors
```
MarkCallow: The deleted comment that was at line 213 explains why. Are you sure it now works when the object is created using the unique_ptr<std::streambuf> variant?

I can't remember what issues I, or possibly it was Paolo, had before that led me/him to have those inline functions outside the template definition.

lib/src/texture2.c
```diff
@@ -2578,6 +2578,7 @@ ktxTexture2_IterateLoadLevelFaces(ktxTexture2* This, PFNKTXITERCB iterCb,
 
     // No further need for this.
     stream->destruct(stream);
+    stream->data.file = NULL;  // So that destruct is not invoked again
```
MarkCallow, line 2581: Why is this necessary? You have already added setting stream->data.file to NULL to stream->destruct.

lib/src/texture2.c
```diff
@@ -2744,6 +2745,7 @@ ktxTexture2_loadImageDataInt(ktxTexture2* This,
 
     // No further need for stream or file offset.
     prtctd->_stream.destruct(&prtctd->_stream);
+    prtctd->_stream.data.file = NULL;
```
MarkCallow, line 2748: Same question as at line 2581.

tests/streamtests/streamtests.cc
```diff
@@ -360,44 +194,66 @@ bool expectSameTextureContent(const ktxTexture* tex1, const ktxTexture* tex2)
 
 // --- Tests ---
 
+/// Test any memory issues (mainly through ASan) that might occur when ktxTexture?
+/// outilives the ktxStream owner StreambufStream. This mainly tests that
+/// ~StreambufStream() is well implemented.
+TEST_F(ktxStreamTest, KtxTextureOutlivesStreambufStreamCase0)
+{
+    KtxTexture<ktxTexture2> texture2;
+    {  // ktxTexture2 outlives StreambufStream but calls destruct before StreambufStream goes out of scope
```
MarkCallow: It took me a while to wrap my head around this comment. Something like

// ktxTexture2_CreateFromStream with KTX_TEXTURE_CREATE_LOAD_IMAGE_DATA_BIT will destroy
// the stream when it finishes loading the data. Thus this tests that no issues occur when the
// StreambufStream outlives its underlying ktxStream.

is clearer, I think.

tests/streamtests/streamtests.cc
```diff
@@ -360,44 +194,66 @@ bool expectSameTextureContent(const ktxTexture* tex1, const ktxTexture* tex2)
...
+
+/// This is commented out because this illustrates an incorrect usage of ktxStreams.
+/// If you create an object (ktxTexture2) that still needs to access data owned by another object
+/// (in this case StreambufStream) then it is your responsibility to guarantee that the dependency
+/// at least outlives (>=) the dependant
+
+// TEST_F(ktxStreamTest, KtxTextureOutlivesStreambufStreamCase1)
+// {
+//     KtxTexture<ktxTexture2> texture2;
+//     {  // ktxTexture2 outlives StreambufStream but doesn't call
```
MarkCallow: Doesn't call what? Is something missing in this sentence. It can't be referring to the next line that creates the srcStream because without that `ktxTexture2_CreateFromStream` will not succeed.

tests/streamtests/streamtests.cc
```diff
@@ -360,44 +194,66 @@ bool expectSameTextureContent(const ktxTexture* tex1, const ktxTexture* tex2)
...
+// TEST_F(ktxStreamTest, KtxTextureOutlivesStreambufStreamCase1)
+// {
+//     KtxTexture<ktxTexture2> texture2;
+//     {  // ktxTexture2 outlives StreambufStream but doesn't call 
+//         StreambufStream srcStream{std::move(_ktx2Streambuf), std::ios::in};
+//         ktxTexture2_CreateFromStream(srcStream.stream(), KTX_TEXTURE_CREATE_NO_FLAGS, texture2.pHandle());
+//         // ktxStream copy created by ktxTexture?_CreateFromStream should be destroyed otherwise you get all
+//         // sorts of issues. StreambufStream simply cannot guarantee this.
+//     }
+// }   // ktxTexture? should NOT call destruct again
```
MarkCallow: This is where it becomes incorrect usage. `srcStream` has gone out of scope and been destroyed but ktxTexture2 still exists. Any call to a `ktxTexture?_*` function that involves reading data will most likely crash as will `ktxTexture?_destruct` when called as `ktxTexture2` goes out of scope.

To fix the latter we need `srcStream` to remove its destruct pointer from its underlying `ktxStream`.

utils/sbufstream.h
```diff
@@ -70,18 +70,16 @@ class StreambufStream
     StreambufStream(StreambufStream&&) = delete;
     StreambufStream &operator=(StreambufStream&&) = delete;
 
-    virtual ~StreambufStream()
```
MarkCallow, line 73: Why did you delete this method?

utils/sbufstream.h
MarkCallow: Please add a space before ktxStream. I suppose it might compile but I find it hard to parse.

MarkCallow (commented):
Please address all these comments.

Sorry for delay in reviewing. I was away the past 2 days.

tests/streamtests/streamtests.cc
```diff
@@ -351,6 +194,34 @@ bool expectSameTextureContent(const ktxTexture* tex1, const ktxTexture* tex2)
...
+/// outilives the ktxStream owner StreambufStream. This mainly tests that
+/// ~StreambufStream() is well implemented.
+TEST_F(ktxStreamTest, KtxTextureOutlivesStreambufStreamCase0)
+{
+    KtxTexture<ktxTexture2> texture2;
+    {  // ktxTexture2 outlives StreambufStream but calls destruct before StreambufStream goes out of scope
+        StreambufStream ktx2Stream{std::move(_ktx2Streambuf), std::ios::in};
+        ktxTexture2_CreateFromStream(ktx2Stream.stream(), KTX_TEXTURE_CREATE_LOAD_IMAGE_DATA_BIT, texture2.pHandle());
+    }
+}  // ktxTexture? should NOT call destruct because it will cause invalid memory access
```
MarkCallow: Not call `destruct` on what? Destroying `ktxTexture2` will not cause any problem because the stream will have been destroyed when loading completed.

walcht:
I address the comments here so they are more visible:

> I don't think `ktxTexture?_CreateFromStream` creates a copy. But indeed, after this call the stream will still be alive but so will srcStream. What is the incorrect usage?

It does create a copy of `ktxStream` struct (not of its underlying data, of course). `ktxTexture2_CreateFromStream` calls  `ktxTexture2_constructFromStream` which calls `ktxTexture2_constructFromStream` which calls `ktxTexture_constructFromStream` which has this line (216 in `texture.c`: `*stream = *pStream`):

```C
KTX_error_code
ktxTexture_constructFromStream(ktxTexture* This, ktxStream* pStream,
                               ktxTextureCreateFlags createFlags)
{
    ktxStream* stream;
    // ...
    stream = ktxTexture_getStream(This);
    // Copy stream info into struct for later use.
    *stream = *pStream;
    // ...
}
```

**This is crucial to explain the following comments/changes.**

> Why is this necessary? You have already added setting stream->data.file to NULL to stream->destruct.

I removed that line in `stream->destruct()`. It's a matter of choice; either:
1. include it in libktx itself so that regardless of who supplies `destruct()`, it is guaranteed to not be called more than once.
2. include that line in `StreambufStream` `destruct()` but then any user supplying custom `destruct()` callback need to not forget to set that field to `NULL` otherwise OOM access issues.

I went with options 1). Let me know if you want me to opt for options 2).

> Why did you delete this method?

I deleted `virtual ~StreambufStream()` because it is misleading. It is checking `if (!_destructed)` which is a flag used to track whether `destruct()` **has been called on the ktxStream copied by ktxTexture?_CreateFromStream** but then the next statement proceeds to **call `destruct()` on `StreambufStream`'s `ktxStream`**.

Should I just keep it as empty virtual destructor? Like `virtual ~StreambufStream() {}`

> This is where it becomes incorrect usage. srcStream has gone out of scope and been destroyed but ktxTexture2 still exists. Any call to a ktxTexture?_* function that involves reading data will most likely crash as will ktxTexture?_destruct when called as ktxTexture2 goes out of scope.
>
> To fix the latter we need srcStream to remove its destruct pointer from its underlying ktxStream.

This will still not fix the latter problem (again, because `ktxTexture?_CreateFromStream` creates a copy of `ktxStream` struct). Only way to address this is to:
- manually call `destruct()` on `ktxTexture2`'s `ktxStream` before `StreambufStream` goes out of scope.
- set `ktxTexture2`'s `ktxStream`'s `data.file` to `NULL` before `StreambufStream` goes out of scope.

> Not call destruct on what? Destroying ktxTexture2 will not cause any problem because the stream will have been destroyed when loading completed.

Without this line `stream.data.file = NULL` `ktxTexture2` will call `destruct()` **again** which will cause invalid memory access. You can remove that line and run tests with ASan and it will detect OOM bound access.

I addressed the typo fixes but I will only push changes after receiving feedback on this comment.

> The deleted comment that was at line 213 explains why. Are you sure it now works when the object is created using the unique_ptrstd::streambuf variant?

Before, it wasn't even linking (undefined reference to `streambuf()`). In all test cases, the `StreambufStream` is created as such: `StreambufStream ktx1Stream{std::move(_ktx1Streambuf), std::ios::in};` which is essentially the same as: `StreambufStream ktx?Stream{std::unique_ptr<std::streambuf>(std::make_unique<std::filebuf>()), std::ios::in};`. Tests are running fine (I'm certain this is fine).

MarkCallow:
> It does create a copy of `ktxStream` struct (not of its underlying data, of course).

Thank you for pointing that out. This means the comment starting at line 205 in sbufstream.h is totally wrong and must be rewritten. Here is a try so you can check if my understanding is now correct.

// ktxTexture?_CreateFromStream makes a copy of the stream structure passed as its pStream
// parameter.  Once the application has called one of these functions this stream object has no
// further use but this class has no way to know when its usefulness has passed.


> > Why is this necessary? You have already added setting stream->data.file to NULL to stream->destruct.
> 
> I removed that line in `stream->destruct()`. It's a matter of choice; either:
> 
>     1. include it in libktx itself so that regardless of who supplies `destruct()`, it is guaranteed to not be called more than once.
> 
>     2. include that line in `StreambufStream` `destruct()` but then any user supplying custom `destruct()` callback need to not forget to set that field to `NULL` otherwise OOM access issues.
> 
> 
> I went with options 1). Let me know if you want me to opt for options 2).

No. 1 is the right choice.

I understand now. I'd prefer if we could do it in the `destruct()` callback but doing it in a way to guarantee not to call `destruct()` more than once would require calling a standard destructor which calls the custom destructor. Too hard to do in c. Leave the code as is and resolve those comments.

> 
> > Why did you delete this method?
> 
> I deleted `virtual ~StreambufStream()` because it is misleading. It is checking `if (!_destructed)` which is a flag used to track whether `destruct()` **has been called on the ktxStream copied by ktxTexture?_CreateFromStream** but then the next statement proceeds to **call `destruct()` on `StreambufStream`'s `ktxStream`**.
> 
> Should I just keep it as empty virtual destructor? Like `virtual ~StreambufStream() {}`

It seems there is nothing for StreambufStream to destruct if it was created with a pointer to a std::streambuf as it holds no data and there is no file to close. If it was created with a `std::unique_ptr` do we need a destructor to release the pointer? If not then having no destructor is fine. Should we consider  adding a move constructor that takes over the std::streambuf? If we do then StreambufStream will need a destructor to destruct the std::streambuf.

> 
> > This is where it becomes incorrect usage. srcStream has gone out of scope and been destroyed but ktxTexture2 still exists. Any call to a ktxTexture?_* function that involves reading data will most likely crash as will ktxTexture?_destruct when called as ktxTexture2 goes out of scope.
> > To fix the latter we need srcStream to remove its destruct pointer from its underlying ktxStream.
> 
> This will still not fix the latter problem (again, because `ktxTexture?_CreateFromStream` creates a copy of `ktxStream` struct). Only way to address this is to:
> 
>     * manually call `destruct()` on `ktxTexture2`'s `ktxStream` before `StreambufStream` goes out of scope.
> 
>     * set `ktxTexture2`'s `ktxStream`'s `data.file` to `NULL` before `StreambufStream` goes out of scope.
> 

Agreed.


> I addressed the typo fixes but I will only push changes after receiving feedback on this comment.

Please resolve the relevant comments and go ahead after updating the comment at line 205 in sbufstream.h.

> 
> > The deleted comment that was at line 213 explains why. Are you sure it now works when the object is created using the unique_ptrstd::streambuf variant?
> 
> Before, it wasn't even linking (undefined reference to `streambuf()`). In all test cases, the `StreambufStream` is created as such: `StreambufStream ktx1Stream{std::move(_ktx1Streambuf), std::ios::in};` which is essentially the same as: `StreambufStream ktx?Stream{std::unique_ptr<std::streambuf>(std::make_unique<std::filebuf>()), std::ios::in};`. Tests are running fine (I'm certain this is fine).

This is puzzling. sbufstream.h is used by most tools in `ktx` and obviously it has been compiling. They use it successfully for reading input. They have problems using it for output as documented here https://github.com/KhronosGroup/KTX-Software/blob/4b5b315d296b471873b2b5901d01f79c3db0726f/tools/ktx/command.cpp#L144-L145 and here https://github.com/KhronosGroup/KTX-Software/blob/4b5b315d296b471873b2b5901d01f79c3db0726f/tools/ktx/command.cpp#L185-L186. I have not had time to investigate this issue but it looks unrelated to referencing `streambuf()`.  Ahhh! The tools never call `streambuf()` and streamtests previously did not use `sbufstream.h`.

tests/streamtests/streamtests.cc
MarkCallow: It does. Oops.

walcht:
> If it was created with a std::unique_ptr do we need a destructor to release the pointer?

I don't think so. The constructor for `StreambufStream` is:

```C++
   StreambufStream(T streambuf,
                   std::ios::openmode seek_mode = std::ios::in | std::ios::out)
        : _streambuf{std::move(streambuf)}
        , _seek_mode{seek_mode}
        , _stream{std::make_unique<ktxStream>()} //< auto-cleaned
        , _destructed{false}
    {
        initialize_stream();
    }
```

The line `_streambuf{std::move(streambuf)}` insures that if a `std::unique_ptr` is passed then it is moved (i.e., owned by `StreambufStream`) and is cleaned up automatically (e.g., if `std::unique_ptr<std::filebuf>` is passed then when `StreambufStream` is destroyed so is the filebuf (with all of its resources/file descriptors).

> Should we consider adding a move constructor that takes over the std::streambuf? If we do then StreambufStream will need a destructor to destruct the std::streambuf.

Isn't this already done? The line `_streambuf{std::move(streambuf)}` either takes of a C-style pointer (which is just assigning a pointer => no ownership) or takes ownership of an `std::unique_ptr` in which case the destructor is not needed because of automatic cleanup of `std::unique_ptr`. I might be misunderstanding you here...

> Ahhh! The tools never call streambuf() and streamtests previously did not use sbufstream.h.

Exactly. I `grep`'ed the whole codebase for `streambuf(` references and it is only called in `streamtests`.

By the way, now I do understand why these (that I have removed) were added (and why I should not remove them):

```C++
// I have not yet found a way to do this inside the template definition.
// However `inline` should prevent any multiple definition errors.
template<>
inline std::streambuf* StreambufStream<std::streambuf*>::streambuf() const
{
    return _streambuf; // in case of a simple pointer (e.g., see tools/ktx/command_compare)
}

template<>
inline std::streambuf* StreambufStream<std::unique_ptr<std::streambuf>>::streambuf() const
{
    return _streambuf.get();  // in case of a smart pointer (e.g., see tests/streamtests/streamtests.cc)
}
```

I should revert my deletion of these and figure out why the second definition (with std::unique_ptr) is not being discovered by the linker (maybe the `inline`?)...

Please do not merge this yet as `streambuf()` **has to also work correctly for ktx tools**, which currently it doesn't (if you add a reference to it, it will cause linkage issues).

Edit: it is not because of the `inline`. This is probably happening because `utils` is a different compilation target as `streamtests.cc`.

This is the linkage errors I get (after reverting the removal of the two functions above):

```
/usr/bin/ld: CMakeFiles/streamtests.dir/streamtests.cc.o: in function `(anonymous namespace)::ktxStreamTest_CanWriteKtx1AsKtx2ToCppStream_Test::TestBody()':
/home/walcht/KTX-Software/tests/streamtests/streamtests.cc:302:(.text+0x8b85): undefined reference to `StreambufStream<std::unique_ptr<std::__cxx11::basic_stringbuf<char, std::char_traits<char>, std::allocator<char> >, std::default_delete<std::__cxx11::basic_stringbuf<char, std::char_traits<char>, std::allocator<char> > > > >::streambuf() const'
```

This seems to have solved the issues (and is cleaner):

```C++
    // Implementation depends on whether a std::unique_ptr or a raw pointer is
    // passed.
    std::streambuf* streambuf() const {
        if constexpr (std::is_pointer<T>::value)
            return _streambuf;
        else
            return _streambuf.get();
    }
```

MarkCallow:
> This seems to have solved the issues (and is cleaner):
> 
> ```c++
>     // Implementation depends on whether a std::unique_ptr or a raw pointer is
>     // passed.
>     std::streambuf* streambuf() const {
>         if constexpr (std::is_pointer<T>::value)
>             return _streambuf;
>         else
>             return _streambuf.get();
>     }
> ```

Brilliant! Thanks.

MarkCallow:
It looks good now but I wonder if, in this case, the destruct callback to StreambufStream is needed. There is nothing for it to do. The `_destructed` instance variable would only be needed if there was no copy of the stream. Make the destruct callback point to a no-op?

MarkCallow:
> Isn't this already done? The line `_streambuf{std::move(streambuf)}` either takes of a C-style pointer (which is just assigning a pointer => no ownership) or takes ownership of an `std::unique_ptr` in which case the destructor is not needed because of automatic cleanup of `std::unique_ptr`. I might be misunderstanding you here...

Yes. My bad!

MarkCallow:
> It looks good now but I wonder if, in this case, the destruct callback to StreambufStream is needed. There is nothing for it to do. The `_destructed` instance variable would only be needed if there was no copy of the stream. Make the destruct callback point to a no-op?

I am waiting for the answer to this before considering merging this PR.

walcht:
> It looks good now but I wonder if, in this case, the destruct callback to StreambufStream is needed. There is nothing for it to do. The _destructed instance variable would only be needed if there was no copy of the stream.

It is only needed for `streamtests` to test if `ktxCreate?_FromStream()` has called `destruct()` on the struct (the copied struct). See `EXPECT_TRUE` line below:

```C++
TEST_F(ktxStreamTest, CanCreateKtx2FromCppStream)
{
    StreambufStream ktx2Stream{std::move(_ktx2Streambuf), std::ios::in};
    KtxTexture<ktxTexture2> texture2;

    KTX_error_code err = ktxTexture2_CreateFromStream(ktx2Stream.stream(), KTX_TEXTURE_CREATE_LOAD_IMAGE_DATA_BIT, texture2.pHandle());
    EXPECT_EQ(err, KTX_SUCCESS) << "Failed to create KTX2 from C++ stream: " << ktxErrorString(err);
    ASSERT_NE(texture2, nullptr) << "Newly-created KTX2 is null";
    EXPECT_TRUE(ktx2Stream.destructed()) << "ktxStream should have been destructed (LOAD_IMAGE_DATA_BIT set)";
}
```

Other than this, it serves no other use.

MarkCallow:
> Other than this, it serves no other use.

Please remove it, set _stream.destruct to point to a nop function and update the tests. As is, it causes confusion.

utils/sbufstream.h
```diff
@@ -196,19 +190,10 @@ class StreambufStream
...
-    {
-        auto self = parent(str);
-        self->_destructed = true;
-    }
+    // NOOP function to be passed to ktxStream's destruct
+    static void destruct_noop(ktxStream*) {}
 
     T _streambuf;
     std::ios::openmode _seek_mode;
     std::unique_ptr<ktxStream> _stream;
```
MarkCallow, line 202: I still want a comment here:

// Note that ktxTexture?_CreateFromStream makes a copy of _stream, the stream structure
// passed as its pStream parameter. The static stream functions above will receive a
// pointer to that copy not to this _stream.

MarkCallow: Feel free to reword this.

MarkCallow:
I have merged https://github.com/KhronosGroup/KTX-Software-CTS/pull/79. Unfortunately the commit id changed - I have no idea why GitHub does not do fast-forward merges - so you must change the CTS ref. in this PR.

MarkCallow:
Thank you for this huge improvement to this software's robustness.

PR #1214, opened 2026-07-15
Fix #1189

### LINKED ISSUE #1189: ktxToolsTest.create.image_slices sometimes fails when parallel testing
Reason:

`image_slices.json`, `image_slices_list_names.json`, and `image_slices_list_paths.json` are all writing to the same output files which, depending on whether they are ran on the same thread or not, could result in a file race condition.

The fix is to simply write outputs for each of these in different folders but keep the same golden files (of course).

Will open a PR shortly to fix this.

walcht:
Fixes #1189.

Note: Github is showing that: `Submodule cts updated 147 files`. I have no idea what those 147 files are, I clearly only updated 2 (see https://github.com/KhronosGroup/KTX-Software-CTS/pull/75).

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -67,6 +67,20 @@ jobs:
             doc: OFF, jni: ON, loadtests: OpenGL+Vulkan, py: ON, tools: ON, tools_cts: ON,
             sse: ON, opencl: OFF
           }
+        # We perform ASan/UBSan/TSan on only one job because they greatly slow
+        # down test runs. We use a linux runner because they consume the fewest
+        # GH Actions credits
+        - name: ASan GCC (x86_64)
+          os: ubuntu-latest
+          generator: Ninja
+          arch: x86_64
+          compiler: gcc
+          vk_sdk_ver: '1.4.313'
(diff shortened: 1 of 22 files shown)


MarkCallow:
I have merged https://github.com/KhronosGroup/KTX-Software-CTS/pull/75. Please update the CTS ref again here.

MarkCallow:
Are the tests running in parallel now or do we need to add something to scripts or CI?

walcht:
@MarkCallow 

If you want to run them in parallel then we have to add `-j 4` flag to `ctest` commands in CI. I haven't changed that command in any CIs whatsoever => so now they are still running single core mode. Maybe I can try these in a PR and see if there is actual speed up? (on my machine, with `-j` is almost x4 times faster).
(`-j 4` because Github CI runners run usually on 4 cores).

MarkCallow:
> Maybe I can try these in a PR and see if there is actual speed up? (on my machine, with `-j` is almost x4 times faster).
> (`-j 4` because Github CI runners run usually on 4 cores).

Please do. It would be great to speed up the runners, especially after the addition of the asan job.

Issue #40, opened 2018-03-13
ktxHashTable_Deserialize does not respect endianness


greatgoogleymoogley:
lib/hashtable.c:316 just uses a cast to read the key value size which will be wrong if the source file uses opposite endianness. The deserialise function needs a parameter to specify endianness.

MarkCallow:
Thanks for the report. Fixed in master (PR #48) and incoming (PR #49).

It is fixed by having the functions that read and return key-value data byte swap the key+value lengths, if necessary, before returning the data. This avoids an incompatible change with the existing API. It also avoids exposing outside the API whether or not swapping is needed.

You might be asking "what about metadata with binary values". At present none is defined in the spec.  In the event of someone defining their own, the spec says  _It is up to the vendor defining the key to specify how those bytes are to be interpreted (including the endianness of any encoded numbers)._

PR #1216, opened 2026-07-17
Parallelize CI builds and tests

walcht:
As discussed, tests *should* now be able to run in parallel and CIs may benefit from running tests in parallel (also build steps may benefit as well).

* Use -j $(nproc) on linux, macos, and windows CIs to speed up CMake building and testing steps.

* Add `echo ${cmake_command}` to all CMake configuration steps for consistently with the already-added echo statement in `scripts/build_linux` (this is useful to replicate CMake configuration command locally).

CMake build stats on my system (time refers to real time. CMake was configured as follows: `cmake -S . -B build/ -DKTX_FEATURE_TOOLS=ON -DKTX_FEATURE_TOOLS_CTS=ON -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DCMAKE_BUILD_TYPE=Debug -DKTX_FEATURE_DOC=OFF -DSANITIZE="address"`, on `Linux 6.17.0-40-generic #40~24.04.1-Ubuntu`):
- `cmake --build build/ --config Debug -j 1` (default): 2m10.458s
- `cmake --build build/ --config Debug -j 4` (GH runners best, probably): 0m39.705s

Compared against CI from latest commit: https://github.com/KhronosGroup/KTX-Software/actions/runs/29578243343/

Stats (cmake test times) with `njobs=$(nproc)` all in seconds:
| CI Name                                      | Time ST | Time MT    | speed up |
| ------------------------------------------- | ------------ | -----------    | ------------ |
| Linux Default                               |   1427.77 |  601.10     | ~2.4 |
| Linux ASan GCC (x86_64)          |     3870.06 | 1609.89 | ~2.4 |
| Linux Package (x86_64)              |    379.56   | 159.39    | ~2.4 |
| Windows Visual Studio 17 2022  |     544.25  | 212.50    | ~2.55 |
| Linux CI Summary                       |   1h 40m 2s  | 31m 21s  | ~3 |

Stats (cmake test times) with `njobs=$(nproc) + 1` all in seconds (Why? If build and/or test commands are IO-bound, setting to `nbr_hardware_threads + 1` might result is even better speed ups):
| CI Name                                      | `njobs=$(nproc)` | `njobs=$(($(nproc) + 1))` | speed up |
| ------------------------------------------- | ------------ | ----------- | ------------ |
| Linux Default                                | 601.10 | 594.79  |  ~1 |
| Linux ASan GCC (x86_64)          |  1609.89 |  1685.05 | ~0.95 |
| Linux Package (x86_64)              | 159.39  | 151.28 | ~1 |
| Windows Visual Studio 17 2022  | 212.50 | 214.20 | ~1 |
| Linux CI Summary                       | 31m 21s  | 33m 54s  | ~0.92 |

Conclusion: using `njobs=$(nproc)` results in better performance than `njobs=$(($(nproc) + 1))` and at least ~2.5 faster CIs than ST CIs.

### DIFF
--- a/scripts/build_android.sh
+++ b/scripts/build_android.sh
@@ -49,7 +49,7 @@ echo ${config_display%??}
 
 # To be supplied as `-j $njobs`. We might add `+ 1` if the particular cmd is IO
 # bound (this is done in a lot of CIs). On GH CIs, this is most likely to be 4.
-njobs=$(nproc)
+njobs=$(($(nproc) + 1))
 
 # Print cmake command to be able to verify configuration and replicate locally
 echo "running cmake command (in source directory): cmake . ${cmake_args[@]}"
--- a/scripts/build_android.sh
+++ b/scripts/build_android.sh
@@ -49,7 +49,7 @@ echo ${config_display%??}
 
 # To be supplied as `-j $njobs`. We might add `+ 1` if the particular cmd is IO
 # bound (this is done in a lot of CIs). On GH CIs, this is most likely to be 4.
-njobs=$(($(nproc) + 1))
+njobs=$(nproc)
 
 # Print cmake command to be able to verify configuration and replicate locally
 echo "running cmake command (in source directory): cmake . ${cmake_args[@]}"
--- a/scripts/build_android.sh
+++ b/scripts/build_android.sh
@@ -46,14 +46,21 @@ for arg in "${cmake_args[@]}"; do
 done
 
 echo ${config_display%??}
+
+# To be supplied as `-j $njobs`. We might add `+ 1` if the particular cmd is IO
+# bound (this is done in a lot of CIs). On GH CIs, this is most likely to be 4.
+njobs=$(nproc)
+
+# Print cmake command to be able to verify configuration and replicate locally
+echo "running cmake command (in source directory): cmake . ${cmake_args[@]}"
 cmake . "${cmake_args[@]}"
 
 pushd "$BUILD_DIR"
 
 echo "Build KTX-Software (Android $ANDROID_ABI $CONFIGURATION)"
-cmake --build . --config $CONFIGURATION -j
+cmake --build . --config $CONFIGURATION -j $njobs
 # echo "Test KTX-Software (Android $ANDROID_ABI Release)"
-# ctest --output-on-failure -C $CONFIGURATION # --verbose
+# ctest --output-on-failure -C $CONFIGURATION -j $njobs # --verbose
 echo "Install KTX-Software (Android $ANDROID_ABI $CONFIGURATION)"
 cmake --install . --config $CONFIGURATION --prefix ../$INSTALL_DIR
 
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -68,8 +68,12 @@ jobs:
             sse: ON, opencl: OFF
           }
         # We perform ASan/UBSan/TSan on only one job because they greatly slow
-        # down test runs. We use a linux runner because they consume the fewest
-        # GH Actions credits
+        # down test runs. We only use a linux runner because they, most likely,
+        # consume the fewest GH Actions credits.
+        # TODO: at some point, we can reduce CI runners time by adding
+        #       ASan/UBSan/TSan to the `Default` job and removing this
+        #       additional job but then we may also get additional errors about
+        #       load tests.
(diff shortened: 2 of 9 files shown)


scripts/build_android.sh
```diff
@@ -46,14 +46,21 @@ for arg in "${cmake_args[@]}"; do
 done
 
 echo ${config_display%??}
+
+# To be supplied as `-j $njobs`. We might add `+ 1` if the particular cmd is IO
+# bound (this is done in a lot of CIs). On GH CIs, this is most likely to be 4.
+njobs=$(nproc)
```
MarkCallow: Why not use $(nproc) on the cmake and ctest commands as in the linux and ming workflows?

This question applies to the other *.sh scripts too.

MarkCallow (approved):

MarkCallow:
Thank you for this.

MarkCallow:
I merged this because it was all green but now CI is producing the message "The process cannot access the file because it is being used by another process" which looks suspiciously related to parallelization. See https://github.com/KhronosGroup/KTX-Software/actions/runs/29632032766/job/88047527486 for the failing build log.

walcht:
I'm investigating this. Will add a comment once I find the reason why this is happening.

It is weird since (as you can see above) this CI passes at least 3 times and I always run CTest with -j12 on my machine (probably so far more than a hundred times, but only on Linux).

walcht:
@MarkCallow 

Cause of issue (most probably):
 - Only Windows with loadtests fails which is explained by this CMake function that only runs on Windows platforms `ensure_runtime_dependencies_windows`
 - `ensure_runtime_dependencies_windows` copies images, shaders, models, etc into exectuable's destination
 - There are 2 loadtests: `glloadtests.cmake` and `vkloadtests.cmake`, each one is a separate target and each one calls `ensure_runtime_dependencies_windows` => when running `cmake --build build/` in parallel, it may occasionally occur that the following CMake copy command is executed, via 2 processes, in parallel, which probably is the issue (notice how they both can write to same directory at the the same time which may cause one to overwrite the other while the other hasn't completed its copy yet, or something like that):
 
 ```CMake
    add_custom_command( TARGET ${target} POST_BUILD
        COMMAND ${CMAKE_COMMAND} -E copy_directory  "${PROJECT_SOURCE_DIR}/tests/loadtests/common/models" "$<TARGET_FILE_DIR:${target}>/resources"
        COMMENT "Copy models to build destination"
    )
```

Anyways, I will open a PR to revert `cmake --build build/ -j $njobs`. Lesson learned: `cmake --build` in parallel and `add_custom_command` don't mix well together. Also, the speed-up from parallelizing cmake build commands is minimal (all benefits come from parallelizing tests).

MarkCallow:
> Anyways, I will open a PR to revert `cmake --build build/ -j $njobs`. Lesson learned: `cmake --build` in parallel and `add_custom_command` don't mix well together. Also, the speed-up from parallelizing cmake build commands is minimal (all benefits come from parallelizing tests).

Thank you for the investigation. I think your conclusion as to the cause is correct. Is there no way to create a mutex to prevent both processes writing to the output directory at the same time?

I reran the failed build and it succeeded the second time.

walcht:
> Is there no way to create a mutex to prevent both processes writing to the output directory at the same time?

I think closest way to achieve that in CMake is via dependency management using something like `add_dependencies`. But I'm definitely not that versed in CMake to confidently suggest a solution (found some details here, but our case is with `POST_BUILD` commands: https://gitlab.com/inorton/cmake-custom-concurrency/)

> I reran the failed build and it succeeded the second time.

Even if they succeeded, I think we should revert the `cmake --build build/` parallel command and add some comments about this.

MarkCallow:
> Even if they succeeded, I think we should revert the `cmake --build build/` parallel command and add some comments about this.

I agree. We can't afford a race condition in the build. CI is remarkably unreliable as it is.

We can continue to look into dependency management and, if we find something, change to it in a future PR.

MarkCallow:
> We can continue to look into dependency management and, if we find something, change to it in a future PR.

I see what they're saying and it is mentioned in CMake's [add_custom_command](https://cmake.org/cmake/help/latest/command/add_custom_command.html) documentation as well. It doesn't matter that it is PRE- or POST-BUILD. The issue and fix is the same. I will change the dependency installation to use custom targets but in the meantime please make a PR to turn off the parallel builds.

PR #1219, opened 2026-07-19
Fix docstring typo in ktx.h

petar-andrejic:
Docstring for `numLayers` was written as `numLevels` which made it not appear in the doxygen pages


MarkCallow:
This is the second copy-paste error that has been found recently in `ktx.h`. Thank you for the fix @petar-andrejic.

PR #1220, opened 2026-07-20
ci: un-parallelize build commands due to occasional failures on Windows

walcht:
As discussed. See commit message for details.

Added small Note in `BUILDING.md` so that users/developers running `cmake --build build/ -j[N]` won't (hopefully) be surprised when building occasionally fails on Windows (not sure about other platforms).

### DIFF
--- a/.github/workflows/mingw.yml
+++ b/.github/workflows/mingw.yml
@@ -80,9 +80,9 @@ jobs:
     - name: Configure Mingw x64
       run: cmake -B build -G "Ninja Multi-Config" -DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++ -DKTX_FEATURE_TOOLS=TRUE -DKTX_FEATURE_TOOLS_CTS=TRUE -DKTX_WERROR=$WERROR
     - name: Build Mingw x64 Debug
-      run: cmake --build build --config Debug  # multi-core builds are not supported
+      run: cmake --build build --config Debug -j $(nproc)
     - name: Build Mingw x64 Release
-      run: cmake --build build --config Release  # multi-core builds are not supported
+      run: cmake --build build --config Release -j $(nproc)
     - name: Test Mingw build
       run: ctest --output-on-failure --test-dir build -C Release -j $(nproc)
 #    - name: Upload test log
--- a/.github/workflows/mingw.yml
+++ b/.github/workflows/mingw.yml
@@ -80,9 +80,9 @@ jobs:
     - name: Configure Mingw x64
       run: cmake -B build -G "Ninja Multi-Config" -DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++ -DKTX_FEATURE_TOOLS=TRUE -DKTX_FEATURE_TOOLS_CTS=TRUE -DKTX_WERROR=$WERROR
     - name: Build Mingw x64 Debug
-      run: cmake --build build --config Debug -j $(nproc)
+      run: cmake --build build --config Debug  # multi-core builds are not supported
     - name: Build Mingw x64 Release
-      run: cmake --build build --config Release -j $(nproc)
+      run: cmake --build build --config Release  # multi-core builds are not supported
     - name: Test Mingw build
       run: ctest --output-on-failure --test-dir build -C Release -j $(nproc)
 #    - name: Upload test log
--- a/scripts/build_android.sh
+++ b/scripts/build_android.sh
@@ -47,9 +47,10 @@ done
 
 echo ${config_display%??}
 
-# To be supplied as `-j $njobs`. We might add `+ 1` if the particular cmd is IO
-# bound (this is done in a lot of CIs). On GH CIs, this is most likely to be 4.
-njobs=$(nproc)
+# To be supplied as `-j $njobs`. This is set to 1 because multi-core builds
+# might, occasionally, fail due to some filesystem race condition(s) caused by
+# `add_custom_command` calls in CMake.
+njobs=1
 
 # Print cmake command to be able to verify configuration and replicate locally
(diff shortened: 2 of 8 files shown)


.github/workflows/mingw.yml
```diff
@@ -80,9 +80,9 @@ jobs:
     - name: Configure Mingw x64
       run: cmake -B build -G "Ninja Multi-Config" -DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++ -DKTX_FEATURE_TOOLS=TRUE -DKTX_FEATURE_TOOLS_CTS=TRUE -DKTX_WERROR=$WERROR
     - name: Build Mingw x64 Debug
-      run: cmake --build build --config Debug -j $(nproc)
+      run: cmake --build build --config Debug  # multi-core builds are not supported
```
MarkCallow: The Android build does not include FEATURE_LOADTEST_APPS so multi-core build should be fine here. Please revert.

scripts/build_android.sh
```diff
@@ -47,9 +47,10 @@ done
 
 echo ${config_display%??}
 
-# To be supplied as `-j $njobs`. We might add `+ 1` if the particular cmd is IO
-# bound (this is done in a lot of CIs). On GH CIs, this is most likely to be 4.
-njobs=$(nproc)
+# To be supplied as `-j $njobs`. This is set to 1 because multi-core builds
+# might, occasionally, fail due to some filesystem race condition(s) caused by
+# `add_custom_command` calls in CMake.
+njobs=1
```
MarkCallow: As with MinGW the offending feature is not included in this build. Please revert.

scripts/build_android_debug.sh
```diff
@@ -40,7 +40,7 @@ cmake . "${cmake_args[@]}"
 pushd "$BUILD_DIR"
 
 echo "Build KTX-Software (Android $ANDROID_ABI Debug)"
-cmake --build . --config Debug -j
+cmake --build . --config Debug -j 1
```
MarkCallow: Revert here too.

scripts/build_linux.sh
```diff
@@ -153,7 +155,7 @@ do
   IFS=$oldifs # Because of ; IFS set above will still be present.
   # Build and test
   echo "Build KTX-Software (Linux $ARCH $config)"
-  cmake --build . --config $config -j $njobs
+  cmake --build . --config $config
```
MarkCallow: I'd prefer if you just set njobs to 1 and leave the `-j $njobs` here so it will be easier to go back to multi-core builds once I've removed the custom commands.

scripts/build_macos.sh
```diff
@@ -135,9 +136,9 @@ do
   #if [ "$config" = "Debug" ]; then continue; fi
   echo "Build KTX-Software (macOS $ARCHS $config)"
   if [ -n "$CODE_SIGN_IDENTITY" -a "$config" = "Release" ]; then
-    cmake --build . --config $config -j $njobs | handle_compiler_output
+    cmake --build . --config $config | handle_compiler_output
```
MarkCallow: Set njobs=1 for this build too.

scripts/build_win.ps1
```diff
@@ -200,7 +196,7 @@ foreach ($config in $configArray) {
   try {
     #git status
     echo "Build KTX-Software (Windows $ARCH $config)"
-    cmake --build . --config $config -j $njobs
+    cmake --build . --config $config
```
MarkCallow: Set njobs=1 here too.

MarkCallow (changes requested):
Please make the suggested changes.

scripts/build_linux.sh
walcht: This will also be supplied to CTest which significantly speeds up testing. I left a comment instead.

scripts/build_macos.sh
walcht: I put `-j 1` because `njobs` is passed to CTest and setting it to 1 will make CIs significantly slower.

walcht:
MacOS build says it's failed but is also showing that it is still running. Seriously, Github Actions consistently impresses of how bad it can be (apologies for the complaints, but this is like the 20th time in a spam of 2 weeks that Github Actions randomly fail).

scripts/build_android_debug.sh
```diff
@@ -40,9 +40,9 @@ cmake . "${cmake_args[@]}"
 pushd "$BUILD_DIR"
 
 echo "Build KTX-Software (Android $ANDROID_ABI Debug)"
-cmake --build . --config Debug -j 1
+cmake --build . --config Debug -j
```
MarkCallow: Does just `-j` work?

Did you intend to omit the `$(nproc)`?

MarkCallow:
> MacOS build says it's failed but is also showing that it is still running. Seriously, Github Actions consistently impresses of how bad it can be (apologies for the complaints, but this is like the 20th time in a spam of 2 weeks that Github Actions randomly fail).

Per the log, the job is still running tests. Could be in a deadlock due to parallelization? There was no indication of an error in the log.

Anyway I have re-run the job and it has succeeded.

walcht:
> Does just -j work?
>
> Did you intend to omit the $(nproc)?

It was like that even before my previous PR. Essentially, this is how it always had been `cmake --build . --config Debug -j`.
I have no idea why this is the only line in CIs that has `-j` like that.

When I saw the revert request, I thought it was to revert it to as it was initially.

scripts/build_android_debug.sh
MarkCallow: I did not write this workflow and had never noticed before that it had only `-j`. According to the cmake help, absent  \<jobs\> "the native build tool's default number is used." Please add `$(nproc)` so we'll know what it is doing without having to consult make or ninja or whatever the tool is.

MarkCallow (approved):

PR #1223, opened 2026-07-23
Add dds to ktx2 lossless converter

### LINKED ISSUE #92: Fully support DDS to KTX2 pipeline
We should provide a straightforward and well-supported way of producing KTX2 files from DDS.
An official open-source DDS implementation is [here](https://github.com/microsoft/DirectXTex).

walcht:
Fixes #92.

<s>Still in draft because testing isn't yet done (a bit difficult since I have to use my other BCn PR to test extracts)</s> => is now ready for review (still a couple of missing edge-case test files and invalid dds test files).

Feel free to give an early review but this is still far from done.


tools/ktx/command_convert.cpp
```diff
@@ -61,11 +66,17 @@ Convert another texture file type to a KTX2 file
     the alternative mapping to @c VK_FORMAT_ASTC_\*_UNORM_BLOCK they will render
     HDR blocks in the error color.
 
+    @note When converting a DDS file, the conversion is lossless (no
+    decoding-encoding cycle is performed and bits are just copied as they are).
+    Not all formats supported by DDS are currently supported. Currently
```
MarkCallow, line 72: What formats are not supported?

tools/ktx/command_convert.cpp
```diff
@@ -329,6 +365,402 @@ void CommandConvert::convertKtx(InputStream& inputStream, OutputStreamEx& output
     outputStream.writeKTX2(texture, *this);
 }
 
+/// Get the corresponding VkFormat from the provided TinyDDS format. Not all
+/// formats have an equivalent in Vulkan. All BC1-BC7 formats are supported.
+/// The usual LDR and HDR uncompressed formats are also supported. For any
+/// non-supported format, this returns VK_FORMAT_UNDEFINED.
+inline VkFormat dds_to_vkformat(TinyDDS_Format dds_format) {
```
MarkCallow, line 373: We have a JSON database in the KTX-Specification repo that maps Vulkan formats to DXGI formats and forms part of the specification. We also have a Ruby script that generates a switch body for the mapping. (I have completed work locally that adds query functions to get the DXGI and Metal equivalent formats for a ktXtexture2 that is waiting for the v5.0.0 release to leave the station.)

I would like generate the reverse mapping from this data to use here but it requires Ruby coding which I am not very familiar with. It should be done from the database because then it is guaranteed spec. compliant.

tools/ktx/command_convert.cpp
```diff
@@ -329,6 +365,402 @@ void CommandConvert::convertKtx(InputStream& inputStream, OutputStreamEx& output
...
+/// Mimics `ktxTexture2_GetImageSize` because TinyDDS' `TinyDDS_ImageSize`
+/// returns the size of an entire 3D texture, or entire array, or entire
+/// cubemap.
+///
+/// Calculate & return the size in bytes of
+/// an image at the specified mip level. For arrays, this is the size of a
+/// layer, for cubemaps, the size of a face and for 3D textures, the size of a
+/// depth slice.
+///
+/// TODO: The size reflects the padding of each row to KTX\_GL\_UNPACK\_ALIGNMENT.
```
MarkCallow: Does DDS use KTX_GL_UNPACK_ALIGNMENT?

I suspect you copied a comment from the KTX code that refers to handling for KTX v1 data. KTX v2 uncompressed images are always tightly packed.

MarkCallow (commented):
This is exciting. A DDS converter has been on our To Do list for a long time.

This together with your BCn codec and the format queries I mention in one of my comments will make a _very_ nice addition to KTX-Software.

Looks good to me.

walcht:
> Does DDS use KTX_GL_UNPACK_ALIGNMENT?
>
> I suspect you copied a comment from the KTX code that refers to handling for KTX v1 data. KTX v2 uncompressed images are always tightly packed.

It doesn't. I just left the comment there so that I verify if there is anything I have to worry about in terms of alignment. Good to know that KTX v2 is always tightly packed (less things to worry about).

> I would like generate the reverse mapping from this data to use here but it requires Ruby coding which I am not very familiar with. It should be done from the database because then it is guaranteed spec. compliant.

I will try to figure this out. Ruby *should* be fairly simple to pick up, I'll open a PR in KTX-Specification if I manage to generate the enum.

> What formats are not supported?

I left them commented out. Examples:

```C
//  case TDDS_AYUV: return VK_FORMAT_UNDEFINED;
//  case TDDS_Y410: return VK_FORMAT_UNDEFINED;
//  case TDDS_Y416: return VK_FORMAT_UNDEFINED;
//  case TDDS_NV12: return VK_FORMAT_UNDEFINED;
//  case TDDS_P010: return VK_FORMAT_UNDEFINED;
//  case TDDS_P016: return VK_FORMAT_UNDEFINED;
//  case TDDS_420_OPAQUE: return VK_FORMAT_UNDEFINED;
//  case TDDS_YUY2: return VK_FORMAT_UNDEFINED;
//  case TDDS_Y210: return VK_FORMAT_UNDEFINED;
//  case TDDS_Y216: return VK_FORMAT_UNDEFINED;
//  case TDDS_NV11: return VK_FORMAT_UNDEFINED;
//  case TDDS_AI44: return VK_FORMAT_UNDEFINED;
//  case TDDS_IA44: return VK_FORMAT_UNDEFINED;
//  case TDDS_P8: return VK_FORMAT_UNDEFINED;
//  case TDDS_A8P8: return VK_FORMAT_UNDEFINED;

// ...

//  case TDDS_R16_UNORM_X8_TYPELESS: return VK_FORMAT_UNDEFINED;
//  case TDDS_X16_TYPELESS_G8_UINT: return VK_FORMAT_UNDEFINED;
//  case TDDS_P208: return VK_FORMAT_UNDEFINED;
//  case TDDS_V208: return VK_FORMAT_UNDEFINED;
//  case TDDS_V408: return VK_FORMAT_UNDEFINED;
//  case TDDS_R10G10B10_SNORM_A2_UNORM: return VK_FORMAT_UNDEFINED;
//  case TDDS_R4G4_UNORM: return VK_FORMAT_UNDEFINED;
//  case TDDS_G4R4_UNORM: return VK_FORMAT_UNDEFINED;
//  case TDDS_A4B4G4R4_UNORM: return VK_FORMAT_UNDEFINED;
//  case TDDS_X4B4G4R4_UNORM: return VK_FORMAT_UNDEFINED;
//  case TDDS_A4R4G4B4_UNORM: return VK_FORMAT_UNDEFINED;
//  case TDDS_X4R4G4B4_UNORM: return VK_FORMAT_UNDEFINED;
//  case TDDS_B4G4R4X4_UNORM: return VK_FORMAT_UNDEFINED;
//  case TDDS_R4G4B4A4_UNORM: return VK_FORMAT_UNDEFINED;
//  case TDDS_R4G4B4X4_UNORM: return VK_FORMAT_UNDEFINED;
//  case TDDS_B5G5R5X1_UNORM: return VK_FORMAT_UNDEFINED;
//  case TDDS_R5G5B5A1_UNORM: return VK_FORMAT_UNDEFINED;
//  case TDDS_R5G5B5X1_UNORM: return VK_FORMAT_UNDEFINED;
//  case TDDS_A1R5G5B5_UNORM: return VK_FORMAT_UNDEFINED;
//  case TDDS_X1R5G5B5_UNORM: return VK_FORMAT_UNDEFINED;
//  case TDDS_A1B5G5R5_UNORM: return VK_FORMAT_UNDEFINED;
//  case TDDS_X1B5G5R5_UNORM: return VK_FORMAT_UNDEFINED;
//  case TDDS_R5G6B5_UNORM: return VK_FORMAT_UNDEFINED;
//  case TDDS_B2G3R3_UNORM: return VK_FORMAT_UNDEFINED;
//  case TDDS_B2G3R3A8_UNORM: return VK_FORMAT_UNDEFINED;
//  case TDDS_G8R8_UNORM: return VK_FORMAT_UNDEFINED;
//  case TDDS_G8R8_SNORM: return VK_FORMAT_UNDEFINED;
```

I don't know what the `X` stands for in these formats. I also have no idea what the first formats are (e.g., `AYUV`).

MarkCallow:
> I don't know what the `X` stands for in these formats. I also have no idea what the first formats are (e.g., `AYUV`).

In Vulkan formats X means unused a.k.a. padding bits. AYUV is almost certainly a YUV format with alpha.

That database I mentioned is to DXGI formats. I do not know if those and DDS formats are equivalent or use the same enumeration values.

walcht:
> I will try to figure this out. Ruby should be fairly simple to pick up, I'll open a PR in KTX-Specification if I manage to generate the enum.

I couldn't get my head around Ruby so I re-wrote the Ruby scripts in Python. I will open a PR feel free to accept it or close it (at least some Python equivalents can be referred to in PR search).

I'm finalizing the testsuite (there are quite a lot of combinations to test). There are also some issues with `tinydds.h` with regards to arrays/volumes/cubemap arrays (which apparently are supported). I will add another comment about these later.

tools/ktx/command_convert.cpp
```diff
@@ -61,11 +66,17 @@ Convert another texture file type to a KTX2 file
...
+    supported formats include all BC1-BC7 GPU block compressed formats, all LDR
+    uncompressed formats, and all HDR uncompressed formats.
+
 @section ktx\_convert\_options OPTIONS
     The following options are available:
     <dl>
         <dt>-t, \--input-type &lt;type&gt;</dt>
-        <dd>Type of input file. Currently @b type must be @c ktx. Case insensitive.</dd>
+        <dd>Type of input file. Currently @b type must be either @c ktx or @c dds. Case insensitive.</dd>
         <dt>-d, \--drop-bad-orientation</dt>
```
MarkCallow, line 81: Add a note that this is ignored if the input type is not ktx and make sure the implementation follows the note.

MarkCallow:
> I couldn't get my head around Ruby so I re-wrote the Ruby scripts in Python. I will open a PR feel free to accept it or close it (at least some Python equivalents can be referred to in PR search).

We want to keep everything in the spec. repo. Ruby, which we need for Asciidoctor, in order to minimize dependencies so I won't accept the PR. Could the Python version of the DXGI to VkFormat generator you created be useful for working on a Ruby version?

What is the answer to my TDDS vs DXGI question?

walcht:
> Could the Python version of the DXGI to VkFormat generator you created be useful for working on a Ruby version?

I will try to translate it and open it as a PR in Ruby but I have never programmed in it before.

> What is the answer to my TDDS vs DXGI question?

Ah, I forgot to answer. They are the same (as far as I see they map one-to-one). So once we generate `DXGI -> VkFormat` the `TinyDDS -> VkFormat` is trivial. More details:

```C
TinyDDS_Format tinydds_format = TinyDDS_GetFormat(handle); // returns format in this TinyDDS-specific struct `TinyDDS_Format`
if (tinydds_format < TDDS_SYNTHESISED_DXGIFORMATS) {
   DXGI_FORMAT dxgi_format = static_cast<DXGI_FORMAT>(tinydds_format);  // they map 1-to-1
} else {
   // unsupported format (see comment below)
}
```

The formats that cannot be translated from TinyDDS to DXGI_FORMAT are these: https://github.com/KhronosGroup/KTX-Software/blob/7b088823dfc71178bc8929776c3a7b547b62c045/external/basis_universal/encoder/3rdparty/tinydds.h#L333-L387

walcht:
> Add a note that this is ignored if the input type is not ktx and make sure the implementation follows the note.

There is this comment already that addresses it, no? `Ignored unless @b type is @c ktx.`
I verified the implementation and this option is only used in `convertKtx` function.

tools/ktx/command_convert.cpp
walcht: Done. See PR in KTX-Specification and see latest commit.

MarkCallow:
[30a7e76](https://github.com/KhronosGroup/KTX-Software/pull/1223/commits/30a7e7642deb492291118283f70760f8c74dc90c) has an incorrect commit id for tests/cts. The id does not exist. Hence the build failures.

walcht:
Yes, I always keep forgetting to push local CTS changes. Apologies for the mess/many CI runs.

On a side note, I'm trying to clean-up all my currently open PRs so that they are finally finalized.

walcht:
There is an edge case where `UNORM` textures are actually storing sRGB values and so they should be somehow mapped to sRGB (e.g., GIMP DDS exporter always exports UNORM). I haven't looked at this yet but hopefully there is some field in DDS that indicates what colorspace (or at least whether sRGB) is the data saved in.

MarkCallow:
> There is an edge case where `UNORM` textures are actually storing sRGB values and so they should be somehow mapped to sRGB

Ouch! That definitely needs fixing. Sounds like a bug in the GIMP exporter to me.

walcht:
> Sounds like a bug in the GIMP exporter to me

It's probably not. Even for older versions of NVTT sRGB is being exported as `UNORM` and not `UNORM_SRGB` because viewers (at that time) did not support `UNORM_SRGB` textures. You can read some details here:
https://forums.developer.nvidia.com/t/color-conversion-exceptions-using-nvtt-export/219142

Also `bc7enc_rdo` does not save sRGB DDS files as `UNORM_SRGB`. From their source code:

```C++
		// Not all tools support DXGI_FORMAT_BC7_UNORM_SRGB (like NVTT), but ddsview in DirectXTex pays attention to it. So not sure what to do here.
		// For best compatibility just write DXGI_FORMAT_BC7_UNORM.
		//hdr10.dxgiFormat = srgb ? DXGI_FORMAT_BC7_UNORM_SRGB : DXGI_FORMAT_BC7_UNORM;
		hdr10.dxgiFormat = dxgi_format; // DXGI_FORMAT_BC7_UNORM;
		hdr10.resourceDimension = D3D10_RESOURCE_DIMENSION_TEXTURE2D;
		hdr10.arraySize = 1;
```

A great argument on why KTX is superior to DDS is that we have the `KHR_DF_TRANSFER_*` field for each KTX file regardless of the set vkformat.

At worst, we might just add a flag to force to-sRGB vkFormat conversion even if input is UNORM.

I will address this once I finalize the BCn PR.

MarkCallow:
> A great argument on why KTX is superior to DDS is that we have the `KHR_DF_TRANSFER_*` field for each KTX file regardless of the set vkformat.

KTX has rules that allow "mismatched" VkFormat and KHR_DF_TRANSFER only in limited circumstances. Using a UNORM format with  `KHR_DF_TRANSFER_SRGB` when there is a matching SRGB format is not one of those. It is explicitly disallowed.

> 
> At worst, we might just add a flag to force to-sRGB vkFormat conversion even if input is UNORM.

This would not be conversion. It would be assignment, `--assign-oetf` in _ktx create_. Only possible if the DDS file provides an indication that the data is really sRGB. If there is no indicator, I suggest raising a fatal error with a brief explanation about broken DDS tools and guiding them to use `--assign-oetf` to specify the correct OETF to avoid the error.

> 
> I will address this once I finalize the BCn PR.

Please. There is no rush as I will not be merging the converter until v5.0.0 is released. Unfortunately I do not know how much longer it will be until the item blocking its release is resolved. It has nothing to do with the software.

MarkCallow:
> because viewers (at that time) did not support `UNORM_SRGB` textures. You can read some details here

What a shit show. In my view broken viewers are no excuse for a texture creation tool to be outputting incorrect data.

walcht:
Some notes about the whole dds-texture-is-saved-as-unorm-even-though-it-is-srgb:

- DirectXTex provides this [MakeSRGB()](https://github.com/microsoft/DirectXTex/blob/868198cb4bcbc4e359372e7ba38d7a6dda3a6afa/DirectXTex/DirectXTexUtil.cpp#L1466) method that is also widely used in Direct3D12 tutorials/samples to focibly set the DDS imported image to sRGB which means that saving sRGB formats as UNORM in DDS files is widely spread (at least for older formats, I think) and loaders are expected to do reset the DXGI format to sRGB based on some extra info (e.g., if albedo in filename => use `MakeSRGB()`).
- Even the official DirectXTex DDS viewer has a [forcesrgb toggle](https://github.com/microsoft/DirectXTex/blob/868198cb4bcbc4e359372e7ba38d7a6dda3a6afa/DDSView/ddsview.cpp#L117).
- [Some Direct3D12 tutorials](https://www.3dgep.com/learning-directx-12-4/) (which I'm currently learning) have something like this in their code:
  ```C++
        if ( textureUsage == TextureUsage::Albedo )
        {
            metadata.format = MakeSRGB( metadata.format );
        }
  ```

Essentially, a `*_UNORM` in DDS can mean that the colorspace can be anything (I can be wrong, this is my understanding). And even though modern DDS format (DX10 header) can explicitly state that the colorspace is sRGB `*_UNORM_SRGB`, a `*_UNORM` format simply doesn't mean that it isn't but rather that the colorspace is just not explictly set and is up-to the loader to interprate.

This is why I stated above that we probably should add `-forcesrgb` CLI flag/option which calls `MakeSRGB()` on the loaded DDS DXGI format before converting it to a KTX2 file. I just see no other way to address this issue. Again, a very strong point on why KTX2 is just better...

MarkCallow:
I suggest printing a warning when encountering a *_UNORM input that explains the issue, what the default choice is (a UNORM?) and how to override the default choice. There should be an option to disable this warning. As for the override my preference is `--assign-oetf srgb` for consistency with _ktx create_. I realize this forces  typing an argument for the option but I would not be surprised if we find other circumstance where DDS has incorrect or misleading OETF information. This is more future-proof.

If you want to default to sRGB when the texture usage is albedo, the warning message must be changed accordingly. This case is another reason for `--assign-oetf` as `--assign-oetf linear` can be used to override the sRGB choice.

Issue #1224, opened 2026-07-23
Feature request: transcode individual mip levels from a partially-downloaded KTX2 file (progressive texture streaming)


SashaRX:
## Summary
First, a brief disclaimer: I’m a 3D technical artist rather than a library or API developer, so the code and API sketch below should be treated as a practical proof of concept, not as a proposed final design. I fully expect the maintainers may prefer a different architecture, and I’d be happy to adapt the implementation to their guidance.

What I can contribute is a working, production-tested implementation of the underlying container manipulation, along with the practical motivation for supporting this directly in libktx.

## Motivation
KTX2 was explicitly designed for progressive delivery — the spec stores mip levels smallest-first and provides a level index with per-level byte ranges precisely so that a client can fetch levels incrementally (e.g. with HTTP Range requests) and show a low-resolution texture immediately, refining it as more data arrives.

However, libktx today only supports the opposite workflow: the **entire file** must be in memory before `ktxTexture2_CreateFromMemory` + `ktxTexture2_TranscodeBasis` can run. For a web game streaming hundreds of textures, downloading a full 2-4K mip chain just to display a 64×64 preview defeats the purpose of the format's streaming-friendly layout. As far as I can tell, this is why most web engines' KTX2 loaders bypass libktx and drive the Basis Universal transcoder directly, each reimplementing DFD/SGD parsing on their own.

## What we do today (the workaround this proposal would eliminate)

In our PlayCanvas-based game we implemented progressive KTX2 streaming on top of libktx (compiled to WebAssembly) by synthesizing a **single-level "mini-KTX2" file** for every mip level we fetch:

1. Fetch the first ~4 KB once: identifier, header, level index; then DFD, KVD and SGD via Range requests.
2. For each desired level, fetch only its payload using the level index byte range.
3. Build a valid single-level KTX2 container in memory:
   - rewrite the header (`levelCount = 1`, level-adjusted `pixelWidth/Height`),
   - recompute section offsets with proper 4/8-byte alignment,
   - rewrite the level index entry (including the `uncompressedByteLength` rules, which differ between supercompression schemes 0, 1 and 2),
   - **for ETC1S/BasisLZ, re-subset the SGD**: keep the 20-byte header and the global codebooks (endpoints/selectors/tables/extended) but extract only the image descriptors belonging to the target level, accounting for layers/faces/depth.
[...]

Issue #1212, opened 2026-07-13
No user pointer for Vulkan suballocator callbacks


monomere:
This isn't really *that* big of an issue, since a lot of the times state is global anyway, but it would be really great to have the ability to pass a user pointer to `ktxTexture_VkUploadEx_WithSuballocator` that the suballocator callbacks could use. Right now the only option is to use a static variable (possibly thread local, which can be slow (although compared to the memory-heavy function itself its probably nothing)).

I'm not sure how this could be implemented in a backward-compatible way, maybe the same way Vulkan does it with a "2" version of the typedefs. There is also an issue with with the implementation, since not all calling conventions allow just adding an extra argument (this can be fixed by using ifs, but maybe with the specific platforms where libktx is supported, the function pointer can just be cast). I might submit a PR for this if I have time!

monomere:
Additionally, the callbacks pointer is non-const, which might be an issue.

MarkCallow:
Would this `userdata` pointer be needed only when using sub-allocators?

If this was implemented using a "2" version, what is the calling convention issue you foresee?

monomere:
Yes, the pointer is only needed to be able to pass data to the callbacks from the callsite of `ktxTexture_VkUploadEx_WithSuballocator`, similar to the `qsort_s` function for example.

Now that I think about it, the calling convention issue can be avoided by passing the original version to the userdata pointer.

MarkCallow:
> Now that I think about it, the calling convention issue can be avoided by passing the original version to the userdata pointer.

Original version of what? I have not used suballocators myself so am unfamiliar with all this. Current support was added by another volunteer. Please provide a PR that adds a "2" version and handles this userdata and callback.

toomuchvoltage:
Hey there, in case you're wondering that was me. If you'd like to see how this is used, check the usage out here: https://github.com/toomuchvoltage/HighOmega-public/blob/sauray_vkquake2/HighOmega/src/gl.cpp#L313-L474

(This is a slightly newer version for which I'm making a PR very soon.)

toomuchvoltage:
[...]

Issue #1092, opened 2025-12-14
Using 'ktxTexture_VkUpload' with an image layout of 'VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL' causes an validation error.


Link459:
When calling `ktxTexture_VkUploadEx_WithSuballocator` the validation layers complain with an error like this:
```0: type: Command Buffer, handle: 94331481154944

vkCmdPipelineBarrier(): pImageMemoryBarriers[0].dstAccessMask (VK_ACCESS_SHADER_READ_BIT) is not supported by stage mask (VK_PIPELINE_STAGE_ALL_COMMANDS_BIT).
The Vulkan spec states: For each element of pImageMemoryBarriers, if its srcQueueFamilyIndex and dstQueueFamilyIndex members are equal, or if its dstQueueFamilyIndex is the queue family index that was used to create the command pool that commandBuffer was allocated from, then its dstAccessMask member must only contain access flags that are supported by one or more of the pipeline stages in dstStageMask, as specified in the table of supported access types (https://docs.vulkan.org/spec/latest/chapters/synchronization.html#VUID-vkCmdPipelineBarrier-pImageMemoryBarriers-02820)
```

Looking at the code it seems to be caused by these lines in  setImageLayout in vkloader.c
https://github.com/KhronosGroup/KTX-Software/blob/main/lib/src/vkloader.c#L1772-L1781
When checking with the specification it is said that `VK_ACCESS_SHADER_READ_BIT` as an access type is not compatible with a stage of `VK_PIPELINE_STAGE_ALL_COMMANDS_BIT` in a pipeline barrier.
One way to fix it is to use `VK_ACCESS_MEMORY_READ_BIT` instead which supports all stage flags. I have changed that line to use `VK_ACCESS_SHADER_READ_BIT` and it does indeed fix the validation error. Another option would be using a different pipeline stage but I could see that being less favorable as the user would have to choose one or use the default which wouldn't work for all use cases.

MarkCallow:
This does not show up in our tests leading to two questions:

1. Which version of the Vulkan SDK are you using? 
2. If the validation error is simply not revealed by our tests, can you donate a test program that shows it?

Link459:
I am on Arch linux and always try to keep the SDK up no date to catch specification violations.
I made a mistake in my last comment.
It's `VK_ACCESS_MEMORY_READ_BIT`.
[...]

PR #1231, opened 2026-07-29
Efficient thread-safe `ktxTexture_VkUploadEx_WithSuballocator()` with queue guard callbacks

toomuchvoltage:
So the main contribution of this PR is to make `ktxTexture_VkUploadEx_WithSuballocator()` more efficient in a threaded environment. Previously, the entire call would have to be guarded with a queue guard which would effectively make a single upload call block other upload calls or any Vulkan call needing the same queue. With this PR and the introduction of `ktxTexture_VkUploadEx_WithSuballocatorAndQueueGuard()`, only submissions to the queue inside the call are guarded individually leaving other calls to `UploadEx()` (or just general queue accesses from Vulkan) unblocked until they need the queue.

The PR also includes a couple of other fixes as well:

- If `ktxTexture_LoadImageData()` mid-call failed, it would return a failure code but leave mapped memory dangling. This normally would be wasteful but not fatal. However, if we are utilizing synchronization primitives in the callbacks this would lead to a deadlock. Mapping memory would need to enter the critical section of a memory guard (protecting `VkDeviceMemory` access) and unmapping would have to leave it. This is originally how the issue was discovered.
- ~~Also changing `destStageFlags = VK_PIPELINE_STAGE_ALL_GRAPHICS_BIT | VK_PIPELINE_STAGE_COMPUTE_SHADER_BIT;` to `destStageFlags = VK_PIPELINE_STAGE_ALL_COMMANDS_BIT;` to allow usage of the transfer queue as well. The PR was tested in an engine which exclusively creates textures (KTX or uncompressed) on the transfer queue before doing a queue family ownership transfer to graphics.~~ (erroneous, see below.)

This was tested on a video game environment with `385` KTX textures being loaded by 6 asset loading threads. Resolutions ranged from `5548x3636` to `32x32` with the across the board average being `1806.04x1755.35`.
The execution environment had the following hardware specs:
- Intel Core i5 10400F-2.9GHz
- 16GB of RAM
- RTX 2080Ti video card

Timing statistics of upload (including the queue guard) before the optimization (3 runs in ms):
```
min: 0.13 max: 1105.21 avg: 273.68
min: 10.21 max: 1254.09 avg: 247.52
min: 2.88 max: 869.47 avg: 239.28
```

Here are the same statistics collected after the optimization:
```
min: 0.12 max: 448.04 avg: 170.62
min: 1.20 max: 421.39 avg: 153.39
min: 1.18 max: 429.06 avg: 160.20
```

Here's a screenshot of said game environment:
image
More information can be provided on the application if requested.


MarkCallow:
> Also changing destStageFlags = VK_PIPELINE_STAGE_ALL_GRAPHICS_BIT | VK_PIPELINE_STAGE_COMPUTE_SHADER_BIT; to destStageFlags = VK_PIPELINE_STAGE_ALL_COMMANDS_BIT;

We used to use `VK_PIPELINE_STAGE_ALL_COMMANDS_BIT` but ran into a validation error in an update to the Vulkan SDK. This flags a genuine validation error. See issue #1092 and PR #1148 for details. What version of the Vulkan SDK are you using and are you testing with validation on?

lib/include/ktxvulkan.h
```diff
@@ -234,6 +237,32 @@ KTX_API void KTX_APIENTRY
...
+
+
+/**
+ * @class ktxVulkanTexture_queueGuardCallbacks
+ * @~English
+ * @brief Struct that contains callbacks necessary for guarding access to the supplied queue in vdi.
+ * 
+ *        Needs to be used in conjunction with suballocator callbacks that are also guarded for thread-safety. Together they can make UploadEx fully thread-safe in an efficient manner.
+ *        The suballocator callbacks need to guard around memory objects and if sparse binding support is introduced, will need to guard around both memory and queue objects.
+ *        As of now, guarding for external bookkeeping mechanisms are also a necessity in a threaded environment.
```
MarkCallow: Make the lines in this comment < 100 characters.

toomuchvoltage:
I'm using `1.4.313.2` and yes, around Vulkan code changes the validation layer is turned on to ensure that it's clean. I'll upgrade and try again. The solution really might be as simple as ORing with ` VK_PIPELINE_STAGE_TRANSFER_BIT`.

MarkCallow:
> I'm using 1.4.313.2

The improved validation was post 1.4.313. The reporter of issue #1092 was using 1.4.335.

toomuchvoltage:
> > I'm using 1.4.313.2
> 
> The improved validation was post 1.4.313. The reporter of issue #1092 was using 1.4.335.

You were right. This turned out to be unnecessary and I've reverted the change related to it. I realized I could just set the initial layout to `TRANSFER_DST` via the `UploadEx()` call and that would automatically line it up for access on the transfer queue. Subsequently, it could just be transitioned towards `SHADER_READ_ONLY` during the QFOT. The net result of all of this being illustrated here: https://github.com/toomuchvoltage/HighOmega-public/commit/3d27a3689716ee3e988c4149ab325b6209848292

toomuchvoltage:
Hi @MarkCallow I just added https://github.com/KhronosGroup/KTX-Software/pull/1231/changes/82a5bad1acf38a7154084eb7fe708bee79f0bef2 to demonstrate sample usage of the guarded callbacks. Truth is, it won't stress test the feature nor is that really feasible with the current single-texture test cases. Even if there were test cases requiring multiple textures, the guards within would need to be used application-wide where ever applicable. (i.e. if the graphics queue is creating textures, that would mean re-use for all accesses to graphics queue. Or any arena VkDeviceMemory accesses globally.)

If you feel like this is unnecessary, I can revert. Eager to hear back.

MarkCallow:
> If you feel like this is unnecessary, I can revert. Eager to hear back.

It is great to have a test even if it is not a stress test.

I would love to have non-interactive tests of the uploaders, maybe using gtest like `texturetests`, but I have no idea how to run such tests on GHA CI runners. Do they headless Vulkan or OpenGL graphics?

I will properly review this PR early next week. Please be aware that I will not merge this until v5.0.0 has been released. I can't give a date for that at present.

toomuchvoltage:
Hi @MarkCallow , just circling back on this. It's perfectly fine if this goes out post-5.0.0. Truth is these are on-the-field improvements resulting from a commercial game on Steam shipped with LibKTX2. I'm hesitant to link it since I personally wouldn't feel comfortable with the self promotion here, but of course figuring out the title is trivial given my handle. And I personally do not have experience with GHA CI, but I suspect paid plans (which this should be?) should have no issues with GPU'd instances.

tests/loadtests/vkloadtests/Texture.cpp
```diff
@@ -123,10 +125,26 @@ Texture::Texture(VulkanContext& vkctx,
...
+            std::thread uploaderThread([&]() {
+                ktxresult = ktxTexture_VkUploadEx_WithSuballocatorAndQueueGuard(kTexture, &vdi, &texture,
+                                                                                static_cast<VkImageTiling>(tiling),
+                                                                                VK_IMAGE_USAGE_SAMPLED_BIT,
+                                                                                VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL,
+                                                                                &subAllocatorCallbacksGuarded, &queueGuardCallbacks);
+            });
+            uploaderThread.join(); // We really have no choice here. This crucible/test-environment only loads one texture at a time.
+                                   // Additionally `QUEUE_GUARD_CALLBACKS::queueAccessGuard` and `QUEUE_GUARD_CALLBACKS::memoryAccessGuard` would
+                                   // need to be re-used where arena `VkDeviceMemory`s or the same `VkQueue`s are being accessed simultaneously.
```
MarkCallow: s/where/when/

MarkCallow (commented):
Is it necessary to use guarded memory allocation callbacks when using the queue guards

As I am no expert in this, I would like to find an expert to review it. From my side it looks fine except for a couple of minor comment issues.

toomuchvoltage:
Hi @MarkCallow , appreciate the feedback. The guarded memory callbacks are absolutely necessary. That said, I may have done a more heavy handed version than is necessary with VMA since I used my own pattern from my engine (which obviously needs to be more explicit): https://github.com/toomuchvoltage/HighOmega-public/blob/sauray_vkquake2/HighOmega/src/gl.cpp#L313-L474

VMA's allocation, image/buffer bind and free calls are all thread-safe. However, mapping and unmapping calls are not. They only check to ensure that no `VkDeviceMemory` arena is mapped twice. But no guarantees if you try to map the same one from separate threads (to the best of my understanding). The memory guard is also necessary when accessing `AllocMemCWrapperDirectory`, irrespective of whether VMA is used or not. I will be pushing an updated unit test addressing these.

For an expert pair of eyes, I would solicit Adam Sawicki's advice. He is the original author of VMA. His handle is @sawickiap on GitHub. I'm confident he's within reach for Khronos.

toomuchvoltage:
All done @MarkCallow , ready for another pair of eyes.

MarkCallow:
> All done @MarkCallow , ready for another pair of eyes.

Thanks. Working on finding a reviewer.

MarkCallow (approved):

MarkCallow:
One of my Khronos colleagues asked codex to analyze this. This is what it said.

----
- [P1] Queue locking does not make a shared ktxVulkanDeviceInfo thread-safe. The mutex is acquired only around vkQueueSubmit (https://github.com/KhronosGroup/KTX-Software/blob/339c04731e38806a288caf7ca4820accdfd08af6/lib/src/vkloader.c#L1317-L1320), but every upload begins and records into the single vdi->cmdBuffer (https://github.com/KhronosGroup/KTX-Software/blob/339c04731e38806a288caf7ca4820accdfd08af6/lib/src/vkloader.c#L1027-L1030). Concurrent calls sharing a vdi therefore concurrently access the same Vulkan command buffer, violating Vulkan’s external-synchronization requirements. Either use per-call/per-thread command buffers and command pools, or explicitly require each concurrent caller to supply a thread-confined vdi and command pool. Until then, the documentation’s “fully thread-safe (https://github.com/KhronosGroup/KTX-Software/blob/339c04731e38806a288caf7ca4820accdfd08af6/lib/include/ktxvulkan.h#L247-L251)” claim is unsafe.
  - [P2] Queue callbacks need state and preferably the queue handle. The callbacks take no arguments (https://github.com/KhronosGroup/KTX-Software/blob/339c04731e38806a288caf7ca4820accdfd08af6/lib/include/ktxvulkan.h#L137-L138), so a C application cannot associate a mutex with a particular VkQueue except through global state. A global mutex is correct but unnecessarily serializes independent queues, undermining the optimization this API introduces. I’d add void* userData to the callback struct and invoke callbacks with that and/or vdi->queue.
  - [P2] The “guarded” sample allocator contains a data race. mt64() is called before acquiring memoryAccessGuard (https://github.com/KhronosGroup/KTX-Software/blob/339c04731e38806a288caf7ca4820accdfd08af6/tests/loadtests/vkloadtests/VulkanLoadTestSample.cpp#L136-L164). std::mt19937_64 mutates internal state, so concurrent guarded allocations have undefined behavior and can produce corrupt or duplicate IDs. Generate the ID under the mutex, or preferably use a nonzero atomic counter.
  - [P2] The new load-error handling still performs incomplete cleanup. On ktxTexture_LoadImageData failure (https://github.com/KhronosGroup/KTX-Software/blob/339c04731e38806a288caf7ca4820accdfd08af6/lib/src/vkloader.c#L1170-L1180), the code unmaps and returns but leaks copyRegions, the staging buffer, and its allocation; it also leaves the command buffer recording. This needs a common cleanup path.
 
  Coverage note: the new “[threaded]” sample creates one worker and immediately joins it, so no accesses ever overlap (https://github.com/KhronosGroup/KTX-Software/blob/339c04731e38806a288caf7ca4820accdfd08af6/tests/loadtests/vkloadtests/Texture.cpp#L128-L140). It cannot detect the command-buffer or RNG races above. A deterministic two-caller test—using separate thread-confined VDIs if that is the intended contract—would be much more valuable.
----

@toomuchvoltage you have already pointed out the last item. It sounds like we need to look at the vdi queue handling. What do you think?

sawickiap:
Hi, I'm the developer of the VMA library. I'm sorry for the delayed response.

Mapping in VMA is thread-safe.

About raw Vulkan (functions `vkMapMemory`, `vkUnmapMemory`), you are right:
- Mapping a single `VkDeviceMemory` block multiple times (whether the same or disjoint memory regions) is illegal.
- It is not thread-safe.

However, using the recommended library functions `vmaMapMemory`, `vmaUnmapMemory` (or other convenient ways like `VMA_ALLOCATION_CREATE_MAPPED_BIT` flag or `vmaCopyMemoryToAllocation` function):
- They are thread-safe, synchronized internally (unless you specified `VMA_ALLOCATOR_CREATE_EXTERNALLY_SYNCHRONIZED_BIT`).
- You can map the same allocation or different allocations coming from the same `VkDeviceMemory` block multiple times (mapping is reference-counted internally).
- You can safely map different allocations coming from the same `VkDeviceMemory` block because VMA internally maps the entire memory block, not just the region of that one allocation.

For more information, see this documentation chapter:
https://gpuopen-librariesandsdks.github.io/VulkanMemoryAllocator/html/memory_mapping.html

MarkCallow:
Thanks @sawickiap. @toomuchvoltage do you have any comments on this or the codex review?

toomuchvoltage:
Hi @MarkCallow @sawickiap , deeply appreciate the reviews. Fantastic to be informed about `vmaMapMemory` and `vmaUnmapMemory` being thread-safe.

On the P1 issue: once again, since I was applying the learning from my own engine, I simply brought the assumptions as well for general use. The point is absolutely correct and my engine has a thread-safe cache for per-thread KTX VDIs.

Thread-safe objects:
https://github.com/toomuchvoltage/HighOmega-public/blob/sauray_vkquake2/HighOmega/src/gl.cpp#L52-L53

Creation:
https://github.com/toomuchvoltage/HighOmega-public/blob/sauray_vkquake2/HighOmega/src/gl.cpp#L4510-L4519

Destruction:
https://github.com/toomuchvoltage/HighOmega-public/blob/sauray_vkquake2/HighOmega/src/gl.cpp#L4166-L4172

I guess this becoming the general usage pattern was implicit in my assumptions. We can ask for assurances on this in the documentation.

On the first P2 issue: my engine effectively has 1 copy queue, 1 transfer queue and 1 graphics queue. Is it safe to assume that this _is_ a globally recommended pattern? Is it safe or performant for other engines to have multiple transfer queues for example? If so, we can simply pass a typed `VkQueue` to the function where there's no ambiguity about what is we want guarded.

On the second P2 issue: Very fine point regarding `mt64()`. I'll address that.

On the third P2 issue: Good point, I guess this really falls under an overall discovery in this process that can be addressed in this PR.

I'll get another commit together to address these.

toomuchvoltage:
Hi @MarkCallow , I just pushed a commit addressing the raised issues. Note that I finally decided to pass a `VkQueue` handle to the queue guard callbacks. This makes their roles much more explicit. Additionally, from what I gather online not all IHVs support multiple queues per family. And in practice it may not yield a performance benefit in all cases. (Knowing this is naturally compatible with the design we have.) It would be interesting to verify this with Khronos's perhaps broader overview of what is actually available in the wild.

MarkCallow:
Perhaps https://vulkan.gpuinfo.org can answer your questions about support for multiple queues per family.

In one of your commit messages you write that a per-thread VDI is necessary. If that is the case does it not make sense to pass the device from that to the allocator callbacks, similar to what was requested in issue #1212? Doing so would require changing the signatures of the allocator callback functions. I wonder if there is any way to do that without having to create another set of `*_VkUploadEx_WithSubAllocator` functions.

lib/src/vkloader.c
```diff
@@ -799,7 +799,8 @@ linearTilingPadCallback(int miplevel, int face,
  *                                        (or additionally queue if sparse binding support
  *                                        is added) and external suballocation managements
  *                                        objects, it can make UploadEx fully thread-safe
- *                                        and efficiently so.
+ *                                        and efficiently so. This is assuming that a
+ *                                        per-thread VDI is used as well.
```
MarkCallow: Is a per-thread VDI needed for thread-safe suballocator use too?

lib/src/vkloader.c
```diff
@@ -1172,10 +1173,21 @@ ktxTexture_VkUploadEx_WithSuballocatorAndQueueGuard(ktxTexture* This, ktxVulkanD
                                       (ktx_size_t)memAllocInfo.allocationSize);
                 if (kResult != KTX_SUCCESS)
                 {
+                    // TODO: This should be a common clean-up path for all failure cases,
```
MarkCallow: Please make it so.

toomuchvoltage:
> In one of your commit messages you write that a per-thread VDI is necessary. If that is the case does it no make sense to pass the device from that to the allocator callbacks, similar to what was requested in issue #1212? Doing so would require changing the signatures of the allocator callback functions. I wonder if there is any way to do that without having to create another set of `*_VkUploadEx_WithSubAllocator` functions.

Hi @MarkCallow, no that is not necessary. `ktxTexture_VkUploadEx()`, `ktxTexture_VkUploadEx_WithSuballocator()` and `ktxTexture_VkUploadEx_WithSuballocatorAndQueueGuard()` all readily take in a KTX VDI (second parameter). In fact the quoted snippets demonstrate how a per thread KTX VDI is provided in a thread-safe manner:

> Thread-safe objects: https://github.com/toomuchvoltage/HighOmega-public/blob/sauray_vkquake2/HighOmega/src/gl.cpp#L52-L53
> 
> Creation: https://github.com/toomuchvoltage/HighOmega-public/blob/sauray_vkquake2/HighOmega/src/gl.cpp#L4510-L4519
> 
> Destruction: https://github.com/toomuchvoltage/HighOmega-public/blob/sauray_vkquake2/HighOmega/src/gl.cpp#L4166-L4172
> 

Each `ktx2VDIPools.dir[ThreadID].elem` (with `ThreadID` being `thread_local` and generated with `mt64()`) is a per-thread KTX VDI supplying its own command pool and command buffer to `UploadEx()`. In fact these elements are reference counted via `ktx2VDIPools.dir[ThreadID].elemCount` perhaps much in the same manner and spirit that VMA handles maps/unmaps internally. Once `elemCount` goes to zero, the KTX VDI is deemed unnecessary and destroyed (i.e. all last images created with this VDI are gone). These operations are synchronized via `ktx2VDIPools.mtx`. My own status quo usage issues with `mt64()` not-withstanding, this is effectively sufficient for our `UploadEx()` set of calls to complete their tasks efficiently and in a thread-safe manner. This solution is currently shipped and tested in the video game I screenshotted in the original post (coupled with performance metrics and the environment specs). It shares the same core engine with the repo posted above.

Ultimately, what is being proposed in https://github.com/KhronosGroup/KTX-Software/issues/1212 breaks separation of concerns. A type erased `void *` being passed in would leave the actual implementation guessing as to what it even means in the first place. I obviously welcome the original author of 1212 to fork and create their own interface and implementation. However, I would be very hesitant to modify the standard version advertised globally taking in nebulous type-erased parameters with no well defined purpose.

I will proceed to prepare a commit for exhaustive clean-ups in all failure cases of `UploadEx()` shortly.

lib/src/vkloader.c
toomuchvoltage: Yes. However, an additional parameter is wholly unnecessary and the status quo interface can readily satisfy this requirement as linked in earlier comments.

toomuchvoltage:
All done @MarkCallow . A quick review of the specification for `freeUploadResources()` would be appreciated to see if it is in line with other specifications. Note that it is used to clean-up `mappableMemory` and the texture's main `allocationId` rather than any staging resources in the linear tiling case (as staging resources do not seem to apply there).

EDIT: Force pushed to get the checks running again. There was a test infrastructure failure and I wanted to make sure I'm not introducing new issues with this commit.

toomuchvoltage:
Hi @MarkCallow , I thought I needed to address the point a bit more directly in the interest of the library. Sorry if we're already discussed this a fair bit.

> If that is the case does it not make sense to pass the device from that to the allocator callbacks, similar to what was requested in issue #1212?

The callbacks that are designed at the moment -- dating back to the original suballocator PR -- are shims around very specific operations that are either doable directly in Vulkan (with some sophistication) or have VMA exposed calls (such as the ones you're seeing in the tests: alloc, bind-image/buffer, map/unmap, free). Basic usage for VMA has none of that. It's basically just `vmaCreateBuffer()`, `vmaCreateImage()`. See: https://gpuopen-librariesandsdks.github.io/VulkanMemoryAllocator/html/quick_start.html . I'm willing to bet it was Adam's foresight that resulted in the exposed calls that we're featuring in our samples today. Or possibly industry feedback.

Resultingly, the parameters that were chosen for those callbacks are to enable those operations. Nothing more. They should -- at least as far as the standard presented to world -- not be designed to be a vehicle for brokering external data. Allocation IDs and page counts suffice. Basic Vulkan types coupled with those are sufficient for those calls to conclude their tasks effectively.

toomuchvoltage:
Amending new links since a recent push shifted line numbers:

Thread-safe objects: https://github.com/toomuchvoltage/HighOmega-public/blob/sauray_vkquake2/HighOmega/src/gl.cpp#L53-L53

Creation: https://github.com/toomuchvoltage/HighOmega-public/blob/sauray_vkquake2/HighOmega/src/gl.cpp#L4566-L4575

Destruction: https://github.com/toomuchvoltage/HighOmega-public/blob/sauray_vkquake2/HighOmega/src/gl.cpp#L4231-L4237

EDIT: they shifted again due to incorporating all feedback from this PR.

toomuchvoltage:
Hi again @MarkCallow , hope all is well. Just circling back to see if there's additional feedback. I actually pushed a commit to my own repo that now incorporates all feedback from this PR: https://github.com/toomuchvoltage/HighOmega-public/commit/085bc1db2c9c8f0131e998f65baa662900db0bf4 (both thread-safe `mt64()` and VMA-style thread-safe maps/unmaps are demonstrated, the latter now makes failures between maps/unmaps inside `UploadEx()` much less dangerous).

Eager to take note of any further potential refinement suggestions.

lib/src/vkloader.c
```diff
@@ -735,6 +735,38 @@ linearTilingPadCallback(int miplevel, int face,
...
+        VK_CHECK_RESULT(vdi->vkFuncs.vkEndCommandBuffer(vdi->cmdBuffer));
+    }
+    if (copyRegions && *copyRegions) {
+        free(*copyRegions);
+        *copyRegions = NULL;
+    }
+    if (hasAllocated) {
+        if (!useSuballocator && vdi) {
+            vdi->vkFuncs.vkFreeMemory(vdi->device, stagingMemory, vdi->pAllocator);
+        } else if (subAllocatorCallbacks) {
```
MarkCallow, line 761: Is `useSubAllocator && !subAllocatorCallbacks` possible?

lib/src/vkloader.c
```diff
@@ -794,6 +826,13 @@ linearTilingPadCallback(int miplevel, int face,
  *                                        They use a uint64_t stored in the @c allocationId
  *                                        field of the structure pointed at by @a vkTexture
  *                                        to reference allocated page(s).
+ * @param [in] queueMutexCallbacks        If used in conjunction with suballocator callbacks
+ *                                        that guard against simultaneous access to memory
+ *                                        (or additionally queue if sparse binding support
+ *                                        is added) and external suballocation managements
+ *                                        objects, it can make UploadEx fully thread-safe
+ *                                        and efficiently so. This is assuming that a
+ *                                        per-thread VDI is used as well.
```
MarkCallow: I am having trouble understanding this description.

Is it the suballocator callbacks that guard against ... (and must they guard against) or is it the queueMutexCallbacks that guard against ....  If it is the suballocator callbacks is the guard necessary only if queueMutexCallbacks are also used? I also don't understand the comment in parentheses about sparse binding support.

I think "managements" should probably be "management".

Instead of "UploadEx" say "this function".

Does "per-thread VDI" need more prominence? If not using suballocator and queueMutex callbacks can UploadEx be used from multiple threads, provided a per-thread VDI is used? If so this should be stated at the description of the vdi parameter.

MarkCallow:
I've been distracted by other work. I'm sorry for the delay.

> A quick review of the specification for freeUploadResources() would be appreciated to see if it is in line with other specifications.

I don't think we have any other similar functions. It seems fine though for some of the calls it does nothing.

> If that is the case does it not make sense to pass the device from that to the allocator callbacks, similar to what was requested in issue https://github.com/KhronosGroup/KTX-Software/issues/1212?

Note that I was not proposing a "type-erased" parameter. I was proposing passing a VkDevice. You have demonstrated it is not necessary so we will leave things as they are.

lib/src/vkloader.c
toomuchvoltage: Only if someone use `freeUploadResources()` in a seriously erroneous fashion. `useSuballocator` is set to `true`, not only when `subAllocatorCallbacks` is non-`NULL`, but when all its member function pointers are also non-`NULL`.

lib/src/vkloader.c
toomuchvoltage: > Is it the suballocator callbacks that guard against ... (and must they guard against) or is it the queueMutexCallbacks that guard against .... If it is the suballocator callbacks is the guard necessary only if queueMutexCallbacks are also used?

if `queueMutexCallbacks` are used, it implies a threaded environment. In that event, the suballocator callbacks _better_ be guarded as well, since simultaneous accesses to the same `VkDeviceMemory` objects require synchronization as per the specs. As noted earlier, VMA handles this internally unless explicitly requested not to. Any other implementation needs to be as cautious. Regardless, accesses to the suballocation directory will need synchronization themselves.

>  I also don't understand the comment in parentheses about sparse binding support.

Calls to `vkQueueBindSparse()` will simultaneously access the queue and memory objects. Currently, `UploadEx()`'s implementation does not support sparse bindings but the interface leaves that door open. My engine, HighOmega, does actually support sparse bindings under the hood and a potentially sparse binding capable version of `UploadEx()` can call into these implementations:

**Buffers:**
https://github.com/toomuchvoltage/HighOmega-public/blob/master/HighOmega/src/gl.cpp#L346-L376

**Images:**
https://github.com/toomuchvoltage/HighOmega-public/blob/master/HighOmega/src/gl.cpp#L396-L426

> Does "per-thread VDI" need more prominence? If not using suballocator and queueMutex callbacks can UploadEx be used from multiple threads, provided a per-thread VDI is used? If so this should be stated at the description of the vdi parameter.

No, it will be either quite dangerous or pointlessly inefficient. If indeed it is an environment with queues shared between uploaders, not guarding accesses to them will violate the spec and cause all sorts of behind the scenes issues. Otherwise, the upload calls are effectively blocking each other which is the status quo. I'll provide more prominence to this especially around the VDI parameter description.

> 
> I think "managements" should probably be "management".
> 
> Instead of "UploadEx" say "this function".
> 
I'll prepare a commit addressing these.

toomuchvoltage:
> Note that I was not proposing a "type-erased" parameter. I was proposing passing a VkDevice. You have demonstrated it is not necessary so we will leave things as they are.

Truth is there's a probably a slight bit of incomplete abstraction with the current callbacks around `VkDevice`. I wasn't thinking of a multi-GPU environment. That said, simply making the following `thread_local` satisfies the ability to upload different textures to different devices under the same Vulkan instance.

https://github.com/toomuchvoltage/HighOmega-public/blob/master/HighOmega/src/gl.cpp#L318

The reality is that the decision of the target device will be made around to which workload the texture belongs. That in all likelihood will happen long before `UploadEx()` even affecting which VDI to use since the image and the backing memory need to be on the same device as per the spec. That decision point can change `deviceCached` for the thread and since the callbacks will happen on the same thread, the right value will be visible to them.

lib/src/vkloader.c
```diff
@@ -810,7 +810,11 @@ freeUploadResources(ktxVulkanDeviceInfo* vdi, VkBufferImageCopy** copyRegions, b
  * @param[in] This                        pointer to the ktxTexture from which to upload.
  * @param [in] vdi                        pointer to a ktxVulkanDeviceInfo structure providing
  *                                        information about the Vulkan device onto which to
- *                                        load the texture.
+ *                                        load the texture. If subAllocatorCallbacks and
```
MarkCallow: This still does not seem clear enough to me. How about

 "If calling this function from multiple threads, a per-thread value is required for this parameter.
This will be the case when subAllocatorCallbacks and queueMutexCallbacks are provided but normally a single thread is used."

lib/src/vkloader.c
```diff
@@ -829,10 +833,10 @@ freeUploadResources(ktxVulkanDeviceInfo* vdi, VkBufferImageCopy** copyRegions, b
  * @param [in] queueMutexCallbacks        If used in conjunction with suballocator callbacks
  *                                        that guard against simultaneous access to memory
  *                                        (or additionally queue if sparse binding support
- *                                        is added) and external suballocation managements
- *                                        objects, it can make UploadEx fully thread-safe
- *                                        and efficiently so. This is assuming that a
- *                                        per-thread VDI is used as well.
+ *                                        is added) and external suballocation management
```
MarkCallow: I am still confused by the part in parentheses.  "Additionally queue" starts the confusion. Do you mean that the suballocator callbacks must guard against simultaneous access to memory and, if sparse binding support is added,  the queue? However isn't guarding the queue what the queueMutexCallbacks are for.

Please clarify.

MarkCallow (commented):
The new comments still confuse me.

lib/src/vkloader.c
toomuchvoltage: Sure thing, though I would just replace the word 'normally' with 'in simple use cases'. Or potentially remove anything after 'normally' if that's ok. My reasoning is that if I were to not use these, load times for the game this has been tested on would balloon and I'd wager that this would be the case in other applications. End goal is to prepare LibKTX2 for much more demanding streaming applications. Let me know your thoughts on these suggestions.

lib/src/vkloader.c
toomuchvoltage: Yes, that is the implication. `queueMutexCallbacks` effectively guards for queue accesses _inside_ `UploadEx()`. However, the comment is simply a hint to the developer that they will still be responsible for guarding against the queue _outside_ of `UploadEx()` in their own callbacks. Looking at:

https://github.com/toomuchvoltage/HighOmega-public/blob/master/HighOmega/src/gl.cpp#L371-L374

The ordering between locking the memory mutex and the queue mutex are intentional and maintained throughout the application to avoid potential deadlocks. Normally, this would call for a `std::scoped_lock`. However, this may not even be possible at all times due to locks happening in nested calls.

Eager for further hints on how to better elaborate this.

toomuchvoltage:
Hi @MarkCallow , just responded to the comments. Would appreciate further guidance.

lib/src/vkloader.c
MarkCallow: Replacing "normally" with "in simple use cases" is fine.

> End goal is to prepare LibKTX2 for much more demanding streaming applications.

Are you aware of issue #1224 where work is underway to support progressive downloading.

lib/src/vkloader.c
MarkCallow: How about

"pointer to a set of queue-mutex callbacks which the function
uses to protect accesses to the queue. When used in conjunction 
with suballocator callbacks that guard against simultaneous 
access to memory and suballocation management objects (which is
the case if using VMA) these will make the function fully thread-safe
and efficiently so. Note that if sparse binding support is added the
suballocator callbacks must also guard against simultaneous access
to the queue. Note also that a per-thread VDI is necessary when
using these callbacks."

toomuchvoltage: Looks good, I would simply remove '(which is the case if using VMA)'. So far, both the VMA examples I've authored and my own code have necessitated those objects. They're basically a necessity no matter the underlying implementation.

"Pointer to a set of queue-mutex callbacks which the function
uses to protect accesses to the queue. When used in conjunction
with suballocator callbacks that guard against simultaneous
access to memory and suballocation management objects, these 
will make the function fully thread-safe and efficiently so.
Note that if sparse binding support is added, the suballocator 
callbacks must also guard against simultaneous access to the 
queue. Note also that a per-thread VDI is necessary when using 
these callbacks."

lib/src/vkloader.c
toomuchvoltage: I still haven't fully explored everything in [#1224 ](https://github.com/KhronosGroup/KTX-Software/issues/1224) but it's fascinating at a cursory glance. I see that residency management on the consumer's part has been noted as needing care in scenarios like this, and that is exactly where calls like `vkQueueBindSparse()` come into play. I would most certainly review this page as it has extensive information on how all of this would need to play out: https://www.asawicki.info/news_1698_vulkan_sparse_binding_-_a_quick_overview

Overall, I see both the suballocator and queue mutex callbacks as still being necessary additions to efficient uploads in addition to streaming and transcoding levels on the fly. I see [#1224 ](https://github.com/KhronosGroup/KTX-Software/issues/1224) as paving the way for features similar to Streaming Mips and even Streaming Virtual Textures found in UnrealEngine.

toomuchvoltage:
Hi @MarkCallow just committed a change based on the documentation feedback. I still haven't fully read up on https://github.com/KhronosGroup/KTX-Software/issues/1224 but it's very intriguing at a cursory glance.

MarkCallow:
The new comments are good. Thanks.

Unfortunately I am still not in a position to be able to merge this to main. As soon as v5.0.0 is released I will.

MarkCallow (approved):

toomuchvoltage:
Thank you @MarkCallow . Totally understandable, no rush.

PR #1237, opened 2026-08-05
Fix SGD image desc indexing for 3D UASTC HDR 6x6i textures

SashaRX:
Fixes transcoding of 3D textures encoded to UASTC HDR 6x6 intermediate.

#### Problem

`transcodeUastcHDR6x6_intermediate` indexes the SGD image description table with

```
sgdImageDescIndex = level * levelImageCount + image
```

using the *current* level's image count. The encoder (`basis_encode.cpp`) writes the `ktxUASTCHDR6x6IntermediateImageDesc` entries in level order, level 0 first, with each level contributing `numLayers × numFaces × depth(level)` descriptions. The two agree only while every level has the same image count. For 3D textures depth halves with each mip level, so the transcoder selects descriptions belonging to other levels; since slice offsets are relative to their own level's data, transcoding fails with `KTX_TRANSCODE_FAILED` (or could silently read wrong slices when the offsets happen to stay in range).

This reproduces with libktx's own encoder output: encode a 24×24×8 `VK_FORMAT_R16G16B16A16_SFLOAT` texture with 4 mip levels via `ktxTexture2_CompressBasisEx` with `KTX_BASIS_CODEC_UASTC_HDR_6x6_INTERMEDIATE`, then `ktxTexture2_TranscodeBasis(..., KTX_TTF_ASTC_HDR_6x6_RGBA, 0)` fails. A 2D texture with otherwise identical parameters transcodes fine — presumably why this hasn't been noticed.

#### Fix

Accumulate per-level first-image indices before the level loop and index with `firstImages[level] + image` — the same approach `transcodeEtc1s` already uses with its `firstImages` table. No behavior change for textures where every level has the same image count (2D, arrays, cubemaps): for those `firstImages[level] == level * levelImageCount`.

Since `firstImages[numLevels]` is the image count implied by the texture's dimensions, the descriptor table size is now also validated up front: a table that is not exactly that many descriptions, or not a whole number of them, is rejected with `KTX_FILE_DATA_ERROR` before any level is processed, instead of surfacing as `KTX_TRANSCODE_FAILED` per image.

#### Tests

Adds generated 2D control and 3D multi-mip round-trip regression tests to `transcodetests` (in-process encode → transcode), with no new binary test resources. The 3D case fails without the fix and passes with it; both verify the post-transcode state (`vkFormat`, supercompression scheme, image data present).

Found while studying the transcoder for the per-level streaming work discussed in #1224.

### DIFF
--- a/lib/src/basis_transcode.cpp
+++ b/lib/src/basis_transcode.cpp
@@ -928,6 +928,30 @@ transcodeUastcHDR6x6_intermediate(ktxTexture2* This, alpha_content_e alphaConten
         reinterpret_cast<ktxUASTCHDR6x6IntermediateImageDesc*>(This->_private->_supercompressionGlobalData);
     const uint64_t totalImageDescs = This->_private->_sgdByteLength / sizeof(ktxUASTCHDR6x6IntermediateImageDesc);
 
+    // The image descriptions are stored in level order, level 0 first, with
+    // each level contributing numLayers * numFaces * depth(level) images (see
+    // the writer in basis_encode.cpp). level * levelImageCount only equals
+    // the index of a level's first description while every level has the
+    // same image count; for 3D textures depth halves with each level, so the
+    // first-image index of each level must be accumulated, as transcodeEtc1s
+    // does with its firstImages table.
+    std::vector<uint64_t> firstImages(This->numLevels + 1);
+    firstImages[0] = 0;
+    for (uint32_t l = 0; l < This->numLevels; l++) {
+        firstImages[l + 1] = firstImages[l]
+                             + (uint64_t)This->numLayers * This->numFaces
+                               * MAX(This->baseDepth >> l, 1);
+    }
+
+    // firstImages[numLevels] is the image count the texture's dimensions
+    // call for, so a descriptor table whose size does not match exactly is
+    // corrupt; reject it before processing any level.
+    if (This->_private->_sgdByteLength
+            % sizeof(ktxUASTCHDR6x6IntermediateImageDesc) != 0
+        || firstImages[This->numLevels] != totalImageDescs) {
+        return KTX_FILE_DATA_ERROR;
+    }
+
     for (ktx_int32_t level = This->numLevels - 1; level >= 0; level--) {
         ktx_uint32_t depth;
         uint64_t writeOffset = levelOffsetWrite;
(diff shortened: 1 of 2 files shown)


MarkCallow:
Excellent catch. Thanks. As with PR #1236 , please sign the CLA. It should be sufficient to do it in just one PR.

lib/src/basis_transcode.cpp
```diff
@@ -928,6 +928,30 @@ transcodeUastcHDR6x6_intermediate(ktxTexture2* This, alpha_content_e alphaConten
...
+    // The image descriptions are stored in level order, level 0 first, with
+    // each level contributing numLayers * numFaces * depth(level) images (see
+    // the writer in basis_encode.cpp). level * levelImageCount only equals
+    // the index of a level's first description while every level has the
+    // same image count; for 3D textures depth halves with each level, so the
+    // first-image index of each level must be accumulated, as transcodeEtc1s
+    // does with its firstImages table.
+    std::vector<uint64_t> firstImages(This->numLevels + 1);
+    firstImages[0] = 0;
+    for (uint32_t l = 0; l < This->numLevels; l++) {
```
MarkCallow: Should be

```
for (uint32_t l = 1; l <= This->numLevels; l++) {
```

lib/src/basis_transcode.cpp
```diff
@@ -928,6 +928,30 @@ transcodeUastcHDR6x6_intermediate(ktxTexture2* This, alpha_content_e alphaConten
...
+        firstImages[l + 1] = firstImages[l]
+                             + (uint64_t)This->numLayers * This->numFaces
+                               * MAX(This->baseDepth >> l, 1);
+    }
+
+    // firstImages[numLevels] is the image count the texture's dimensions
```
MarkCallow: // firstImages[numLevels] has the total image count for the texture's dimensions
// so a descriptor table ...

MarkCallow (commented):
I won't run the workflows until you fix the incorrect for loop.

SashaRX:
Done — switched the accumulation to the same `1..numLevels` loop form `transcodeEtc1s` uses and applied your comment wording. Both regression tests still pass.

lib/src/basis_transcode.cpp
```diff
@@ -928,6 +928,30 @@ transcodeUastcHDR6x6_intermediate(ktxTexture2* This, alpha_content_e alphaConten
...
+    // The image descriptions are stored in level order, level 0 first, with
+    // each level contributing numLayers * numFaces * depth(level) images (see
+    // the writer in basis_encode.cpp). level * levelImageCount only equals
+    // the index of a level's first description while every level has the
+    // same image count; for 3D textures depth halves with each level, so the
+    // first-image index of each level must be accumulated, as transcodeEtc1s
+    // does with its firstImages table.
+    std::vector<uint64_t> firstImages(This->numLevels + 1);
+    firstImages[0] = 0;
+    for (uint32_t l = 1; l <= This->numLevels; l++) {
```
MarkCallow: Apologies. I did not see the `l + 1` in your original which is actually better than what I asked you to change it to because there is only one `l + 1` vs. two `l - 1`s in this. Feel free to change it back and change the ETC1S transcoder too.

SashaRX: Done — restored the l + 1 form and restyled the ETC1S transcoder's table to match, as suggested. Tests still pass.

MarkCallow (approved):

PR #1238, opened 2026-08-05
feat: capture mipmap generation options in writer metadata params

### LINKED ISSUE #1196: Feature request: record mipmap generation options for ktx create/encode
Since I'm still working on OIIO KTX2 support, regenerating as-close-as-possible of the original KTX2 input to OIIO is crucial. One of the aspects is to re-generate the same  (as usual, as close as possible given the lossy nature) mipmaps.

`ktx create` (probably also `ktx encode`) does not record mipmap re-generation options so the following command doesn't save any metadata about the used mipmap filter (among other options as well).

```bash
ktx create --testrun --format R8G8B8_SRGB --generate-mipmap --mipmap-filter quadratic_approx tests/resources/input/png/kodim17.png kodim17_quadratic_approx_filer.ktx2
```

If this is reasonable, I will open a PR + CTS PR to add this.

For the moment, from the OIIO side, I just default to the same defaults used by ktx tools for mipmap regeneration (same with all other defaults, e.g., for UASTC).

I might open quite a lot of issues since I deal with a lot of edge cases (however, no pressure at all to address these as soon as possible).

Edit: wrongly used `--runtime-mipmap` at first. Issue still stands.

walcht:
* Fixes #1196

* This commit captures mipmap generation options in KTXWriterScParams metadata entry only if mipmap generation is enabled. This might help ktx2 file consumers to, for instance, regenerate the mipmaps using the same params.

### DIFF
--- a/tools/ktx/command_create.cpp
+++ b/tools/ktx/command_create.cpp
@@ -2056,13 +2056,17 @@ void CommandCreate::executeCreate() {
 
     compress(texture, options);
 
-    // We don't want to capture mipmap generation options if mipmap generation is not activated
-    // (because looking at kv metadata might give the wrong impression that mipmaps were generated)
-    std::string mipmapGenerationOptionsCpy;
-    if (options.mipmapGenerate) mipmapGenerationOptionsCpy = options.mipmapGenerationOptions;
-
     // Add KTXwriterScParams metadata if ASTC encoding, BasisU encoding, or other supercompression was used
-    const auto writerScParams = fmt::format("{}{}{}{}{}", options.astcOptions, options.codecOptions, mipmapGenerationOptionsCpy, options.commonOptions, options.compressOptions);
+    // Note:
+    //  We don't want to capture mipmap generation options if mipmap generation
+    //  is not activated (because looking at kv metadata might give the wrong
+    //  impression that mipmaps were generated). This is the same for ASTC.
+    const auto writerScParams = fmt::format("{}{}{}{}{}",
+        options.encodeASTC ? options.astcOptions : "",
+        options.codecOptions,
+        options.commonOptions,
+        options.mipmapGenerate ? options.mipmapGenerationOptions : "",
+        options.compressOptions);
     if (writerScParams.size() > 0) {
         // Options always contain a leading space
         assert(writerScParams[0] == ' ');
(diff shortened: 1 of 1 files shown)


walcht:
CTS PR: https://github.com/KhronosGroup/KTX-Software-CTS/pull/82

The MingW CI is failing because I pushed CTS after pushing this PR; again, I keep forgetting to push CTS first ...

tools/ktx/command_create.cpp
```diff
@@ -2039,8 +2056,13 @@ void CommandCreate::executeCreate() {
...
     compress(texture, options);
 
+    // We don't want to capture mipmap generation options if mipmap generation is not activated
+    // (because looking at kv metadata might give the wrong impression that mipmaps were generated)
+    std::string mipmapGenerationOptionsCpy;
+    if (options.mipmapGenerate) mipmapGenerationOptionsCpy = options.mipmapGenerationOptions;
+
     // Add KTXwriterScParams metadata if ASTC encoding, BasisU encoding, or other supercompression was used
-    const auto writerScParams = fmt::format("{}{}{}{}", options.astcOptions, options.codecOptions, options.commonOptions, options.compressOptions);
+    const auto writerScParams = fmt::format("{}{}{}{}{}", options.astcOptions, options.codecOptions, mipmapGenerationOptionsCpy, options.commonOptions, options.compressOptions);
```
MarkCallow: I think `commonOptions` are the common encoding options so should be right after `codecOptions`. Move mipmapGenerationOptionsCpy after `commonOptions`.

I suspect that the reasons you created `mipmapGenerationOptionsCpy` exist for these other options too. There are options that are ignored unless a certain codec is specified. Should we fix those similarly? As far as I am aware, it has not been a problem.

B.t.w with regards to the OIIO reader, there are other tools used to create .ktx2 files so you cannot rely on having this information.

walcht:
> I suspect that the reasons you created mipmapGenerationOptionsCpy exist for these other options too.

Yes (to some degree). I enabled this for ASTC as well. For Basis Universal codecs it requires more work which is why I just ignored it.

> B.t.w with regards to the OIIO reader, there are other tools used to create .ktx2 files so you cannot rely on having this information.

Yes. This is just a heuristic for a rare use case that is advised against (unless input KTX2 file is not lossy compressed which is probably very rare).

PR #1240, opened 2026-08-06
fix: metrics output to stderr if stdout is in use

### LINKED ISSUE #1233: cli: compare-psnr doesn't work when writing ktx2 file to stdout
To replicate:

```bash
ktx create --format ASTC_4x4_UNORM_BLOCK kodim23.png --compare-psnr - | ktx extract - test.png
```

This errors with this message:

```bash
fatal-2001: Not a KTX2 file.
    Not a KTX2 file. The beginning of the file does not match the expected file identifier "«KTX 20»\r\n\x1A\n".
```

But when ran as follows (without writing to stdout) it works:

```bash
ktx create --format ASTC_4x4_UNORM_BLOCK kodim23.png --compare-psnr test.ktx2
```

As usual, I will open a PR if this an actual issue.

walcht:
* PSNR and SSIM metrics now ouput to stderr if stdout is in use. This is done to avoid polluting KTX file ouput to stdout which causes issues when, for instance, output gets piped to ktx extract tool.
* Fixes https://github.com/KhronosGroup/KTX-Software/issues/1233

### DIFF
--- a/tools/ktx/command_create.cpp
+++ b/tools/ktx/command_create.cpp
@@ -2035,7 +2035,13 @@ void CommandCreate::executeCreate() {
     if (options.encodeASTC)
         encodeASTC(texture, options);
 
-    metrics.decodeAndCalculateMetrics(texture, options, *this);
+    OutputStream outputFile(options.outputFilepath, *this);
+    const auto outputPath = std::filesystem::path(DecodeUTF8Path(options.outputFilepath));
+    if (outputPath.has_parent_path())
+        std::filesystem::create_directories(outputPath.parent_path());
+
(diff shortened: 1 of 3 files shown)


MarkCallow:
Windows, MingW and macOS all have test failures, the first two, many, the last, one. The other failures look like some GitHub runner issue. I have triggered rebuilds of those.

Linux too, now that it has got past the runner issue, has many test failures.

walcht:
I forgot to move parent folders creation before creating the `outputFile` stream. My local tests passed because I ran the tests before on main branch which created the parent folders.

Anyway, it should be all good now.

MarkCallow (approved):

PR #1245, opened 2026-09-05
VK_IMAGE_LAYOUT_GENERAL support for image layout

### LINKED ISSUE #1244: Support VK_IMAGE_LAYOUT_GENERAL as a final layout for Vulkan texture upload
The texture upload path accepts a finalLayout parameter, where it converts to the corresponding layout using a barrier after the image is created. However, currently it does not support VK_IMAGE_LAYOUT_GENERAL, where if set it will trigger a false assertion. I'm running into this issue when I'm working on a project that has VK_KHR_unified_image_layouts enabled, so it uses VK_IMAGE_LAYOUT_GENERAL for every image layout. Although I could first create the image into another layout, and then convert the layout by myself, but it's quite breaking the "only general layout" rule for that project, and I also think it would be nice if we could have VK_IMAGE_LAYOUT_GENERAL be an option for Vulkan texture upload.

OppositeNor:
Resolves #1244

Since VK_IMAGE_LAYOUT_GENERAL could be used anywhere, we shouldn't assume how it's accessed either before/after the transition. So, we're using the most conservative VK_ACCESS_MEMORY_READ_BIT | VK_ACCESS_MEMORY_WRITE_BIT for both.

- Added VK_IMAGE_LAYOUT_GENERAL as an image layout option for oldLayout and newLayout for setImageLayout,

### DIFF
--- a/lib/src/vkloader.c
+++ b/lib/src/vkloader.c
@@ -1748,6 +1748,12 @@ setImageLayout(
         imageMemoryBarrier.srcAccessMask = VK_ACCESS_SHADER_READ_BIT;
         break;
 
+    case VK_IMAGE_LAYOUT_GENERAL:
+        // Image may be used for arbitrary read/write access.
+        // Make sure any reads and writes to the image have finished.
+        imageMemoryBarrier.srcAccessMask = VK_ACCESS_MEMORY_READ_BIT | VK_ACCESS_MEMORY_WRITE_BIT;
+        break;
+
     default:
         /* Value not used by callers, so not supported. */
         assert(KTX_FALSE);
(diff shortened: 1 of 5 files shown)


MarkCallow:
Thanks. Please add the same change to `setImageLayout` in `tests/loadtests/appfwSDL/VulkanAppSDL/vulkantools.cpp` for completeness.

Adding a test using `VK_IMAGE_LAYOUT_GENERAL` to `tests/loadtests/vkloadtests` would be enormously helpful.

Unfortunately we have no non-interactive tests for the Vulkan loader. I will have to merge this with my WIP fixes for Vulkan SDK 1.4.357 and run the interactive tests before I can consider merging. It may take a few days.

MarkCallow:
The macOS build failure is because Brew no longer supports Intel x86_64. It has nothing to do with the change in this PR. I will have to update the workflow to remove the x86_64 job.

MarkCallow:
Please rebase against current main to get the fix for the macOS x86_64 build.

MarkCallow (approved):

MarkCallow:
I merged this with my fixes for invalid-usage warnings that are new in Vulkan SDK 1.4.357.0 and everything works with no invalid-usage warnings. I will merge this to main tomorrow so you won't have to wait for me to finish my WIP update for 1.4.357.0 before your fix is available.

OppositeNor:
> I merged this with my fixes for invalid-usage warnings that are new in Vulkan SDK 1.4.357.0 and everything works with no invalid-usage warnings. I will merge this to main tomorrow so you won't have to wait for me to finish my WIP update for 1.4.357.0 before your fix is available.

Sure! Thank you so much!

PR #1249, opened 2026-09-10
Add RISC-V and LoongArch detection

Catty2014:
Add RISC-V and LoongArch detection in `cputypetest.cmake`, in order not to block the build on these platform.

### DIFF
--- a/cmake/cputypetest.cmake
+++ b/cmake/cputypetest.cmake
@@ -44,6 +44,14 @@ armv7S
 arm64
 #elif defined(mips) || defined(__mips__) || defined(__mips)
 mips
+#elif defined(__riscv) && (__riscv_xlen == 32)
+riscv32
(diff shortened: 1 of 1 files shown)


MarkCallow:
Are no other changes needed to make builds work on RISC-V platforms?

Catty2014:
I personally don't have RV64 device available but I did the following:

1. Cross-compiled libktx for RV64 and produced a "looking good" shared object. NO runtime verification on RV hardware yet.

2. Built on LoongArch, integrated into LWJGL(see ref above) and use it run MineCraft.

Thus LoongArch is verified e2e and I hope this works on RISC-V. 

There's no prior issue or PR attempting to accept RV for building libktx.
Glad if you or someone could tell me what else might be blocking on RISC-V.

btw no SIMD support is enabled on RISC-V and LoongArch thus might expect lower performance.

